PR #3, opened 2017-12-06
Kill xmkmf on 10.13

### LINKED ISSUE #2: Remove xmkmf and its reverse-dependencies from the 10.13 distribution
Currently `xmkmf` can't be built on 10.13, so it needs to go away.  Packages that need it either need to be updated not to use it, or themselves to go away.

nieder:
Fixes #2. I think I got all the primary and secondary dependees on xmkmf.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/xmkmf.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/xmkmf.info
@@ -1,6 +1,8 @@
 Package: xmkmf
 Version: 1.0.6
 Revision: 1304
+# No llvm-gcc* on 10.13 XMKMFHS
+Distribution: 10.9, 10.10, 10.11, 10.12
 Description: X11 legacy build tools
 License: BSD
 Maintainer:  Dave Morrison <drm@finkproject.org>
--- a/10.9-libcxx/stable/main/finkinfo/graphics/gocr.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/gocr.info
@@ -1,8 +1,8 @@
 Package: gocr 
 Version: 0.50
-Revision: 2
+Revision: 3
 Maintainer: None <fink-devel@lists.sourceforge.net>
-Depends: netpbm11-shlibs, libjpeg-bin, transfig-graphicx
+Depends: netpbm11-shlibs, libjpeg-bin, fig2dev
 BuildDepends: netpbm11
 #Source: mirror:sourceforge:jocr/%n-%v.tar.gz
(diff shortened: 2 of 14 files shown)


dmacks:
I pushed 76b93f5 "build fixes against openssl110 and cmake-3.9"; looks fine and independent of xmkmf stuff.

dmacks:
Losing transfig means we need to lose gocr. Can gocr switch to using fig2dev instead?

nieder:
If gocr needs transfig, I think it can get away with fig2dev. The only mention I can find of transfig in the code is for fig2dev here (none of the other transfig progs are mentioned):
````
/*
    feel free to expand this list of usable converting programs
    Note 1: the last field must be NULL.
    Note 2: "smaller" extensions must come later: ".pnm.gz" must come
       before ".pnm".
    calling external programs is a security risk
    ToDo: for better security replace gzip by /usr/bin/gzip !
 */
char *xlist[]={
  ".pnm.gz",	"gzip -cd",  /* compressed pnm-files, gzip package */
  ".pbm.gz",	"gzip -cd",
  ".pgm.gz",	"gzip -cd",
  ".ppm.gz",	"gzip -cd",
  ".pnm.bz2",	"bzip2 -cd",
  ".pbm.bz2",	"bzip2 -cd",
  ".pgm.bz2",	"bzip2 -cd",
  ".ppm.bz2",	"bzip2 -cd",
  ".jpg", 	"djpeg -gray -pnm",  /* JPG/JPEG, jpeg package */
  ".jpeg",	"djpeg -gray -pnm",
  ".gif",	"giftopnm -image=all",  /* GIF, netpbm package */
  ".bmp",	"bmptoppm",
  ".tiff",	"tifftopnm",
  ".png",	"pngtopnm", /* Portable Network Graphics (PNG) format */
  ".ps",	"pstopnm -stdout -portrait -pgm", /* postscript */
  ".eps",	"pstopnm -stdout -portrait -pgm", /* encapsulated postscript */
   /* gs -sDEVICE=pgmraw -sOutputFile=- -g609x235 -r141x141 -q -dNOPAUSE */
  ".fig",	"fig2dev -L ppm -m 3", /* xfig files, transfig package */
  NULL
};
````
So this seems to mean it should be a Recommends or Suggests instead. fig2dev is used extensively in making the examples, but we don't build them and are commented out in configure.in.
And Ubuntu has fig2dev as the depends for their gocr, so if it is needed, fig2dev should be enough.

dmacks:
I pushed a squashed version of the two changes to gocr, getting that package out of the blockers list

dmacks:
...and pushed the main "No xmkmf on 10.13." a4e1e4b as-is

PR #10, opened 2017-12-13
Lazarus: Update to new upstream version (1.8.0)

kamischi:


dmacks:
No idea why this pull request has so many other changes in it...maybe changes from master after the branch got merged in rather than rebasing? I assume 6ae9ef7 is the only intended one here, so I cherry-picked it into master. Feel free to re-open if something else is needed here.

PR #12, opened 2017-12-16
Update R to latest releases and add 10.13 compatibility tweaks

nieder:
@babayoshihiko
From @jwhowarth submissions
https://sourceforge.net/p/fink/package-submissions/4981/
https://sourceforge.net/p/fink/package-submissions/4978/
https://sourceforge.net/p/fink/package-submissions/5021/

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
@@ -3,6 +3,7 @@ Package: r-base31
 Version: 3.1.3
 Revision: 6
 Description: R Framework
+Type: rversion (3.1)
 Maintainer:  BABA Yoshihiko <babayoshihiko@mac.com>
 Depends: <<
 	%N-shlibs (=%v-%r),
@@ -106,8 +107,6 @@ InfoTest: <<
 <<
 InstallScript: <<
   #!/bin/sh -ex
-  FW_VER=`echo %v | sed -e's/[\.][0-9]$//'`
-
   # prefix=%i/Library/Frameworks
   make install DESTDIR=%d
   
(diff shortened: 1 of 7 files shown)


dmacks:
We currently use a Distribution tag to limit r-base32 to 10.{9-12}, but we don't limit any other r-base*. Should we similarly restrict r-base31? Do we even need to keep dragging r-base33 along, or is r-base34 sufficient to keep going forward?

dmacks:
I just pushed a tweak to remove a stray hardcoded value...one less difference among the r-base*.info files.

Would it be cleaner to use a `Type:` to hold the r-base version (3.4) rather than calculating it using shell? That would allow it to be used in more places (RuntimeVars, Shlibs).

nieder:
Personally, I don't know if there's any reason to keep r33 since I think most rmods that use r33 are available as r34. There are some in r31 that are not available in later releases (but I would imagine that's most likely because they're abandoned).

nieder:
Ugh. I don't know how to fix the conflicts that are now popping up. Presumably something to do with 86b604b.

dmacks:
I think you didn't merge master into your fork or branch before you started editing. The "command line instructions" might or might not be clear enough to be useful:/ I can work around it by redoing d91647c in master, and then you can rebase the actual "latest releases and 10.13 compat" branch on top of it.

nieder:
The first time I tried to update my R branch to master, it duplicated the 20 or so commits from master to the R branch as separate commits and I was afraid that those would then show up here and also end up being duplicated in master once this PR is merged. So I rolled back my branch to before I merged the master changes to it, and pushed, and THUNK! I'll try the command line checkout instructions to see if they help the conflicts.

nieder:
Going to start from scratch

PR #16, opened 2017-12-19
Update R to latest releases and add 10.13 compatibility tweaks (try 2)

nieder:
This tries to fix the mess from PR #12.

Adds a new commit to add Type: fields for easier maintenance.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
@@ -3,6 +3,7 @@ Package: r-base31
 Version: 3.1.3
 Revision: 6
 Description: R Framework
+Type: rversion (3.1)
 Maintainer:  BABA Yoshihiko <babayoshihiko@mac.com>
 Depends: <<
 	%N-shlibs (=%v-%r),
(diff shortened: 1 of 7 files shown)


dmacks:
If we can add -r33 variants to cran-chemospec-rXX and cran-js-rXX, we can jetison -r31 altogether on 10.13 (see #17 for variant comparison)

dmacks:
Might be nice to merge af49716 with 4bd3b0d, and aeba2f2 with 20229fb. They're non-adjacent, so you can't "just" squash them together, but you could re-order first (/me learns scary-powerful new git trick!), or else patch master manually and write a new commit for them instead of merging the branch.

dmacks:
f92e6c3 looks fine.

dmacks:
PR #18 makes R31 not go beyond 10.12, so we could not bother with 10302b5

nieder:
Can specific commits be omitted when merging a pull request?

For af49716 with 4bd3b0d (both r34), and aeba2f2 with 20229fb, I would prefer to keep them separate to trace their person of origin.

dmacks:
From within the target (master), 'git cherrypick <commit>' clones <commit> from some other fork or branch (as if you'd applied the change manually, git-add'ed, and git-committed with the same msg). I don't think you can do a selective merge in the "merge" technical realm.

nieder:
Can this be merged? And then proceed with killing r-base31 (#18)? This at least preserves a record on how r-base31 can be made to work if it ever needs to.

dmacks:
Go for it.

dmacks:
I merged it. GitHub on iPhone is...interesting.

Issue #17, opened 2017-12-20
R version support


dmacks:
Spawning a new topic from #12, wherein:

@dmacks:
>We currently use a Distribution tag to limit r-base32 to 10.{9-12}, but we don't limit any other r-base*. Should we similarly restrict r-base31? Do we even need to keep dragging r-base33 along, or is r-base34 sufficient to keep going forward?

@nieder:
>Personally, I don't know if there's any reason to keep r33 since I think most rmods that use r33 are available as r34. There are some in r31 that are not available in later releases (but I would imagine that's most likely because they're abandoned).

Here, let's figure out what rmods need to be brought forward vs be marked "do not bring forward" (so we don't have to have this conversation again in the future), and then whether to drop r31 and/or r33 on 10.13....

dmacks:
Here are the differences among *-r3{1,2,3,4} on my 10.12 machine:

package | -r31 | -r32 | -r33 | -r34
-|-|-|-|-
bioconductor-biocparallel-rXX | no | no | no |
bioconductor-preprocessorcore-rXX ||| no | no
cran-blob-rXX | no |||
cran-cellranger-rXX | no |||
cran-chemospec-rXX ||| no | no
cran-gcexplorer-rXX | no | no | no |
cran-its-rXX | no | no | no |
cran-js-rXX ||| no | no
cran-latticist-rXX | no | no | no |
cran-mfda-rXX | no | no | no |
cran-modelmetrics-rXX | no |||
cran-multicore-rXX | no | no | no |
cran-mvpart-rXX | no | no | no |
cran-nmrs-rXX | no | no | no |
cran-pkfit-rXX | no | no | no |
cran-pkmodelfinder-rXX | no | no | no |
cran-pktools-rXX | no | no | no |
cran-plotly-rXX | no |||
cran-psm-rXX | no | no | no |
cran-purrr-rXX | no |||
cran-rcmdrplugin.doby-rXX | no | no | no |
cran-rcmdrplugin.ezr-rXX | no | no | no |
cran-rgdal-rXX | no | no | no |
cran-rglwidget-rXX | no |||
cran-robustrankaggreg-rXX | no | no ||
cran-sets-rXX | no |||
cran-spgrass6-rXX | no | no | no |
cran-tidyr-rXX | no |||
cran-tidyselect-rXX | no |||
cran-weightedkmeans-rXX | no | no | no |
metaboanalyst-tomcat6-rXX | no | no | no |
[...]

PR #18, opened 2017-12-20
Turn off R31 on 10.13

dmacks:
Per #17, add -r33 as necessary and Distribution tag to stop carrying -r31 along beyond 10.12 (same as -r32 already is)


dmacks:
Pending coordination with #16 (will make one of its commits moot) to avoid diverging forks...

nieder:
I've got a bunch of people in my house limiting access to my production machine. Will clear up by Sunday night.

dmacks:
I should have re based rather than merged, so there is no cruft. And now there's still a xconflict even after merging that? I'll fiddle with it in a day or two, or can just cherry pick the relevant commits instead of merging.

dmacks:
All the key parts are cherry-picked or hand-redone to avoid conflicts.

PR #21, opened 2017-12-24
Update mopac7 to use gcc7

kamischi:


kamischi:
hm. It looks, that i screwed it again with all the in between commits duplicated. Simply kill this pull request. I probably need a different workflow without my own fork on Github.

dmacks:
Is a3392beb818be05fa61ee36422f22b4ff409d09b the actual single commit you want to do? If so, I can chery-pick it out of this mess and not need to redo the fork/branch/PR/etc.

nieder:
I think to avoid having all the commits show up is to do your work on a branch in your forked copy rather than from its master (just use origin/master to track upstream/master and branch off master as needed for works). Then do the PR from that branch and that should only pull in your specific commits, and not all the commits from your master.

kamischi:
Am 24.12.2017 um 20:43 schrieb Daniel Macks <notifications@github.com>:
> 
> Is a3392be the actual single commit you want to do? If so, I can chery-pick it out of this mess and not need to redo the fork/branch/PR/etc.

Yep. that is it. Sorry for the trouble - Michael.

dmacks:
Handling via #22 ...

PR #29, opened 2018-01-01
ncurses-6.0 from tracker

nieder:
This is a transfer of @jwhowarth's update for ncurses-6 and libncurseswfrom the tracker: https://sourceforge.net/p/fink/package-submissions/4803/ and https://sourceforge.net/p/fink/package-submissions/4804/

Things to figure out:
- the submission uses a new libN, but ./configure has a `--with-abi-version` flag that can be used to set the libN back to 5 as in our current package. Do we want a new libN for ncurses?
- if we do keep the new libN=6, does the ncurses-%v=6 package get to keep its `Essential: true` field?

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
@@ -1,0 +1,113 @@
+Package: libncurses6-shlibs
+Version: 6.0
+Revision: 1
+
+GCC: 4.0
+Source: ftp://ftp.gnu.org/pub/gnu/ncurses/ncurses-%v.tar.gz
+Source-MD5: ee13d052e1ead260d7c28071f46eefb1
+SourceDirectory: ncurses-%v
+BuildDepends: fink (>= 0.13.0)
+Depends: base-files
+Essential: true
+PatchScript: <<
+#!/bin/sh -ex
+	# apply upstream patchsets
+	for p in patches/*.patch.gz; do
+		gzip -cd $p | patch -p1
+	done
+
+	# Fix warnings in clang by removing some useless compiler
+	# switches (they are not useful for gcc either).
+	perl -pi -e "s/-no-cpp-precomp//" configure
+
+	#Need to remove these entries to avoid failure on case sensitive file systems
+	perl -pi -e "s/hp70092a\|//; s/hp2621a\|//; s/2621a\|//" misc/terminfo.src
+<<
+ConfigureParams: <<
+  --without-ada \
+  --without-pkg-config \
+  --with-shared \
+  --with-debug \
+  --enable-termcap \
+  --mandir=%p/share/man \
+  --with-terminfo-dirs="%p/share/terminfo:/usr/share/terminfo" \
+  --with-default-terminfo-dir="%p/share/terminfo" \
+  --enable-no-padding \
+  --enable-sigwinch \
+  --enable-tcap-names \
+  --enable-hard-tabs \
+  --enable-overwrite \
+  --disable-mixed-case \
+  --libdir=%p/lib/ncurses \
+  cf_cv_typeof_chtype=long \
+  cf_cv_gcc_inline=no
+<<
+NoSetCFLAGS: true
+NoSetLDFLAGS: true
+NoSetCPPFLAGS: true
+UseMaxBuildJobs: false
+InstallScript: <<
+#!/bin/sh -ev
+  make install DESTDIR=%d DYLD_LIBRARY_PATH=`pwd`/lib:`pwd`/lib/ncurses
+# add extra symlinks in the main lib directory
+  ln -s ncurses/libform.6.dylib %i/lib/libform.dylib
+  ln -s ncurses/libmenu.6.dylib %i/lib/libmenu.dylib
+  ln -s ncurses/libncurses.6.dylib %i/lib/libncurses.dylib
+  ln -s ncurses/libpanel.6.dylib %i/lib/libpanel.dylib
+  ln -s ncurses/libncurses.6.dylib %i/lib/libcurses.dylib
+  ln -s ncurses/libncurses.6.dylib %i/lib/libtermcap.dylib
+<<
+SplitOff: <<
+ Package: libncurses6
+ Depends: libncurses6-shlibs (= %v-%r)
+ BuildDependsOnly: true
+ Conflicts: ncurses-dev, libncurses5-64bit, libncurses5
+ Replaces: ncurses-dev, ncurses (<= 5.3-20031018-2), libncurses5-64bit, ncurses (<= 5.7-20100227-1), libncurses5
+ Provides: libncurses6-dev
+ Files: bin/ncurses6-config lib/ncurses/*.a lib/ncurses/libcurses.dylib	lib/ncurses/libform.dylib lib/ncurses/libmenu.dylib lib/ncurses/libncurses.dylib lib/ncurses/libpanel.dylib include lib/libcurses.dylib lib/libform.dylib lib/libmenu.dylib lib/libncurses.dylib lib/libpanel.dylib lib/libtermcap.dylib
+ DocFiles: README ANNOUNCE NEWS INSTALL TO-DO MANIFEST
+ Description: Full-screen ascii drawing library
+<<
+SplitOff2: <<
+ Package: ncurses
+ Depends: libncurses6-shlibs (>= %v-%r)
+ Essential: true
+ Files: bin lib/terminfo share/man share/terminfo
+ DocFiles: README ANNOUNCE NEWS INSTALL TO-DO MANIFEST
+ Description: Executable files for libncurses5
+<<
+Shlibs: <<
+ %p/lib/ncurses/libform.6.dylib 6.0.0 %n (>= 6.0-1)
+ %p/lib/ncurses/libmenu.6.dylib 6.0.0 %n (>= 6.0-1)
+ %p/lib/ncurses/libncurses.6.dylib 6.0.0 %n (>= 6.0-1)
+ %p/lib/ncurses/libpanel.6.dylib 6.0.0 %n (>= 6.0-1)
+<<
+DocFiles: README ANNOUNCE NEWS INSTALL TO-DO MANIFEST
+Description: Shared libraries for libncurses5 package
+DescPort: <<
+More changes by Chris Zubrzycki to be more compatable with apple's lib
+-udates to the (sometimes) monthly releases of ncurses (in patch form)
+
+The --enable-overwrite flag creates the libcurses.dylib symlink.
+
+Must pass cf_cv_typeof_chtype=long to configure to maintain ABI
+compatability for 64-bit library. Otherwise chtype is unsigned long
+in 5.4 but unsigned int in 5.7.
+
+Added patchscript to avoid failure on case sensitive file systems.
+<<
+DescPackaging: <<
+First revision by David Ross.
+Previous versions by Christoph Pfisterer.
+All the hard work in this version done by Chris Zubrzycki.
+Updated to 5.7 by Daniel Johnson.
+
+	Fink needs to mirror Essential packages on its own server).
+
+	Disable pkg-config so we don't need to make it Essential:true
+	(as dependency of Essential:true ncurses). gen-pkgconfig says
+	the *-config scripts are the recommended way anyway.
+<<
+License: OSI-Approved
+Maintainer: Fink Core Group <fink-core@lists.sourceforge.net>
+Homepage: http://www.gnu.org/software/ncurses/ncurses.html
--- a/10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
@@ -11,11 +11,6 @@ Depends: base-files
 Essential: true
 PatchScript: <<
 #!/bin/sh -ex
-	# apply upstream patchsets
-	for p in patches/*.patch.gz; do
-		gzip -cd $p | patch -p1
-	done
-
 	# Fix warnings in clang by removing some useless compiler
 	# switches (they are not useful for gcc either).
 	perl -pi -e "s/-no-cpp-precomp//" configure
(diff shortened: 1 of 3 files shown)


dmacks:
I see `--without-pkg-config`, and agree with its reasoning:
> Disable pkg-config so we don't need to make it Essential:true
(as dependency of Essential:true ncurses). gen-pkgconfig says
the *-config scripts are the recommended way anyway.
But I wonder if there is a way to get the .pc installed even without detecting the presence of /sw/bin/pkg-config. It looks like the only reason to detect that is to decide if to generate and install the .pc file (and simply short-circuits some otherwise-usable logic for determining the installation path).

dmacks:
Switching up to a new libversion (== new package-name for the -dev) would allow us to stop dragging along the static libs and fix detection/linking/runtime-dependencies when reverse-builddepends are migrated. If we keep the same -dev package-name, we should keep the .a out of inertia because we don't know how many reverse-builddepends actually care about it.

10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
```diff
@@ -0,0 +1,113 @@
...
+  ln -s ncurses/libncurses.6.dylib %i/lib/libcurses.dylib
+  ln -s ncurses/libncurses.6.dylib %i/lib/libtermcap.dylib
+<<
+SplitOff: <<
+ Package: libncurses6
+ Depends: libncurses6-shlibs (= %v-%r)
+ BuildDependsOnly: true
+ Conflicts: ncurses-dev, libncurses5-64bit, libncurses5
+ Replaces: ncurses-dev, ncurses (<= 5.3-20031018-2), libncurses5-64bit, ncurses (<= 5.7-20100227-1), libncurses5
+ Provides: libncurses6-dev
```
dmacks: There was a libncurses5:Provides:libncurses5-dev since long ago; do we really need this cargo-culting if we are starting with a new libversion? "foo/foo-shlibs" is a well-established pattern for library package-sets.

dmacks:
The .info filenames are getting chaotic. Especially if we switch to a new libversion, we should standardize on *-shlibs.info to make it easier to compare the libversions to each other (harder if the parent vs splitoff switch places).

dmacks:
libncursesw6/libncursesw6-shlibs are not essential, so the DescPackaging notes are not correct. They are explicitly tagged `Essential:false` IIRC because they originally were `Essential:true` (maybe even part of the main package instead of separate *w5* set?) and this was the only way to move towards allowing them to be uninstalled. If we're switching to a new libversion, can probably scrap this tag altogether since I think it is only boolean (false==undef) not tri-state (true/false/undef)...someone please double-check core code.

beren12:
I chose to follow the Debian library scheme when we did splitoffs… since the library is the main package it's the libfoo name, with -dev and -bin splitoffs. Personally I prefer that for library packages still.

dmacks:
But "the library is the main package it's the libfoo name, with -dev and -bin splitoffs." does not match current layout practice or packging policy. I think libfoo (or foo) as the runtime libraries (rather than libfoo-shlibs or foo-shlibs) is unique for libiconv. For the record, I'm agnostic about libfoo/libfoo-shlibs vs libfoo-dev/libfoo-shlibs.

TheSin-:
I agree with @beren12 I prefer the Debian scheme as well, it makes porting much easier, specially since I ported all of debhelper.

dmacks:
Switching to a new libversion would also allow us to unbury the library files (use %p/lib instead of %p/lib/ncurses). Presumably we did that in order to avoid masking apple's libraries, but then we moved the main links to %p/lib anyway and use fink-package-precedence enforce use of fink's.

dmacks:
Regarding `--with-abi-version`, is the new version actually ABI-compatible? There are about 135 reverse-depends, some of which look like no-longer-relevant inherited deps and many of which have complex versioning dating back to the 10.4-transitional era. The migration would be a chance to clean up some uselessness.

nieder:
I managed to get the following in ncurses-5  with pkg-config **un**installed.
~~~
checking if you want to use pkg-config... no
checking if we should install .pc files for none... yes
... and inside ncurses.deb ...
drwxr-xr-x root/admin        0 2018-01-01 20:37 ./sw/lib/pkgconfig/
-rw-r--r-- root/admin      262 2018-01-01 20:37 ./sw/lib/pkgconfig/form.pc
-rw-r--r-- root/admin      266 2018-01-01 20:37 ./sw/lib/pkgconfig/form_g.pc
-rw-r--r-- root/admin      262 2018-01-01 20:37 ./sw/lib/pkgconfig/menu.pc
-rw-r--r-- root/admin      266 2018-01-01 20:37 ./sw/lib/pkgconfig/menu_g.pc
-rw-r--r-- root/admin      288 2018-01-01 20:37 ./sw/lib/pkgconfig/ncurses++.pc
-rw-r--r-- root/admin      254 2018-01-01 20:37 ./sw/lib/pkgconfig/ncurses.pc
-rw-r--r-- root/admin      272 2018-01-01 20:37 ./sw/lib/pkgconfig/ncurses_g.pc
-rw-r--r-- root/admin      264 2018-01-01 20:37 ./sw/lib/pkgconfig/panel.pc
-rw-r--r-- root/admin      268 2018-01-01 20:37 ./sw/lib/pkgconfig/panel_g.pc
~~~
~~~
$ cat ncurses.pc 
prefix=/sw
exec_prefix=${prefix}
libdir=/sw/lib/ncurses
includedir=${prefix}/include
major_version=5
version=5.9.20110507

Name: ncurses
Description: ncurses 5.9 library
Version: ${version}
Requires: 
Libs: -L${libdir} -lncurses 
Cflags: -I${includedir}
~~~
The .pc file generation doesn't need to have %p/lib/pkgconfig exist during buildtime either. With ncurses-6, I can build the .pc files, but not yet install them properly.

I have no idea if setting `--with-abi-version=5` with ncurses-6 is compatible with ncurses-5. I'm trusting upstream did the right thing. But if there's any doubt, then clean break to abi=6 (default) would make sense.

My only tests so far were to make sure v6 of libncurses and libncursesw built as is. I wanted to have finalized decisions on our various options before I went out to really modify the build.
- [ ] new libN.
- rename headers package to libncurses6-dev?
- standardize all 4 packages as FOO-shlibs.info (libncurses{,w}{5,6}-shlibs.info)?
- get rid of carryover stuff from libncurses5
-- get rid of static libs
-- unbury libraries
-- kill Provides: in headers pkg
- [ ] generate .pc files w/out needing pkg-config

dmacks:
For ncurses6:
```
  --with-pkg-config=/usr/bin/true \
  --with-pkg-config-libdir=%p/lib/pkgconfig \
  --enable-pc-files \
```
gets me .pc created and installed, both with and without pkg-config itself installed:
```
checking if you want to use pkg-config... /usr/bin/true
checking for /usr/bin/true library directory... /sw/lib/pkgconfig
checking if we should install .pc files for /usr/bin/true... yes
checking for suffix to add to pc-files... none
```
Weird but functional spaghetti. Need to add lib/pkgconfig to the Files: for libncurses6 (to accompany the headers, etc.)

dmacks:
Thanks for the ncurses5 example. Turns out it works for ncurses6 also:
```
  --enable-pc-files \
[...]
  PKG_CONFIG=%p/bin/pkg-config \
  PKG_CONFIG_LIBDIR=%p/lib/pkgconfig
```
and doesn't even need the "Don't test for presence of PKG_CONFIG_LIBDIR" PatchScript (ncurses6 doesn't appear to have anything comparable in its configure scripts).

danielj7:
I would recommend not having a separate libncursesw6 package. There's no good reason to keep carrying both around. Just build libncurses6 with --enable-widec and add --disable-lib-suffixes to cause it to build without the 'w' suffix. That's what the system ncurses does. I'd also recommend --with-pthread while we're doing a new libN.

Also adding cf_cv_typeof_chtype=long to ConfigureParams hasn't been necessary since 32 bit days.

The big ABI difference between 5 and 6 is that many types became opaque and you need to use accessor functions instead of accessing struct fields directly.

dmacks:
While looking at whether a new libversion package's "ncurses" package should retain Essential:true, I noticed that the API manpages are in ncurses rather than the headers package. TODO: fix that.

But ncurses in the new .info should retain Essential:true. Otherwise it is no longer essential and might not get updated (sync problem with the new lib?). Or if it got updated it would no longer be essential and then could get removed altogether. The question is: what *is* that package's role? It looks like it's both user-level utility programs and runtime terminfo data for the library? Maybe the terminfo should be sent to a separate package (maybe even a separate .info?) that can be Essential and have the libs Depends on it, and then ncurses utility programs could be a no-longer-Essential splitoff of the libs' .info.

nieder:
FWIW, Ubuntu sends their /sw/share/terminfo/ to a different package `ncurses-term`. Only `ncurses-base` and `ncurses-bin` are essential. `ncurses-base` has a subset of basic terminfo files. `ncurses-bin` are the executables. For them, they link to libtinfo.so.5. For us, they link to libncurses.5.dylib.

The libtinfo library is a subset of the libncurses library that can be built with the `--with-termlib` flag. We don't do this split.
~~~
When building the ncurses library, organize this as two parts:  the
	curses library (libncurses) and the low-level terminfo library
	(libtinfo).  This is done to accommodate applications that use only
	the latter.  The terminfo library is about half the size of the total.
~~~

nieder:
TODO:
- [ ] bump ABI to 6
- [ ] drop libncurses**w**6 and instead put —enable-widec in libncurses6 (plus extra flags as @danielj7 suggested)
- [ ] unbury libraries and put them back into %p/lib
- [ ] kill static .a
- [ ] standardize how to generate the .pc files across new and existing .info to make file comparisons easier
- [ ] drop cf_cv_typeof_chtype flag
- [ ] move man3 files from ncurses to libncurses{5,6}
- [ ] retain Essential: true for ncurses and libncurses6-shlibs
- [ ] offload terminfo files to a new splitoff
- [ ] decide on naming for new libN=6 headers: libncurses6 vs libncurses6-dev. libncurses6-dev would allow dropping the `Provides` field.

beren12:
> On Jan 6, 2018, at 4:39 PM, nieder <notifications@github.com> wrote:
> 
> TODO:
> 
> 	•  bump ABI to 6
> 	•  drop libncursesw6 and instead put —enable-widec in libncurses6 (plus extra flags as @danielj7 suggested)
> 	•  unbury libraries and put them back into %p/lib
> 	•  kill static .a
> 	•  standardize how to generate the .pc files across new and existing .info to make file comparisons easier
> 	•  drop cf_cv_typeof_chtype flag
> 	•  move man3 files from ncurses to libncurses{5,6}
> 	•  retain Essential: true for ncurses and libncurses6-shlibs
> 	•  offload terminfo files to a new splitoff
> 	•  decide on naming for new libN=6 headers: libncurses6 vs libncurses6-dev. libncurses6-dev would allow dropping the Provides field.
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

I still object to the name libncurses6-shlibs, with TheSin. It didn’t used to be policy to require it this way, and we should consider adding allowances in the guide for library packages being libfoo, libfoo-bin and libfoo-dev. It just makes sense and it how everyone else names library packages

nieder:
I'm agnostic on requiring -dev for the headers SplitOff, and since the current package is ncurses/libncurses5-shlibs/libncurses, I would lean towards keeping the name pattern this way with libN=6 to simplify transitioning packages to it.

This current SplitOff name pattern is ancient for ncurses and dates back to the 10.2-gcc3.3 period.
https://github.com/fink/fink-distributions/blob/master/10.2-gcc3.3/stable/main/finkinfo/base/ncurses.info

The TODO list above is not final. It's the current state (to my eye) of the discussion. If something changes, the TODO list will be edited.

nieder:
Apart from the SplitOff naming issue, are there other new issues to address or other things in the TODO list are still under debate?

dmacks:
What is it about "ncurses" that makes it Essential, the bin/* or the share/{tabset,terminfo}? I had assumed it was the share/ stuff, as "/sw/share/terminfo" is accessed by the .dylib, rather than some command-line programs that themselves link to the libraries. By keeping ncurses Essential, we assure it is installed, so we avoid a circular dependency of binary->library->datafile where binary and datafile are in the same package separate from library. Assuming so, I would think only the package with the datafiles would need to retain the Essential flag. Debian has a similar binary/library/datafile division and calls their datafiles package "ncurses-term" or "ncurses-base".

dmrrsn-fink:
Back in the day, this had to do with bootstrapping IIRC.

nieder:
Current remaining issues to sort out from the list above:
- [ ] retain Essential: true for ncurses and libncurses6-shlibs
- [ ] offload terminfo files to a new splitoff
- [ ] decide on naming for new libN=6 headers: libncurses6 vs libncurses6-dev. libncurses6-dev would allow dropping the `Provides` field.

The first two would be to somewhat align with Debian's layout if we so desired, but it would slightly increase packaging complexity (making a SplitOff for a new terminfo library and moving some terminfo files to a new SplitOff)

SplitOff naming hasn't come to a final decision. Current naming as submitted continues the scheme we have had for libncurses5-shlibs since 2003.
Options are:
* libncurses6-shlibs, libncurses6, Provides::libncurses6-dev (current)
* libncurses6-shlibs, libncurses6-dev (I prefer this because it reduces ambiguity)
* libncurses6, libncurses6-dev, libncurses{6?}-bin (similar to other distros)

danielj7:
How is this coming along? Since it's been a year I figured I'd bring it up again. :)

As far as package naming goes, I'd prefer either 1 or 2 but have no strong opinions otherwise. 3 might be what other distos use but is wildly different from the vast majority of fink packages.

BTW, the current upstream version is 
https://invisible-mirror.net/archives/ncurses/current/ncurses-6.1-20181229.tgz

nieder:
There was disagreement between the naming of the SplitOffs with no resolution between using the general Fink naming convention and Debian's. If there's consensus on the Debian scheme, I'm OK with that, but that goes against the vast majority of our package naming (only 2 out of ~650 in my installed Fink tree).

nieder:
ncurses 6.2 is out.
SplitOff names are still unchanged from original and match what we currently have for ncurses 5.9:
```
libncurses6-shlibs (libraries)
libncurses6, Provides::libncurses6-dev (headers)
ncurses (binaries and terminfo)
```

cooljeanius (approved):
Just tested this PR with the `-m` flag; it appears to work fine (on Big Sur, with Xcode 13, at least). The fact that libncurses6 needs `-Wno-error=implicit-function-declaration` is a bit worrying, and ideally should be fixed so as to not require that, but that can be handled separately, IMO. Also libncursesw5 prints a lot of `-Wunused-command-line-argument`, but that's pretty minor.

nieder:
The `-Wno-error=implicit-function-declaration` is from some upstream autoconf macro. I changed it after the fact to `-Werror=implicit-function-declaration` and the build finished cleanly w/ no actual patching needed.

10.9-libcxx/stable/main/finkinfo/base/libncurses6-shlibs.info
```diff
@@ -17,6 +17,9 @@ PatchScript: <<
 
 	#Need to remove these entries to avoid failure on case sensitive file systems
 	perl -pi -e "s/hp70092a\|//; s/hp2621a\|//; s/2621a\|//" misc/terminfo.src
+
+	# Turn this flag on to catch any issues
+	perl -pi -e 's|-Wno-error=implicit-function-declaration|-Werror=implicit-function-declaration|g' configure
```
cooljeanius, line 22: I think I'd just remove it entirely instead of replacing it? I mean, since it's the default now, anyways... idk, I wonder what @akhansen thinks?

dmacks:
```--without-normal``` indeed inhibits the regular *.a libs. And ```-without-debug``` rather than ```--with-debug``` inhibits the *_g.a static libs.

My mistake about the leakage of libncurses++.a (I was looking at wrong sandbox). The ```rm``` does indeed kill it. Annoying that lib is not controlled by the same flags as the main lib. ```--without-cxx-binding``` inhibits libncurses++ altogether. Or instead ```--with-cxx-shared``` builds libncurses++.dylib with the same libversioning as the rest of the lib suite (install_name .6. and compatibility_version 6.0.0). I have no idea the merits of supplying that .dylib (debian does only supplies it as .a not .so but redhat does supply .so).

I recognized each of the three Replaces:ncurses versions. It just seemed odd to have the first entries that are subsets of what the third handles.

Shlibs entries...that makes sense.

PR #30, opened 2018-01-02
Remove old distributions prior to 10.7

fingolfin:
They haven't been supported for years, in some cases over a decade.
If anybody wants to use them, they are free to create a fork of this
repository, with a new branch based before this commit.

Leave 10.7 in, because (a) it is conceivable a few people might still
be using it, and (b) it seems to be sometimes useful to dig through it
when working on the current tree, 10.9-libcxx


nieder:
Before deleting pre-10.7 here, can we clone fink-distributions into fink/fink-distributions-archive (or similar name), and then delete 10.7 and 10.9-libcxx from there? I've found it many useful many times to be able to check the history of a file to figure where/when a line came from. With our current way of populating new trees, history dies when it's first populated and so you need older dists. I don't want to lose that.

The archive tree could be read-only. I just want access to the history. And git doesn't have an attic the way CVS does, so tracking deleted files (or remembering a specific commit ID to start my search from) is easier in CVS.

Aside: would git submodules do what we want to get each tree into a separate 'repo' so that users don't have to checkout ALL of dists?

TheSin-:
This will break the PDB, please wait on this till I fix the pdf scripts to stop pulling it.

beren12:
People still use older distributions on older systems, could we just not show it by default or something? I still have PPC machines and a 10.6 VM.

TheSin-:
honestly I'm not sure what the benefit to removing some of our history is?  Just to make it look cleaner?

Even if we haven't used them, what's the benefit?

gecko2:
On one side, I would not remove it as it's our history. On the other side the older fink versions are not using the stuff or any changes made here.

I prefer to keep it and recommend to backport the git pull stuff to the latest fink version used in those distributions or if possible (and compatible) push the latest released version to those distributions. Both solutions should be handled as one-time action, as we dropped the old distributions for good reasons =)

akhansen:
In my view, we don't have enough active person-hours to keep up with the distributions that we officially support.
If somebody on an older distribution wants to try to add in updates, they are absolutely free to create a fork of the project to do so.  Ideally they'd also update things so that we aren't the primary support channel.

dmacks:
For the record, I just had occasion to look back as far as 10.3, including history:)

PR #31, opened 2018-01-05
Update or note relationship with libidn2

nieder:
Notes packages that currently fail to build w/ libidn2 or fail to detect it.
Other maintainers with libidn packages not included here:
@dmacks 
@danielj7 
@gecko2 
@bcully 
@akhansen
Martin Costabel
Dave Reiser
Tomoaki Okayama

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/crypto/gsasl9.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/gsasl9.info
@@ -1,16 +1,31 @@
 Package: gsasl9
-Version: 0.2.24	
-Revision: 1005
+# FTBFS LIBIDN2
+# https://gitlab.com/libidn/libidn2/issues/28
+Version: 0.2.29
+Revision: 1
 ###
-Source: http://josefsson.org/gsasl/releases/gsasl-%v.tar.gz
-Source-MD5: b5840357baf74310569045faf93f9ce5 
+Source: ftp://alpha.gnu.org/gnu/gsasl/gsasl-%v.tar.gz
+Source-MD5: f513d8eed4f8d47f598cf1ac052fefb3 
 Depends: <<
-  gnutls28-shlibs (>= 3.3.0-1), %N-shlibs (>= %v-%r), readline6-shlibs,
-  libgettext8-shlibs, libiconv, libidn-shlibs
+	%N-shlibs (>= %v-%r),
+	gnutls28-shlibs (>= 3.3.0-1),
+	readline7-shlibs,
+	libgettext8-shlibs,
+	libiconv,
+	libidn-shlibs
 <<
 BuildDepends: <<
-  fink-package-precedence, gawk, gettext-tools, gnutls28 (>= 3.3.0-1), gtk-doc, libgettext8-dev,
-  libiconv-dev, libidn, libxslt-bin, pkgconfig, readline6
+	fink-package-precedence,
+	gawk,
+	gettext-tools,
+	gnutls28 (>= 3.3.0-1),
+	gtk-doc,
+	libgettext8-dev,
+	libiconv-dev,
+	libidn,
+	libxslt-bin,
+	pkgconfig,
+	readline7
 <<
 
 DocFiles: INSTALL.txt ABOUT-NLS README AUTHORS COPYING ChangeLog NEWS THANKS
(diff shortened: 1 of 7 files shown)


dmacks:
jwhois has added BuildDepends:gdbm3, which appears to be used directly but it doesn't have the library linked. Here's ./configure output:
```
checking for dbm_open... yes
```
then
```
checking gdbm.h usability... yes
checking gdbm.h presence... yes
checking for gdbm.h... yes
checking ndbm.h usability... yes
checking ndbm.h presence... yes
checking for ndbm.h... yes
checking dbm.h usability... yes
checking dbm.h presence... yes
checking for dbm.h... yes
checking db1/ndbm.h usability... no
checking db1/ndbm.h presence... no
checking for db1/ndbm.h... no
```
Apple gives us ndbm via libSystem, so the first dbm_open test succeeds without having to test adding various -l*dbm links. Crossing fink's headers from one db flavor with apple's ABI from another flavor is...not good.

Adding:
```
ConfigureParams: ac_cv_func_dbm_open=no
```
causes:
```
checking for dbm_open... (cached) no
checking for gdbm_open in -lgdbm... yes
```
to give self-consistent gdbm use, as `otool -L` confirms the need for added `Depends:gdbm3-shlibs`.

nieder:
Thanks. And it even works with an update to gdbm4

dmacks:
Fixed mine (elinks as 2736e26d6c). Thanks for tracking!

dmacks:
I just scanned all 10.13 packages and don't see any other dependency mismatches between gdbm* and gdbm*-shlibs.

nieder:
@akhansen wget-idn-1.19.2 fails to detect libidn. needs libidn2
~~~
checking for library containing idn2_lookup_u8... no
configure: WARNING: *** LIBIDN2 was not found. You will not be able to use IDN2008 support
~~~
I can update (only (B)Dep modification and throw on here

akhansen:
I just updated wget to use libidn2.0 in 1704410 .

danielj7:
libidn and libidn2.0-dev don't actually have any files in common and should be able to coexist. The Conflicts/Replaces can be removed.

danielj7:
I tried to update monotone to libidn2.0 but it uses the old stringprep api which isn't there anymore. It's just using it for converting to/from utf8 so it might be fixable.

nieder:
I put the C/R fields to make sure the packages didn't coexist and be more deterministic in case a Dependee had code to detect both and only use one. But I don't think any package (at least those we have) checks for one and then for the other. Is there a need to be able to have both present at the same time?

nieder:
These are probably OK to commit (either marked as FTBFS with libidn2 or updated). The rest would be up to their respective maintainers as they see fit. What to do about @danielj7 's comment about not needing to C/R libidn and libidn2? There's no file collision, but it would be more deterministic when building to not have them both around in case something can use one or the other.

bcully:
mutt doesn't build with libidn2 yet. Is there some action you need though?

nieder:
No action is needed by others for this PR. I should have split this up into two issues: 1 for the changes I made, and 2 for tracking the remaining packages.

dmacks:
The closest other libraries set I know is lcms/lcms2, which do not C/R each other. I remember several cases where fallback detection of the different versions made dependency and ./configure control difficult because packages had to ignore the undesired one.

dmacks:
midori indeed does not use libidn/libidn2. It needs "some IDN support" but only falls back to looking for actual libidn that if it doesn't find libsoup>=2.28, which is listed as a dep.

dmacks:
In the gsasl9 update (8301990), you used the token `# FTBFS LIBIDN2`, but the other tagging (8b17df2) you use `# LIBIDN2 FTBFS`. The latter is probably more portable and makes it easier to grep *.info.

dmacks:
For gsasl, a better link is http://lists.gnu.org/archive/html/help-gsasl/2018-01/msg00000.html (upstream bug report to gsasl, which also links to libidn's migration-info page)

dmacks:
If gnutls updates to be >=3.4, it might have onboard idn and not need external at all. See https://dev.solus-project.com/T3736#66725

clepple:
> Does github have a way to set a group-writeable tasklist so we can all track all the libidn progress?

@dmacks the [Github-flavored Markdown task list](https://guides.github.com/features/mastering-markdown/) on a separate issue seems pretty useful, but writability (and therefore the ability to check off tasks) is probably limited to the set of maintainers who can edit comments on issues. Putting a link to a tracking issue into a commit message helps track things across peoples' forks and branches.

PR #32, opened 2018-01-05
Updates for rpy2-py building against R3.4

### LINKED ISSUE #17: R version support

dhomeier:
Update to 2.8.6 (last version with Python 2.7 support) and 2.9.1
(for the Python 3.x types); upgraded R dependencies to 3.4 to help resolve #17.
Tested and built on 10.12 and 10.13; note that some of the tests use pandas-py*, if present.
The 2.9.1 version (`rpy2-py.info`) may need an update of `cran-dplyr-r34` to 0.7.4 (at least for the tests), but I haven't upped the Depends yet.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/tzlocal-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/tzlocal-py.info
@@ -1,0 +1,50 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info3: <<
+
+Package: tzlocal-py%type_pkg[python]
+Version: 1.5.1
+
+Revision: 1
+Homepage: https://github.com/regebro/tzlocal
+Maintainer: Derek Homeier <dhomeie@gwdg.de>
+Type: python (2.7 3.4 3.5 3.6)
+Depends: python%type_pkg[python], pytz-py%type_pkg[python], setuptools-tng-py%type_pkg[python]
+BuildDepends: setuptools-tng-py%type_pkg[python]
+
+Source: https://pypi.io/packages/source/t/tzlocal/tzlocal-%v.tar.gz
+Source-MD5: 4553be891efa0812c4adfb0c6e818eec
+Source-Checksum: SHA256(4ebeb848845ac898da6519b9b31879cf13b6626f7184c496037b818e238f2c4e)
+
+CompileScript: python%type_raw[python] setup.py build 
+
+InstallScript: python%type_raw[python] setup.py install --root=%d
+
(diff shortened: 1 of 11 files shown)


dmacks:
The version-bump and migration to r34 in rpy2-py.info looks fine. Is the goal of the new rpy2-py-2.8.6.info file because the new version (2.9.1) does not support python27, so we keep the last 2.8.x for older python? If that's the situation, is it useful to still have python3x as variants for it, which would be masked by the newer .info?

dhomeier:
Yes, `2.8.6` is required for python27-compatibility; I just kept the py3x variants in case anyone wanted the exact same version for some reason, but there is probably no real use case (and I guess for e.g. `apt-get install rpy2-py36=2.8.6-1` to work one would have to incrementally update first to 2.8.6 and then to 2.9.x for the package to be built in the first place)...
I submitted another update to the now-current 2.9.2 from upstream; found that the tests require a bunch of updates to various R modules, but with those in place, should all succeed (see `DescPort`, tested on 10.12 and 10.13).

dhomeier:
@dmacks, I've updated `rpy2` again to the current upstream and added a new package now required by its `pandas` loader. See `DescPort` for remaining issues when running the tests.

dmacks:
Some of the cran mods look like they have incorrect Distribution tags for their langversion types. If you supply -r31, you need to make sure it's only for 10.9-10.12 (not 10.13). In a few of your commits, you removed the Dist tags for r31, which makes it visible on all platforms (even 10.13).

nieder:
I cherry picked the new-to-fink tzlocal-py and cran-dbplyr-r commits from here into the main tree. All the other included deps (glue, purrr, etc) had been updated independently past what this PR did so they all have git conflicts. rpy2-py also has git conflicts. I'm going to close this because it's going to be a mess to fix the git tangle and it'll be cleaner and easier to understand to go with a fresh start at updating rpy2-py.

PR #35, opened 2018-01-12
Updates gmt-4.5.17 and gmt5-5.4.3, plus change of maintainer (Kurt to…

remkos:
… Remko)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/gmt.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/gmt.info
@@ -1,13 +1,13 @@
 Package: gmt
-Version: 4.5.16
+Version: 4.5.17
 Revision: 1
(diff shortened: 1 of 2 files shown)


nieder:
Can gmt/gmt5 work with the package libpcre2, or does it still need libpcre1?

remkos:
GMT code could be made compatible with libpcre2, but it is coded with the older API (of libpcre1). From the (sparse) libpcre web pages, I do not think libpcre2 has any backward compatibility build in, so it would require adjusting the libpcre calls.
(As one of the maintainers of GMT, I can certainly look into that).

remkos:
By the way, gmt5 depends on libpcre1. The package gmt does not depend on libpcre1.

nieder:
If it needs recoding to work with pcre2, it doesn't matter for this submission then.

PR #36, opened 2018-01-14
libvpx/liboil/libwebp version upgrades

dmacks:
@nieder, a few libversion-dependency upgrades for your packages.
@danielj7, when this gets merged, libvpx15 (old libversion static-only) will become completely unused.


dmacks:
Makes sense. I know a few packages also directly used liboil and then scrapped that or switched to liborc. Lots of obsolete IBD in the repos.

PR #37, opened 2018-01-16
Webp cleanup

dmacks:
This migrates the last few packages using libwebp6 to use libwebp7:
* @fingolfin: sdl-image
* @nieder: two imagemagick suites and kde4-runtime-mac
* @schwehr: gdal

Once this is merged, @danielj7: libwebp6 will be unused and can be stubbed back to a -shlibs like the other old libversions of it.


dmacks:
These are fairly small and I test-built them all, so will push (yay issue-tracker linking so maintainers can see what happened!)

PR #38, opened 2018-01-17
pascal-p5: update to new upstream version (1.4)

kamischi:
Subject says it all. At this occasion, many thanks to nieder for his detailed description of the git workflow. Finally, I seem to got it right.


nieder:
Upstream source is 404 and it didn't make it to our mirrors. Sourceforge only caches code snapshots for a very short time after you manually create/download one from the website.

PR #42, opened 2018-01-18
Remove Depends:pkgconfig that is no longer needed

dmacks:
We used to need this when .pc used some new syntax to assure a
pkg-config that processed it properly, but no currently supported
distro has ever had a version of pkg-config older than the one that
supports it. Current practice is not to have this dependency solely
because there is a .pc file, and there are other programs now that
could be used instead. Only retain in cases where a program in the
.deb actually runs the pkg-config program.

Affected package-sets:
* @bcully: fuse4x-dev, osxfuse
* @danielj7: libpar.0-dev, libaprutil.0-dev, neon27
* @kamischi: lhasa0-dev, protobuf-c-dev, protobuf7-dev
* Martin Costabel: lcms, lcms2
* @mommsen: clamav8-dev

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/lcms.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/lcms.info
@@ -1,12 +1,12 @@
 Package: lcms
 Version: 1.19
-Revision: 1
+Revision: 2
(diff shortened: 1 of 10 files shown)


dmacks:
I pushed the ones that were not done elsewhere.

PR #43, opened 2018-01-18
protobuf, protobuf-c, lhasa: removed Deps: pkgconfig

kamischi:
Remove dependence on pkgconfig as proposed by dmacks.


dmacks:
Changing Depends changes the .deb, so you need to increase the Revision. See the commit at Issue #42 if you'd like to handle your own separately (if nobody does their own or objects, I'll just push that whole one in a week or so).

kamischi:
I am fine with you doing the push in a week.

PR #44, opened 2018-01-18
Pascal-P5: Change source to a permanent server.

kamischi:
Thanks to nieder for pointing out the problem with snapshots from sourceforge.


nieder:
It looks like the mirrors eventually circled back and was able to download it after I triggered the cache again this morning.

Our SF page has a snapshot folder for stuff like this. In the future you can host snapshots there or on your own site like you're doing now.
https://sourceforge.net/projects/fink/files/miscellaneous/snapshots/

PR #45, opened 2018-01-21
Update pythia8 to latest version

mommsen:
Update to latest upstream version.


mommsen:
@nieder. Thanks for merging the PR. As long as there is only one PR at the time, it will not cause any issues.

PR #46, opened 2018-01-21
mutt 1.9.2

bcully:
Update mutt to 1.9.2


bcully:
I'll add the LIBIDN2 comment and the pointer to the mutt ticket. I'm hesitant about ispell, it's kind of a needless dependency. All the configure script does with it is set the default path for the $ispell config variable, which is trivially overridden at run time.

dmacks:
`ConfigureParams:ac_cv_path_ISPELL=no` forces non-detection even if user has it installed, so at least everyone's binary would behave the same way (having to set ispell variable if desired) at runtime.

bcully:
I've added the config param you mentioned, and the libidn2 comment and pointer.

PR #49, opened 2018-01-27
Update to gcc 7.3.0

jwhowarth:
Tested on 10.13 with 'fink -m' for against both Xcode 6.2 and 6.3 beta.


jwhowarth:
Note that gcc 7.3.0 contains Spectre V2 mitigation support.

PR #51, opened 2018-01-29
rustc 1.23.0

bcully:


dmacks:
Are those shared libraries actually for public linking by other compiled programs, or are they merely shared among various binaries that are in the rustc package itself? They all appear to have changed names, so if other programs or packages' binaries had been compiled previously, will this change break them or require recompiling?

bcully:
I don't know exactly what those libraries are for, but I do know that none of them are linked into anything I've built. I believe they are just shared for the benefit of all the tools that the compiler ships with, based on the names of the libraries and the fact that the tools link to them but nothing else appears to.

I wish there were a way to get fink to ignore these shared libraries, honestly.

dmacks:
Sounds like a perfect example of "Private Shared Libraries", which have a slight variation on the Shlibs field syntax.

dmacks:
Less data for you to update each time and less confusing for others using it, but still pain-ish.

bcully:
Updated to use private shlibs

PR #52, opened 2018-01-30
kotlin 1.2.21

bcully:


dmacks:
Does this thing need some sort of dependency on java? I don't know what is standard on OS X...can we just assume that everyone will have "some java" installed and "any java" is good enough?

nieder:
https://github.com/JetBrains/kotlin says you need `JDK 1.6, 1.7 and 1.8` to _build_ the kotlin distribution, which seems to suggest that any of those javas should be good enough for runtime. And this [release note](https://blog.jetbrains.com/kotlin/2017/08/kotlin-1-1-4-is-out/) suggests there's java 9 support.

So `RuntimeDepends: system-java` should cover things?

bcully:
ack, thanks @nieder I'll do that.

PR #55, opened 2018-02-03
viennarna: Separate .info and revision for each different platform-perl version

dmacks:
The Type/Distribution was attempting to tie the build results to the
perl version (based on the OS X version) but the result was that the
same %n-%v-%r would use different installation pathnames for different
build-systems. That breaks policy. Instead, separate .info for each
perlversion, so each can have different %r, with bonus of simplifying
the Type and Distribution fields. Now when user moves to newer OS X,
the package self-identifies as needing to be updated.


dmacks:
It's been almost a month and this is a packaging policy detail. Pushing...

PR #56, opened 2018-02-04
update gdb.info to 8.1 release

jwhowarth:
Tested on 10.13.3 with 'fink -m' against Xcode 9.2.


nieder:
@jwhowarth 
build failure on 10.11: https://sourceforge.net/p/fink/bugs/1665/

PR #60, opened 2018-02-04
python-ldap-py dependency cleanup

dmacks:
Remove a bunch of seemingly-unused dependencies (instrumented the
build with f-p-p to help diagnose). Possibly these are long-obsolete
inherited build-depends of the only directly used library or lingering
from older upstream version that used these directly as well.


dmacks:
@clepple this is yours

clepple:
@dmacks I was about to put in a pull request for an updated upstream version - I'll re-test with this.

PR #61, opened 2018-02-06
python-ldap-py27 new upstream 2.4.28

clepple:
Includes #60.


clepple:
Hmm, setup.cfg searches for includes/libraries in `/usr/local`. Let me know if we should patch setup.cfg to guard against that, or deal with breakage if it comes up. (FAQ tells users to move /usr/local out of the way if needed, so they have at least been warned)

dmacks:
Probably best to clear that out. Nicer to pre-empt if it's easy to do so; from working the help-channels, users are often mystified by the /usr/local situation.

clepple:
@dmacks no problem - see commit 11bd11e.

clepple:
@dmacks thanks. What's the next step? (I don't have write access to the git repo)

dmacks:
I did the pull, and also asked the other core folks what our process is for getting you write access.

PR #62, opened 2018-02-06
xrootd4: new version and misc packaging fixes

dmacks:
Instrument the build with f-p-p, update (Build)Depends and cmake flags
to get consistent use of fink's libraries (some already missing in old
package, some newly needed in new version).

Resolve some .dylib/.so mixups (NB: some public Shlibs entries removed
compared to previous package; they should have always been private and
.so--not usable by linking by others anyway).

Switch to openssl110.

Adjust license per LICENSE and include licenses in DocFiles

Remove unused (since 4.0.0) perl flags, which haven't been the correct
paths for recent OS X versions anyway.


mommsen:
Superseded by pull request #77.

PR #65, opened 2018-02-10
tgif: fix libidn usage (dlopen needs path) and LIBIDN2 FTBFS

dmacks:


10.9-libcxx/stable/main/finkinfo/graphics/tgif.info
```diff
@@ -76,7 +83,11 @@ Japanese charset patch from Tomoaki Okayama.
 (NLS support priginally suggested by Asari Takashi)
 Original version by Jeff Whitaker
 
-Hopefully ensure libXt stability with versioned XQuartz dependency and avoid libxt*.
+Hopefully ensure libXt stability with versioned XQuartz dependency and
+avoid libxt*.
+
+Loads libidn (for stringprep_convert(), which is not in libidn2) at
+runtime. Q License is incompatible with linking agaist GPLish library.
```
akhansen: Upstream offers a "free-of-charge" license version, too.  Wonder if that's compatible.

dmacks: No idea. But they also don't have a way to do the linking (it's dlopen or disabled altogether). Bigger question is why they haul in libidn for just one non-IDN-related function.

PR #66, opened 2018-02-10
whois: autodetects libidn2 via .pc in preference to libidn

dmacks:


dmacks:
It's been ~3 months, and this is just dep cleanups, so...

PR #70, opened 2018-02-13
djview: new version, dep cleanup

dmacks:


dmacks:
The added PatchScript inserts /opt/X11 into the list of places to search for x11. That patch (maybe with a different list-ordering for the search) is the one that newer autoconf has, which I had tested on a "no /usr/X11*" system.

PR #71, opened 2018-02-14
cargo 0.23.0. Fixes FTBFS against rustc 1.23.0

bcully:


dmacks:
Confirming previous version FTBFS on 10.13 whereas this commit fixes it. This package is on my list of untriaged ones still uses openssl100; I tried switching the deps to openssl110 and got identical build logs...maybe do so as part of this update?

PR #80, opened 2018-02-28
astng-py3x: call '2to3' with fink prefix and match pythonvariant

dmacks:
This fixes FTBFS on -py3x and has no effect on -py2x. See http://bindist.finkproject.org/10.13/logs/astng-py34_0.24.3-1_71c13179f47da313f6035b043b5b50c3.log


nieder:
This belongs to @schwehr

PR #82, opened 2018-03-03
Pyflakes and associated pymod updates

nieder:
@danielj7 these are needed for an update to another package. All tested with py27 and py36.


danielj7:
Committed with some changes

PR #83, opened 2018-03-04
Turn off -pm5162 variants on 10.13

dmacks:
Pursuant to Issue #6


dmacks:
Okay to defer this for now; it's been months and nobody has a workable solution for texi2html on 10.13 except to activate perl5162 there, so less urgency for removing *-pm5162 (no longer a dep-tree hole).

nieder:
Should this be closed as wontfix since we're no closer to finding a solution?

dmacks:
It's BLOCKED-BY whatever issue/pr would fix/update texi2html (we don't have one specifically titled for it, but it's what #6 has morphed into).

dmacks:
meh

PR #84, opened 2018-03-04
optipng: new version

dmacks:
Fink's current 0.7.6 fails to build on 10.13 for me and also in the [bindist for 10.13](http://bindist.finkproject.org/10.13/logs/optipng_0.7.6-1_9ceab30cd68c7fb77253fc36d942c5b6.log). I don't see an error in the 10.11/10.12 bindist. Upstream has a [bug report about it](https://sourceforge.net/p/optipng/patches/9/) that says 10.13 being different than older versions, but that 10.12 would be like 10.13?

The attached upgrades to the new 0.7.7 upstream (which also has some major security fixes), and builds for me on 10.13. Probably should get someone with 10.12 (I don't have it), and maybe something older to test-build, or else have this specifically tagged for 10.13 only for now.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/optipng.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/optipng.info
@@ -15,11 +15,11 @@
 # maintainership of) the affected package.
 
 Package: optipng
-Version: 0.7.6
+Version: 0.7.7
 Revision: 1
 Maintainer: Max Horn <max@quendi.de>
 Source: mirror:sourceforge:%n/%n-%v.tar.gz
-Source-MD5: 568e0738358450eca69ecf578d48f26c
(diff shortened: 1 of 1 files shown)


nieder:
This built for me on 10.11.

Is it worth building against either Fink libpng or system-zlib? For 0.7.7, internal libpng is 1.6.34, same as latest upstream and in Fink, but 0.7.6 had 1.6.21, which was quite old. On the other hand, on 10.11, system-zlib is only 1.2.5 vs 1.2.8 and 1.2.11 in the two latest optipng releases. We track upstream libpng across supported platforms. macOS poorly tracks upstream zlib.

dmacks:
This has FTBFS 5 months. Improvements can be discussed elsewhere.

PR #89, opened 2018-03-06
New upstream version of rkhunter

mommsen:
Update rkhunter to version 1.4.6

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/rkhunter.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/rkhunter.info
@@ -1,6 +1,6 @@
 Package: rkhunter
 Version: 1.4.6
-Revision: 11
+Revision: 1
 Description: Rootkit Hunter
 License: GPL
 Maintainer: Remi Mommsen <remigius.mommsen@cern.ch>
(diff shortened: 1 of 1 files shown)


nieder:
The build has a lot of lines like:
```
chown: ./files: Operation not permitted
```
Is this expected?

mommsen:
Yes, the build scripts is changing the file permissions to root, which fink-bld is not allowed to. During the installation step (run as root), this is rectified.

dmacks:
Teeny nit: usually good practice to reset Revision to 1 when upgrading Version.
Possible improvement: fink has several variants of wget, but this package has a dependency only on one of them. Users might appreciate being able to use a different one (the differences seem mostly internal, so I assume other variants would suffice). Instead of ```wget```, use ```wget | wget-gnutls | wget-idn | wget-gnutls-idn```

dmacks:
Probably best to patch install.sh to omit the chmod. No sense giving diagnostics that look like a problem (a non-detected error leading to a misbuilt package?) that aren't.

mommsen:
At some point in the past, we started using higher revisions for the latest tree in order to be able to pack port packages to older trees. I don't mind to reset the revision to 1 as there are no older trees maintained anymore.

The wget package indeed provides different variants, but also defines the virtual package gnu-wget. Thus, I think it would be enough to depend on gnu-wget, isn't it?

I'll see if I can patch the install.sh. IIRC, the same method is used during the compile and install cycle. The chmod is needed in the latter. I'll see if I can patch it. However, nobody has complained about it since fink started using fink-bld by default years ago....

mommsen:
@dmacks, can you please check if your concerns have been addressed by commit 5b2c15f?

Thanks,
Remi

mommsen:
ping @dmacks (;

PR #92, opened 2018-03-06
update pyephem-py to latest upstream version. 

khorton:
update pyephem-py to latest upstream version.  Tested on macOS 10.13 with python 2.7 & 3.4.  Builds with fink -m


nieder:
This commit 4bfe51 seems to be a partial duplicate of what's found in PR #91. Plus the .info created here has an extra `<<` at the end and fails validation.

PR #102, opened 2018-03-14
dosbox: fix FTBFS by using freebsd patch

dmacks:
...and also resolve a few compiler warnings. Former .patch was unused.


dmacks:
2 months no response, let's clear this FTBFS. Easy enough to continue/change the solution later if  @fingolfin wants

PR #103, opened 2018-03-14
update  postgresql94 9.4.17 and  postgresql95 9.5.12

wodev:
update postgresql95 to version 9.5.12
update postgresql94 to version 9.4.17

changes are tested with fink -m


dmacks:
Any thoughts about having postgresql*.info use a variant to minimize diffs among the versions?

wodev:
@nieder 
Perhaps someone still stays on 9.4 or 9.5 so I updated also the 9.4 and 9.5 Version.  They  are supported  until 2019 / 2020 .  

@dmacks 
I will check the use of variants with the Postgresql 10 Version which I currently preparing.

PR #104, opened 2018-03-14
tora: switch pg94->pg96; improve f-p-p use

dmacks:
Same build transcript for the pg update.

f-p-p had a ton of warnings about files not recognized by dpkg and f-p-p didn't recognize because they were source-distro but in a sister subdir to the one where it was running. f-p-p isn't so smart about out-of-source or even build-subdir setups, so just run it from the main builddir.


dmacks:
There is a newer version upstream, but I didn't look at it because "this built for me" and my current itch is just the pg zoo:)

PR #105, opened 2018-03-15
gdcm26 2.6.9

nieder:
Now works with java9.
The current v2.6.1 release needs java5, which is incompatible with java9. v2.6.9 uses java7.
It also now builds the manpages, but can be disabled w/ a simple configure flag.
This is the last 2.6.x release.


akhansen:
I'm OK with this.

PR #108, opened 2018-03-16
A2ps sec

nieder:
Add Debian security patches
activate tests


dmacks:
While you're cleaning up DescPort, "Libtool and config files copied to auxdir directory." looks obsolete.

Switching to:
```
InstallScript: make install DESTDIR=%d
```
means the --mandir= and --infodir= flags can be scrapped in ConfigureParams.

I tried removing the ```ac_cv_exeext=``` setting, and got identical build transcript and no obvious difference in the resulting .deb contents.

dmacks:
It's been 1.5 years. I say push this security fix. Debian's patchset is now bumped to -4, but there are no relevant changes in that packaging upgrade of theirs.

PR #110, opened 2018-03-16
ppl13: glpk->glpk36

dmacks:


nieder:
This built for me on 10.11. @fangism told J. Howarth some time ago that we could work on his packages.

dmacks:
Nobody's objected in 6 weeks, so...

Issue #111, opened 2018-03-18
package-deprecationmanager-pm will cause circular dependencies


nieder:
I'm building something that eventually needs package-deprecationmanager-pmXXXX in its dep tree. Currently, it fails due to Package::Stash not being present:
```perl
# The following REQUIRED prerequisites were not satisfied:
# 
# Package::Stash is not installed (any version required)
# 
t/00-report-prereqs.t .. ok   
t/basic.t .............. Can't locate Package/Stash.pm in @INC (@INC contains: /sw/build.build/package-deprecationmanager-pm5162-0.17-3/Package-DeprecationManager-0.17/blib/lib /sw/build.build/package-deprecationmanager-pm5162-0.17-3/Package-DeprecationManager-0.17/blib/arch /sw/lib/perl5/5.16.2/darwin-thread-multi-2level /sw/lib/perl5/5.16.2/darwin-thread-multi-2level /sw/lib/perl5/5.16.2 /sw/lib/perl5/darwin-thread-multi-2level /sw/lib/perl5/5.16.2/darwin-thread-multi-2level /sw/lib/perl5/5.16.2 /sw/lib/perl5/darwin-thread-multi-2level /sw/lib/perl5 /sw/lib/perl5/darwin /sw/lib/perl5/5.16.2/darwin-thread-multi-2level /sw/lib/perl5/5.16.2 /sw/lib/perl5/darwin-thread-multi-2level /sw/lib/perl5 /sw/lib/perl5/darwin /sw/lib/perl5/site_perl/5.16.2/darwin-thread-multi-2level /sw/lib/perl5/site_perl/5.16.2 /sw/lib/perl5-core/5.16.2/darwin-thread-multi-2level /sw/lib/perl5-core/5.16.2 .) at /sw/build.build/package-deprecationmanager-pm5162-0.17-3/Package-DeprecationManager-0.17/blib/lib/Package/DeprecationManager.pm line 10.
```
If I add package-stash-pmXXXX to the `[Depends]`(http://deps.cpantesters.org/?module=Package%3A%3ADeprecationManager;perl=latest) field for package-deprecationmanager-pmXXXX, then I get a circular dependency warning by Fink. The circle includes just these two packages. But skimming through the package-stash source, I don't see Package::DeprecationManager used any more (0.21 tarball did).

Also, package-deprecationmanager needs TestDepends on test-warnings-pmXXXX. I can commit these 3 dependency changes if you want.

nieder:
Fixed by #112

PR #112, opened 2018-03-18
Perlmod updates

nieder:
updates for a bunch of perlmods.
Most belong to @dmacks 
Includes fix for #111


10.9-libcxx/stable/main/finkinfo/libs/perlmods/yaml-tiny-pm.info
```diff
@@ -13,8 +13,7 @@ BuildDepends: <<
 	perl%type_pkg[systemperl]-core
 <<
 Source: mirror:cpan:modules/by-module/YAML/YAML-Tiny-%v.tar.gz
-Source-MD5: 9fd7940c05b025d6be08217c9a5061b6
-DefaultScript: ModuleBuild
```
dmacks: Now that it's using Makefile.PL not Build.PL and I don't see any Module::Build residue in Makefile.PL or META.yml, can it drop ```BuildDepends:module-build-tiny-pmXXX``` and therefore drop ```BuildDepends:perlXXX-core``` and therefore drop the whole systemperl/platform-dependent varianting (unfying yaml-tiny-pm-10.9.info back into yaml-tiny-pm.info)?

dmacks:
3b0fd54 and c899aa5 lgtm by eye (except for the one improvement I mentioned to yaml-tiny) as simple version updates. I assume you checked that there are no added/removed deps needed.

10.9-libcxx/stable/main/finkinfo/libs/perlmods/string-diff-pm.info
```diff
@@ -1,19 +1,34 @@
 Info2: <<
 Package: string-diff-pm%type_pkg[perl]
-Version: 0.06
-Revision: 2
+Version: 0.07
+Revision: 1
 Depends: <<
 	algorithm-diff-pm,
-	perl%type_pkg[perl]-core
+	module-install-githubmeta-pm%type_pkg[perl],
```
dmacks: Sucks that such a simple pure-perl module has such a large and perlvarianted dependency chain. Are these all really Depends not BuildDepends? By grepping, I only see "Module::" in Makefile.PL and related places, not the library itself. I don't even see *anything* that is perlversion-related in the runtime (neither in the existing 0.06 version by comparison), just a TestDepends, so maybe it could become a build-varianted rather than runtime-varianted.

I can redo the packaging separately for that if you want to get the rest of this PR merged.

This package is Jesse Alama's...let's see if @jessealama is the right github account.

nieder:
I generally tested with both purging installed perlmods as well as with deps.cpantesters.org to make sure we weren't missing deps (several found this way).

10.9-libcxx/stable/main/finkinfo/libs/perlmods/yaml-tiny-pm.info
```diff
@@ -3,14 +3,11 @@ Package: yaml-tiny-pm
...
-Distribution: 10.10, 10.11, 10.12, 10.13
-Type: perl, systemperl (5.18.2)
+Type: perl
 License: Artistic/GPL
 Maintainer: Benjamin Reed <yaml-tiny-pm@fink.raccoonfink.com>
 BuildDepends: <<
-	fink (>= 0.30.2),
-	module-build-tiny-pm%type_pkg[systemperl] (>= 0.027),
-	perl%type_pkg[systemperl]-core
+	fink (>= 0.30.2)
```
dmacks: Can probably even drop the BDep:fink (remnant of DefaultScript?) and the Info2: wrapper (remnant of varianting and part of RangerRick's template).

10.9-libcxx/stable/main/finkinfo/libs/perlmods/string-diff-pm.info
nieder: module-install-githubmeta, M::I::readmefrompod, and M::I::readmemarkdownfrompod all Dep on module-install, which is varianted. Based on the deb filelist, I don't think they have anything that links/requires a specific perlversion, so they could conceivably be unvarianted, but would need a separate .info for 10.9.

dmacks: I'm not proposing unvarianting those packages--as you say, they have legitimate Depends on module-install-pmXXX, which is legitimately varianted. I'm just wondering about string-diff-pmXXX itself.

jessealama: @dmacks I'm pretty sure they're just BDeps.

10.9-libcxx/stable/main/finkinfo/libs/perlmods/yaml-tiny-pm.info
jessealama: You're welcome to drop the BDep:fink and Info2 wrapper (you're right that I borrowed them).

10.9-libcxx/stable/main/finkinfo/libs/perlmods/string-diff-pm.info
nieder: If all 4 M::I::* current Deps are really BDeps, then perl-core can also move to BDep. But that still leaves string-diff as a varianted package unless the usual systemperl 10.9 split is done, right?

dmacks: Right. ```Package:``` could be unvarianted and same ```Revision:``` on all platforms, but a systemperl *build* variant separate for 10.9.

nieder:
Will give @danielj7 a couple days to respond since 2 of these are his that required a version bump to break circular deps (#111 )

dmacks:
Do it.

jessealama:
thanks

PR #115, opened 2018-03-21
libdbi upgrade

dmacks:


dmacks:
This is all Dave Reiser's. Anyone know if he's on github?

nieder:
Do the tests for the new (or old) libdbi-drivers-pgsql94 work?  They hang here for me with the old version.

dmacks:
They had never worked for me IIRC. Can't remember if they hang or required that the package be already installed, or required some other setup of system servers.

nieder:
@dbreiser

nieder:
These look OK to me and built on 10.13.
For the C/R issue, upstream has all of 2 commits since 2014, so I doubt it's going to get any love for newer pg versions? So just ignore resolving that aspect unless it becomes necessary?

dmacks:
It's been a long time with no action and no objection.

PR #116, opened 2018-03-22
pil-py: Install headers #included by own Imaging.h.

nieder:
Debian bug report: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=879788

Imaging.h #includes ImagingUtils.h, so it needs to be installed. This is Debian's fix:
````
	install -o root -g root -m 644 \
		src/libImaging/Imaging.h \
		src/libImaging/ImagingUtils.h \
		src/libImaging/ImPlatform.h \
		src/libImaging/ImDib.h \
		debian/python-pil/usr/include/python$*
````
which I adapted.


dmacks:
There's a newer [upstream release](https://github.com/python-pillow/Pillow/releases), but I don't know if that includes a fix. But browsing setup.py I see a lot of other suspicious details (using pkg-config, assuming fink is in /sw, autodetecting homebrew, etc)

PR #117, opened 2018-03-24
New package postgresql10

wodev:
I created a new package for the latest postgresql major release Postgresql 10.
Package tested with fink -m build

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
@@ -254,75 +254,26 @@ SplitOff2: <<
 <<
 
 PreInstScript: <<
-PG_ID=`id -u postgres 2>/dev/null || true`
-if [ -z "$PG_ID" ]; then
+   PG_ID=`id -u postgres 2>/dev/null || true`
+   if [ -z "$PG_ID" ]; then
 	echo "Whoa there!  You must have a postgres user to install this package."
 	echo "Please do a 'fink reinstall passwd-postgres' and make sure you hit"
 	echo "'y' when it asks if you want to update your users."
 	exit 1
-fi
+   fi
 <<
 PostInstScript: <<
-INSTALL_PHASE="$1"
-
-[ -x %p/var/postgresql-%type_raw[postgresql]/update-alternatives.sh ] && %p/var/postgresql-%type_raw[postgresql]/update-alternatives.sh
-
-# remove the old "pgsql" entries from netinfo; the username was switched to
-# "postgres" but the old ones hang around because of the way niload works
-niutil -destroy . /users/pgsql  >/dev/null 2>&1 || true
-niutil -destroy . /groups/pgsql >/dev/null 2>&1 || true
-
-die () {
-	echo "failed"
-	echo ""
-	echo "*** bailing because an error ocurred:"
-	echo ""
-	echo "$*"
-	exit 1
-}
-
-# update daemonic init script if necessary
-daemonic install %N >/dev/null 2>&1 || :
-
-# get a nice port to run on
-while true; do
-	PGPORT=$RANDOM;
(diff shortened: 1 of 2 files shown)


dmacks:
I think there's a logic flaw in the ```update-alternatives --install``` commands. Previously, you had been using 94, 95, 96 as the "priority" value. These values, which are the dotless strings of postgresql versions 9.4, 9.5, and 9.6, respectively, allow certain files from newer postgresql versions to be seen in preference to older ones via symlinks from certain "unversioned" filenames. Now, however, we have postgresql 10, whose dotless string (now via ```type_pkg``` where ``type_raw``` is "10") is less than 94. That means users would no longer be seeing "the newest installed version" by default from the unversioned-filename symlinks. By definition, priority values are integers.

I wonder if the current claim of this package as "postgresql10" (compared to others as "postgresql96" for version "9.6") is actually following upstream in making an implicit assumption that there will never be a 10.1. It's already pattern-breaking (let alone a u-a priority problem) to go "94 95 96 10", but then we would go "94 95 96 10 101"? Maybe better to call this here "postgresql100" (type_raw 10.0), which would mean the u-a priority of type_pkg=100 is properly in order. I'm not sure it's worth the effory to allocate two digits for the minor version in order o future-proof (in case we get as far as "9.x after 9.9" or similarly deep into the 10.x series).

10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
```diff
@@ -0,0 +1,318 @@
...
+	find contrib -name README.\* -exec cp {} %i/share/doc/%N/ \;
+
+	install -d -m 755 %i/bin
+	install -c -m 755 pgsql.sh %i/bin/pgsql.sh-%type_raw[postgresql]
+
+	install -d -m 755 %i/var/postgresql-%type_raw[postgresql]
+	echo "be sure to back up this database before any upgrades!" >> %i/var/postgresql-%type_raw[postgresql]/README
+
+	install -d -m 755 %i/var/log
+	ln -sf %p/var/postgresql--%type_raw[postgresql]/pgsql.log %i/var/log/pgsql-%type_raw[postgresql].log
```
dmacks: Doubled hypen in the ```ln``` here?

wodev:
Thanks for pointing out the update-alternatives priority issue.

With major release 10 the Postgresql Team changed the versioning scheme. Beginning with version 10 the second part indicates a minor release and a major release is indicated by increasing the first part of the version, e.g. 10 to 11.
[https://www.postgresql.org/support/versioning/](https://www.postgresql.org/support/versioning/)

For that reason I named the package posgresql10 , like other package distributers (freebsd ports or mac ports). In my latest commit is a correction in the update-alternatives logic, so that the new postgresql versions 10, 11 ... are higher prioritized as the old 9.x major versions.

Please check the package again.
Thanks.

dmacks:
Thanks for explaining the versioning! If they go 9.4,9.5,9.6,10,11, the priority is always "```type_raw```*10".

wodev:
Correct, for  9.4, 9.5 and 9.6 the priority is "type_raw" and for the version in the new schema (10, 11,  ... )  the priority is "type_raw*10". 

Also the the "doubled hypen" typos from the last review is corrected with commit "420aed6"

dmacks:
If 'the priority is "type_raw*10', you could just code it that way in all .info (here, and also if the 9x become varianted), and not need to special-case for MAJOR_C1=9.

wodev:
Thanks for the hint.
latest commit simplifies the ua priority calculation.

wodev:
@dmacks could you please approve the PR or let me know if there are any changes necessary

wodev:
Commit 73ea833 updates to latest upstream version 10.4
Please review and approve the PR or let me know if there are any changes necessary , Thanks

PR #124, opened 2018-04-04
coot: version update (minor)

wgscott:


nieder:
No BuildDep on libidn2.0-dev to match the -shlibs?

nieder:
I see no evidence of libidn2.0 being used by coot. Release tarball has no link command to libidn2.0.dylib, no usage of idn2.h, etc. Upstream's build helper script says it uses libidn for libcurl because RHEL4 (!) doesn't provide libidn.so. So seems to be an outdated reference to an inherited dep.

nieder:
Merged in with #561

Issue #127, opened 2018-04-11
moose-pm51*2 test failure


nieder:
On 10.11, moose-pm 2.1600-1 fails tests as follows. Current moose upstream is 2.2010.

````
t/type_constraints/util_type_coercion.t .......................... Semicolon seems to be missing at t/type_constraints/util_type_coercion.t line 20.
syntax error at t/type_constraints/util_type_coercion.t line 21, near "coerce "
Execution of t/type_constraints/util_type_coercion.t aborted due to compilation errors.
t/type_constraints/util_type_coercion.t .......................... Dubious, test returned 255 (wstat 65280, 0xff00)
No subtests run 
t/type_constraints/util_type_constraints.t ....................... ok     
t/type_constraints/util_type_constraints_export.t ................ ok   
t/type_constraints/util_type_reloading.t ......................... ok   
t/type_constraints/with-specio.t ................................. 
    #   Failed test '[ 42.4, 84 ] is an acceptable array of ints'
    #   at t/type_constraints/with-specio.t line 67.
    #                   'Attribute (array_of_ints) does not pass the type constraint because: Validation failed for type named ArrayRef[Int] declared in package Foo (t/type_constraints/with-specio.t) at line 29 with value Specio::Constraint::Parameterized={ _ancestors: [ Specio::Constraint::Parameterizable=HASH(0x7f98... at /sw/build.build/moose-pm5182-2.1600-1/Moose-2.1600/blib/lib/Moose/Object.pm line 24
    # 	Moose::Object::new('Foo', 'array_of_ints', 'ARRAY(0x7f98a34e1848)') called at t/type_constraints/with-specio.t line 66
    # 	main::__ANON__ at /sw/lib/perl5/Test/Fatal.pm line 129
    # 	Test::Fatal::try {...}  at /sw/lib/perl5/Try/Tiny.pm line 76
    # 	eval {...} at /sw/lib/perl5/Try/Tiny.pm line 72
    # 	Try::Tiny::try('CODE(0x7f98a34cd738)', 'Try::Tiny::Catch=REF(0x7f98a45142f8)') called at /sw/lib/perl5/Test/Fatal.pm line 136
    # 	Test::Fatal::exception('CODE(0x7f98a34a2b48)') called at t/type_constraints/with-specio.t line 67
    # 	main::__ANON__ at /sw/lib/perl5/5.18.2/Test/Builder.pm line 337
[...]

PR #128, opened 2018-04-11
Update perlmods

nieder:
updates to some perlmods.
Moose fix for issue #127 is here (version update and update some deps)
@jessealama has a couple packages here too


nieder:
module-build-tiny-pmXXX is not needed anymore for class::load.

In my local branch, how do I fix the devel::stacktrace conflict? Do I make a commit that generates a file identical to the unvarianted one you checked in?

dmacks:
For pathtools-pm.info, it would be good to retain some sort of license statement in DocFiles. You could use the PatchScript one-liner from statistics-shannon-pm.info to extract it from the main .pm POD.

nieder:
bah. Using github's webbased conflict resolution adds changes from Master to the branch. Just waiting if @jessealama has any comments.

jessealama:
I took a look -- nothing weird stood out to my eyes, so I think this can be merged.

PR #140, opened 2018-04-19
Add missing variants to pythonmods

dmacks:
These patches add -py35 and -py36 to python modules that were only present in lower -py3X variants. I omitted obsolete ones, and ones that have known issues, PR filed, or are work in progress elsewhere. Is it worth backporting *-py36 to -py35 and -py34 for completeness, or do we not really care about those lower python3X any more? Should one or both be turned off?

Affected maintainers: @bcully, @danielj7, @dhomeier, @khorton, @nieder, Scott Hannahs

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/crypto/pycrypto-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/pycrypto-py.info
@@ -2,7 +2,7 @@
 Info2: <<
 
 Package: pycrypto-py%type_pkg[python]
-Type: python (2.7 3.4 3.5)
+Type: python (2.7 3.4 3.5 3.6)
 
 Version: 2.6.1
 Revision: 1
--- a/10.9-libcxx/stable/main/finkinfo/database/psycopg2-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/psycopg2-py.info
@@ -6,7 +6,7 @@ Revision: 3
 Description: PostgreSQL database adapter for Python
 License: LGPL
 Maintainer: Kurt Schwehr <goatbar@users.sourceforge.net>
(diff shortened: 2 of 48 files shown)


nieder:
typing-py is a backport to -py27 and -py34 of a module included in python35 and up. It should not need being present as -py35 or -py36 variants. It's easy to conditionally Dep on it, so I don't think we need to add py35 and py36 variants and shadow a base module.

It was packaged because newer sphinx will eventually need it.

Other changes to my pymods are OK.

danielj7:
I'm fine with it.

I would only backport from py36 if necessary for a dep hole. Py34 is the oldest version still being supported by upstream but I'd say that if there's no packages that don't work with newer versions, we should get rid of it. Same for py35. 3.7 will be out soon.

khorton:
I have no concern about this change.

Kevin Horton

nieder:
ping @dhomeier @bcully

dmacks:
bb9e529 addresses @nieder's concern. Two more commits cherry-picked from master that add variants as part of a larger set of updates.

dhomeier:
I am absolutely fine with retiring `python3.4`; a number of updates I've pushed still include `py34` variants, but I'd rather remove those in a separate changeset.
For `py35` I am still seeing 32 packages without a `py36` counterpart; although many of them already have updates under way, I vote for keeping `python3.5` around for now.

bcully:
I'm fine with this and not particularly worried about py3 < py36, but of course it's good to be consistent about what we're supporting and not needlessly have missing deps.

nieder:
What's the status here? Seems to all be in agreement?

dmacks:
Conflict resolved, so let's do this thang...

PR #142, opened 2018-04-25
wxwidgets build fixes

nieder:
* Updates wxcocoa293 to build against 10.10+.
* Updates wxcocoa295 to remove dependencies on a specific SDK.
* Bumps wxwidgets300-osxcocoa/gtk to latest upstream 3.0.4 (not 3.1.x)

Even after these changes, wxcocoa293, wxcocoa295, and wxwidgets300-cocoa still don't build on 10.13 (maybe even 10.12) due to removal of certain things (QTKit and QuickTime library) from the OS. It might theoretically be possible to patch them like wxwidgets300-osxcocoa/gtk is. However, there are only 2 dependencies on them: spatialite-gui (wxcocoa295) and grass70 (wxwidgets300-cocoa). I would propose that these 3 old libraries be dist limited to 10.11 and under. spatialite-gui can be updated to wxwidgets300-osxcocoa. Not sure about grass70, but it's not the latest version we have, so it could also be restricted alongside wxwidgets300-cocoa.

@babayoshihiko


nieder:
Alright. I'll add dist restrictions to wxcocoa293, wxcocoa295, and wxwidgets300-cocoa (and deps) when I get back to my box.

nieder:
I'm trying to finish fixing wxwidgets300-cocoa for 10.12 and 10.13. It has the `'abs' is ambiguous` problem (fixable), the `verify_noerr()` deprecation issue (fixable), the use of `QuickTime.h` (fixable), and the QTKit deprecation situation (fixable with a lot of trouble of dist specific .info files). I'm going to push with `Distribution: 10.9-10.11` and similarly restrict grass70. If grass70 wants to be made available to 10.12/10.13, it can be moved up to wxwidgets300-osxcocoa.

PR #143, opened 2018-04-26
Graphviz update

dmacks:
Continuing from https://sourceforge.net/p/fink/package-submissions/5008/, this is the extent of that ticket in the first commit and then a few more cleanups in a separate second commit. Works for me on 10.13, but build-testing on something lower would be useful.

TODO: given the php detection is different on 10.{9-12} vs 10.13, do we need separate %r for the dists?

TODO: can we hack ./configure and some Makefile.in to use external libtool2/-shlibs more consistently instead of included libltdl subdir?


nieder:
BDep: system-tcltk-dev already implies Conflicts: tcltk-dev, so no need to keep `BuildConflicts: tcltk-dev`.

php7 support requires swig-3.0.11, while this currently has `>= swig-3.0.10`.  I think php7 support is slightly different but not sure. Attached is `%i/lib/graphviz-2.38/graphviz/php/gv.php` from a php5 system. Is it different from that on a php7 system?
[gv-php5.txt](https://github.com/fink/fink-distributions/files/1953784/gv-php5.txt)

If different then it makes sense to have different %r.

Upstream also changed the location of the tarball. It's now unversioned at `https://graphviz.gitlab.io/pub/graphviz/stable/SOURCES/graphviz.tar.gz` on their main site. A versioned one lives in gitlab by their tag list `https://gitlab.com/graphviz/graphviz/-/archive/stable_release_2.40.1/graphviz-stable_release_2.40.1.tar.bz2`, but the MD5 is different. Have not tested it to see if it builds the same.

Do we need a Dep on system-java?

nieder:
Also, ConfigureScript logic doesn't know about java9 or java10. And at least for java9, the `java -version` game is failing for me.

dmacks:
The gitlab tagged item is a snapshot of their git repo, not a generated release tarball, so it doesn't contain the autotool'ed and other generated material. Filed at https://gitlab.com/graphviz/graphviz/issues/1371

dmacks:
That gv-php5.txt matches the gv.php I got on 10.13

dmacks:
I refactored the existing java version-dependent flag handling, during which I found we're not even using fink's standard -I/-L at all(?).

nieder:
24bf9d0 built for me on both 10.11 (java8) and 10.13 (java9).
````
checking for java... java
checking jni.h usability... yes
checking jni.h presence... yes
checking for jni.h... yes
````
Not sure how java9 is being detected. Maybe the internal java logic in configure got better between the older versions and 2.40.1.

**Comment by dhomeier** [comment]
24bf9d0 builds for me on 10.12 as well; on 10.13 it gets past the `javac -version` non-detection where 2.38.0 fails, but then fails to build against the system ruby installation:
```
libtool: compile:  g++ -DHAVE_CONFIG_H -I. -I../.. -I../.. -I../../lib/gvc -I../../lib/common -I../../lib/cgraph -I../../lib/cdt -I../../lib/pathplan -DDEMAND_LOADING=1 -I/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0 -I/System/Library/Frameworks/ApplicationServices.framework/Versions/A/Frameworks/CoreGraphics.framework/Versions/A/Headers -I/sw/lib/ocaml -idirafter /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0 -I/sw/include -I/sw/include -g -O2 -MT libgv_ruby_la-gv_ruby.lo -MD -MP -MF .deps/libgv_ruby_la-gv_ruby.Tpo -c gv_ruby.cpp  -fno-common -DPIC -o .libs/libgv_ruby_la-gv_ruby.o
In file included from gv_ruby.cpp:879:
In file included from /System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby.h:33:
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby/ruby.h:24:10: fatal error: 'ruby/config.h' file not found
#include "ruby/config.h"
         ^~~~~~~~~~~~~~~
```
That header is in `/System/Library/Frameworks/Ruby.framework/Versions/2.3/Headers/ruby`, while `find /System/Library/Frameworks -name ruby.h` happened to first find the headers in `.../usr/include/`.

Changing the `ruby_h_path` definition to
```
diff --git a/10.9-libcxx/stable/main/finkinfo/graphics/libgraphviz238-shlibs.info b/10.9-libcxx/stable/main/finkinfo/graphics/libgraphviz238-shlibs.info
index 872fe1c755..ca4fade670 100644
--- a/10.9-libcxx/stable/main/finkinfo/graphics/libgraphviz238-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/libgraphviz238-shlibs.info
@@ -176,7 +176,7 @@ ConfigureParams: <<
 CompileScript: <<
 	#!/bin/sh -ev
 	ruby_h_path=`find /usr/lib/ruby -name ruby.h | head -n 1`
-	test -f "$ruby_h_path" || ruby_h_path=`find /System/Library/Frameworks -name ruby.h | head -n 1`
+	test -f "$ruby_h_path" || ruby_h_path=`find /System/Library/Frameworks -name ruby.h | grep -m1 /Headers/`
 	test -f "$ruby_h_path" && echo "Found ruby.h as $ruby_h_path, using its dirname..."
 	system_ruby_include=`dirname $ruby_h_path`
 	test -d "$system_ruby_include"
```
let it build on the 10.13 system too.

**Comment by nieder** [comment]
On my 10.13 system with xcode-cli-9.3.0, the ruby -I path in that compiler command is in `-I/System/Library/Frameworks/Ruby.framework/Versions/2.3/Headers/ruby`.

````
find /System/Library/Frameworks -name ruby.h`
Ruby.framework/Versions/2.3/Headers/ruby/ruby.h
Ruby.framework/Versions/2.3/Headers/ruby.h
Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby/ruby.h
Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby.h
````
And the graphviz build completes.

**Comment by dhomeier** [comment]
Here (10.13.5, Xcode cli 9.4.1.0.1.1528165917) it's
```
> find /System/Library/Frameworks -name ruby.h                                                                                                                                                                                                           
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.3/usr/include/ruby-2.3.0/ruby/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.3/Headers/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.3/Headers/ruby/ruby.h
```
I'd assume the order is simply random.

**Comment by nieder** [comment]
Is your system on a case sensitive HD?

**Comment by dhomeier** [comment]
Yes, but so is the 10.12 system, which has
```
> find /System/Library/Frameworks -name ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.0/Headers/ruby/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.0/Headers/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/ruby/ruby.h
/System/Library/Frameworks/Ruby.framework/Versions/2.0/usr/include/ruby-2.0.0/ruby.h
```

**Comment by dmacks** [comment]
Versions/2.0 vs Versions/2.3

Frameworks provide a versionless Current symlink that points to the latest Versions/ subdir.

**Comment by dmacks** [comment]
So yeah, 10.13 *only* has system ruby2.3 (not even retaining 2.0 for backward compatibility, which means we need separate packaging (different revisions) for 10.{9,10,11,12} vs 10.13 if we stay with apple ruby. I'm fine with the alternative of a unified package with fink's ruby as a dep (we're at 2.5 so we're better than apple anyway:).

**Comment by nieder** [comment]
Fine if it keeps us from splitting by dists.
```
	RUBY=%p/bin/ruby2.5 \
	RUBY_CFLAGS="-I%p/include/ruby-2.5 -I%p/include/ruby-2.5/%m-darwin" \
	RUBY_LIBS=%p/lib/libruby.2.5.dylib
```
and get rid of all the ruby.h tricks in CompileScript. Can even have `Type: ruby(2.5)`.

dhomeier:
So was there a problem with the Ruby version as such? I can only say the package built on 10.13 after fixing the order of include dirs, but haven't been able to test it beyond that, since the iMac is suffering from hardware (or heat...) troubles again.

PR #144, opened 2018-04-26
Updates to ECMWF API software

remkos:
This covers the upstream updates to two APIs from ECMWF:
GRIB API 1.26.1 (deprecated but still updated)
ECCODES 2.7.3 (replacement for GRIB API)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
@@ -64,7 +64,7 @@ InfoTest: <<
 	TestScript: <<
 		#!/bin/sh -ev
 		cd build
-		PATH=%b/build/bin:/%p/bin:/usr/bin:/bin DYLD_LIBRARY_PATH=%b/build/lib ctest || exit 2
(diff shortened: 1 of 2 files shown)


nieder:
For eccodes, testscript has a hardcoded /sw (not %p).

Also, might be good to add `-DCMAKE_C_FLAGS:STRING=-MD` to the cmake configure and then a `fink-package-precedence` check at the end of CompileScript to verify that the header dependencies that are being used.

Lastly, as before, unfortunately tests only pass if the current version is already installed. This is not really fixable (thanks SIP).

remkos:
I have replaced /sw by %p and added -DCMAKE_C_FLAGS:STRING=-MD
No *.Plo files are generated, so fink-package-precedence --headers does not work. Hence, I left that suggestion out.

dmacks:
Pass ```--depfile-ext='\.d'``` to fink-package-precedence to sync with passing ```-MD``` to the compiler via a *FLAGS variable.

remkos:
Thanks for the hint. fink-package-precedence is now called and passes.

PR #152, opened 2018-04-30
trackballs: Apply upstream patch to build on 10.13

nieder:
@microtherion 
On 10.13, trackballs fails to build with this error:
```
g++ -DLOCALEDIR=\"/sw/share/locale\" -DHAVE_CONFIG_H -I. -I.. -DSHARE_DIR=\"/sw/share/trackballs\" -DPACKAGE=\"trackballs\" -DVERSION=\"1.1.4\"  -I/sw/include -I/opt/X11/include  -g -g -O2 -I/sw/include -I/sw/include -D_THREAD_SAFE  -I./ -I../ -DDEFAULT_RESOLUTION=1 -MT trackballs-glHelp.o -MD -MP -MF .deps/trackballs-glHelp.Tpo -c -o trackballs-glHelp.o `test -f 'glHelp.cc' || echo './'`glHelp.cc
glHelp.cc:132:13: error: 'abs' is missing exception specification 'throw()'
inline Real abs(Real v) {return v>0.0?v:-v;}
```
The original SF trackballs has been discontinued, but an active fork at https://trackballs.github.io/ has the patch needed to fix the error.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/games/trackballs.info
+++ b/10.9-libcxx/stable/main/finkinfo/games/trackballs.info
@@ -32,7 +32,7 @@ Depends: <<
 Source: mirror:sourceforge:%n/%n-%v.tar.gz
 Source-MD5: 84e2e8bb68842a636da91673751279a0
 PatchFile: %n.patch
-PatchFile-MD5: f4f35c7b1b36a1872a242acbd5ecad08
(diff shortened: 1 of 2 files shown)


microtherion:
👍 haven't tested the fix, but it looks principled and correct to me.

PR #155, opened 2018-05-05
Gnome icon theme 2.31.0

nieder:
Update gnome-icon-theme to 2.31.0, the last of the 2.x series.
Only substantial change is to remove the chunk modifying the `allow_rendering` flag because that's no longer present in the source code and there's no indication of any git still being used in the code.


nieder:
Tested on 10.11/xcode7

PR #156, opened 2018-05-09
Viennarna cleanup

dmacks:
It didn't pass -m (self-test), and while fixing that I added f-p-p and looked at the ./configure output and found lots of other oddities. There are lots of upstream build-toolchain poor practices, and fixing them reduced the amount and fragility of the build overall and the other workarounds in the packaging.

I see there is a new upstream, which claims to be more portable in its library building, so maybe we would get .dylib and/or improved other aspects of the build process. I can look at it if anyone is interested, but for now, just need to get this thing buildable.


dmacks:
(sorry for the rebasing mess)

dmacks:
2 months no response. Pushing it (squashed to avoid the rebase mess)

PR #157, opened 2018-05-09
Alidot missing dependencies and new version

dmacks:


dmacks:
No response in >2 months...pushing (it's runtime breakage for 10.9 users who migrate to 10.10+)

PR #159, opened 2018-05-10
ucblogo: Fix FTBFS, add debian patches, update deps

dmacks:
As found by the bindist http://bindist.finkproject.org/10.13/logs/ucblogo_6.0-1_c644de3780c94d41addb76db2c7a8d50.log the autoconf2.13-era ./configure script broke again. While fixing it, I noticed missing ncurses deps, and added fink-package-precedence to help track it. Debian's ucblogo package had some useful patches (including some we already had), so I pulled them in also.


dmacks:
It's been a month for FTBFS with simple fixes. Easy enough for maintainer or someone else to undo or make further changes...

PR #160, opened 2018-05-10
sjeng: Fix FTBFS and import unreleased upstream revisions

dmacks:
Permissions problem under --build-as-nobody (fink's default mode) led me down a rabbit-hole of fixing things.


nieder:
With the new revision here:
````
test -z "/sw/build.build/root-sjeng-11.2-4/sw/bin" || ./install-sh -c -d "/sw/build.build/root-sjeng-11.2-4/sw/bin"
/bin/sh: ./install-sh: Permission denied
````
install-sh is 644.

microtherion:
Very interesting. Where did this rollup patch come from?

dmacks:
@microtherion https://github.com/gcp/sjeng/

dmacks:
@nieder Looks like that only happens if coreutils is not detected...yeah, chmod is better than a BDep there. Fixed.

nieder:
@microtherion : review ping

dmacks:
@microtherion, or @nieder in their absence?

Issue #139, opened 2018-04-19
ipython-genutils MIA


dmacks:
nbconvert-pyXX and nbformat-pyXX are unbuildable because they have a Depends:ipython-genutils-pyXX but no such package exists in our repo. While trying to see if some other package supplied a pymod with that name, I got the following in python3.6:
```
>>> import traitlets
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/sw/lib/python3.6/site-packages/traitlets/__init__.py", line 1, in <module>
    from .traitlets import *
  File "/sw/lib/python3.6/site-packages/traitlets/traitlets.py", line 56, in <module>
    from ipython_genutils import py3compat
ModuleNotFoundError: No module named 'ipython_genutils'
```
so our traitlets-pyXX package is missing a dependency on python-genutils-pyXX (and is also therefore broken because we have no such package). Could you please look into packaging that module? Or else, are there newer versions of the nb* and traitlets available that don't use what is seemingly a deprecated module?

schwehr:
ipython has turned into a zillion little python modules.  Such a pain to package the whole deal now.  sigh.

dhomeier:
@schwehr, I've been working on an ipython/jupyter update set for a while and finally got a working installation in my local tree. Includes a couple of circular dependencies, and putting the whole bunch on the Sourceforge tracker seemed almost as much of a pain...
I will try to put a PR together here in the not so distant future, just so you don't duplicate too much work.

dhomeier:
There also is the question of naming conventions - I think neither of the existing `jupyter-{client, core}-py` packages actually provide an executable to run `.ipynb` notebooks, right?
I have created a `notebook-py` packages that contains the `jupyter-notebook-pyNN` executable; one might argue if this would better be named `jupyter-py`...

dhomeier:
Congratulations from me as well!
I've started to add the various micropackages required by ipython in #164 now, got as far as to the new `nbconvert` version.
I wanted to make sure nothing breaks at an intermediate stage, but wasn't even able to get any of the existing `ipython` variants (3.2.1-2) to run at all:
```
> ipython-py35                                                                                                                                     
Traceback (most recent call last):
  File "/sw/bin/ipython-py35", line 4, in <module>
    from IPython import start_ipython
  File "/sw/lib/python3.5/site-packages/IPython/__init__.py", line 45, in <module>
    from .config.loader import Config
[...]

PR #164, opened 2018-05-14
IPython / Jupyter-notebook infrastructure

dhomeier:
This is an ongoing attempt to put together the umpteen packages required for the current upstream versions of IPython (5.4/6.2) and (Jupyter-)Notebook (5.4), providing some core dependencies like `ipython-genutils` and `ipykernel` and numerous smaller new packages or updates of same.
Building and especially testing the whole ecosystem suffers from a number of circular dependencies; at this point I've updated everything that should be required to upgrade to the new `jupyter-core` and `jupyter-client` and `nbconvert`.
The main challenge is that `ipykernel` only builds with a working `ipython` installed, while the other packages can be (partially) built without `ipykernel`, but some or all tests will fail. As upgrade path I therefore suggest
```
fink -m update dateutil-pyXX traitlets-pyXX zmq-pyXX
fink --validate update jupyter-core-pyXX jupyter-client-pyXX
fink -m update entrypoints-pyXX nbformat-pyXX pygments-pyXX testpath-pyXX
fink --validate update nbconvert-pyXX
```
Feedback desired, cherrypicking missing packages e.g. for #139 welcome.


dhomeier:
Should be everything required for the `ipython` update at this point; afterwards `ipykernel` can be built and then the packages first built without tests verified:
```
fink -m update pickleshare-pyXX prompt-toolkit-pyXX ipywidgets-pyXX
fink --validate validate ipython-pyXX
fink -m update ipykernel-pyXX
fink -m rebuild jupyter-core-pyXX jupyter-client-pyXX nbconvert-pyXX ipython-pyXX
```

dhomeier:
The versioned 5.7.0 was intended for the long-term support for Python 2.7 (although the minor release numbers still seem to be incremented fairly regularly), `ipython-py` would be for the latest upstream (currently 6.4.0), see next commit coming up. Of course we can also keep the 3.x branch around in a versioned file (there is a `3.2.3` release though, and a whole series `4.0.0-4.2.1`), with a `Conflicts: path-py (>= 10.0-1)`.

dhomeier:
Though I wouldn't know how to build the older version with fink once the newer `.info` is in place - only `apt-get install`ed legacy stuff so far...

nieder:
How does one start committing (cherry picking) individual commits from here to simplify the master diff? For example, baae7a7 doesn't have any other new deps.

dhomeier:
I don't have any experience with cherrypicking as such, but clearly baae7a7 would be a possible start (which set it all off in #139). Taking the entire commit history up to `ipykernel`, `jupyter-core` or jupyter-client` (752357f) might also work.

nieder:
The good news is that I can cherry pick directly into master using SourceTree and push individually to origin from there (baae7a7 and 5d9e6a7) The bad news is that it's not being reflected in the diff when viewed here in the PR. The commits are showing up with different hashes in master, so maybe that's why it can't track them?
@dmacks I've seen you do individual cherry picks. What was your procedure?

nieder:
Testing 8a2b33a (ipykernel-pyXX), I get this error:

````
ipykernel-4.8.2/setup.cfg
ipykernel-4.8.2/setup.py
sudo -u fink-bld [ENV] sh -c /tmp/fink.V0j_Q
python2.7 setup.py build
  File "setup.py", line 91, in <module>
    from ipykernel.kernelspec import write_kernel_spec, make_ipkernel_cmd, KERNEL_NAME
  File "/sw/build.build/ipykernel-py27-4.8.2-1/ipykernel-4.8.2/ipykernel/__init__.py", line 2, in <module>
    from .connect import *
  File "/sw/build.build/ipykernel-py27-4.8.2-1/ipykernel-4.8.2/ipykernel/connect.py", line 13, in <module>
    from IPython.core.profiledir import ProfileDir
ImportError: No module named IPython.core.profiledir
### execution of python2.7 failed, exit code 1
````
I do not have ipython-py (3.2.1-2 from stable) installed.

nieder:
I understand. I'll go forward with testing the [above build path](https://github.com/fink/fink-distributions/pull/164#issuecomment-389359893). !lart things that need their own dependees.

For the ipython 5.7.0 version: is there a reason to keep py35 and py36 variants of it since those will be available in the 6.x series? If there isn't a need to keep the 5.7-py3X, then ipython-5.7.0.info can instead be called ipython-py27.info and there won't be .info naming issues if a new release is made in the 5.x series. Thoughts?

dhomeier:
Yes, that would be a good option, as development on 5.x is obviously continuing. I had only kept the Python 3 variants as a backup in case something turned out to be broken in ipython 6.x, but since this is well past 6.0 now, that does not appear to be the case.

nieder:
Building jupyter-core-py27 (`fink --validate update jupyter-core-py27`):
````
#!/bin/bash -ev
 find . ../jupyter_core  -name __pycache__ -or -name '*.py[oc]' -exec rm -r {} \;
find: ../jupyter_core: No such file or directory
````

dhomeier:
Sorry, missed updating the change in `cwd` when I moved the cleanup from `TestScript` to `InstallScript`.

dmacks:
I notice some TestScript have ```|| exit 1```, which makes fink treat test-suite failures as warnings, some have ```|| exit 2``` (test-suite failures treated as failures), and some have no shell exit-code control. That last category is not clear as to your intended effect (I assume it's equivalent to exit=1?). The exit=1 case is reasonable for where you have expected failures, although it would be even better if those specific tests could be disabled. Or else could you have a user-visible ```echo "some test failures expected: [list them]``` so when users see a failure that is then ignored, they don't get confused or whiny:)

dhomeier:
Updated `TestDepends`, seems I haven't tested install independently of `nbconvert`.

dhomeier:
@dmacks, sometimes I've taken over the test return codes from previous package versions, otherwise my priority was to at least have some test coverage, as there were/are many packages without any `TestScript` at all. Yes, the standard python return  from failed tests is 1 (not 100% sure about direct calls of `pytest` or `nosetests`); might be 2 above a certain number of `Fail`/`Error`s - at least one can tweak the python call to return this.
I am less concerned about scaring users who actually build with `-m`, in most cases the error messages are already out of sight once the install phase is completed, but as @nieder has already commented, accepting failed tests might mask new errors that could pop up at a later time. I will try to rebuild everything and change the test returns to `exit 2` where possible, when I have some time.

dhomeier:
Superseded by #391 and #413.

PR #130, opened 2018-04-12
Pylint update (including dependencies for py36)

dhomeier:
This is bringing `pylint-py` up to date to the latest upstream `1.8.4` and adds/updates missing dependencies for `type_pkg[python] = 36`.
Built and tested for python 2.7-3.6 on 10.12 and 10.13 - 3 test failures on Python 3.x, > 222 (basically the entire test_stdlib) mostly unicode-related on 2.7.14.
Changed names of versioned executables to standard -py%type_pkg[python] scheme (pylint-gui no longer included). 2to3 treatment of sources no longer required. Added cleanup of byte-compiled files to InstallScript.

PR #166, opened 2018-05-16
Added missing futures-py dependency to Python2.7 variant of isort-py

dhomeier:
In `4.3.4` `isort-py` (updated in #130) added an import from `concurrent.futures`; the `py27` variant needs the backport in `futures-py` for this. 3 test failures remaining.


bcully:
LGTM

PR #170, opened 2018-05-20
libspatialindex cleanup

dmacks:
Scrap the BDO packages of unused old libversions (retain -shlibs per policy), turn off static libs, and misc packaging cleanups.

@babayoshihiko, these are yours.


dmacks:
It's been almost 4 months and it's a fairly walled garden...

PR #171, opened 2018-05-21
Geos upgrades

dmacks:
New versions of django-pyXX and shapely-pyXX and switch their dependencies from geos3.4 to 3.5 or 3.6.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/shapely-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/shapely-py.info
@@ -1,35 +1,40 @@
 Info2: <<
 Package: shapely-py%type_pkg[python]
-Version: 1.5.8
+Version: 1.6.4.post1
 Revision: 1
 Homepage: https://pypi.python.org/pypi/Shapely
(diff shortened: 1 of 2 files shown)


nieder:
Use a source URL of the form `https://pypi.io/packages/source/S/PKG/...` to avoid needing the hashed URL. `pypi.io` is the new shorter URL for PyPi packages.

Other suggestions:
1) Add a note in django why it's at geos-3.5.0 vs 3.6.1 (assuming it can't build w/ the 3.6 series).
2) Add a new geos entry to the `Type:` field and use that to keep track of the linked version of geos and replace hardcoded versioning with `%type_raw[geos]` throughout the info file to simplify tracking necessary changes across upgrades.

None are stoppers.

dmacks:
django has a note:
```
  # See docs/ref/contrib/gis/install/geolibs.txt for
  # information about supported versions
```
It does runtime loading rather than linking so it's fragile towards ABI changes, and I don't actually use this module so I don't know which ones I would even need to check.

dmacks:
All concerns seem addressed...

PR #176, opened 2018-05-26
Postgresql9x: : update to latest upstream + backport info file changes from pg10

wodev:
the PR updates the pg 9.x packages to the latest upstream version
postgresql94: Update to 9.4.18 
postgresql96: Update to 9.5.13 
postgresql96: Update to 9.6.9 

and the latest info file changes from pg10 (using a variant to minimize diffs among the version).

All versions testet with fink -m build.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgresql94.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgresql94.info
@@ -1,8 +1,9 @@
 Info4: <<
-Package: postgresql94
-Version: 9.4.17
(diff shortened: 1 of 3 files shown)


nieder:
Looked fine to me with the exception that I think you need to keep the epoch versioning of the `Depends:` in the -dev SplitOffs against parent package (around line 203). The postgresql10 package is not Epoched, so it's fine over there, but the 9.x packages are all epoched and the leading `1:` is needed to make sure the postgresql9x-dev packages match the corresponding version of the sister -shlibs package.

PR #180, opened 2018-06-12
checkmol: Update to new update version (0.5b).

kamischi:


kamischi:
Patch applied as well as to the package mol2ps.

PR #182, opened 2018-06-13
readline7: scrap static lib; switch to '.plX' versioning (PR#178)

dmacks:
I rebuilt all rdeps on readline7 (except intermediate libversions where I only built the earliest and latest) and there were no differences in detection. Saved a minute of build time, and hey, what good are static libs anyway (esp. noting that this is the active release series, so prone to changes). I used this rebuild as an opportunity to switch to ```Version: X.Y.plZ``` for upstream X.Y upstream patchlevel Z, as discussed in PR #178

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/readline7.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/readline7.info
@@ -1,6 +1,7 @@
 Info4: <<
 Package: readline7
-# teeny-version is the upstream patchlevel ("x.y.z" is "x.y" pl "z")
+# Upstream releases official patches to their major.minor version
+# We use "X.Y.plZ" for upstream "X.Y patchlevel Z"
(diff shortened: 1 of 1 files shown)


dmacks:
hold a moment...saw something weird...

nieder:
Nit:
````
"x.y.z" is "x.y" pl "z"
````
This comment as written is no longer 100% true to the new versioning format.
````
"X.Y.plZ" is "X.Y" patch level "Z"
````
Are we screwed if they ever have a release with a true teeny version?

dmacks:
No more screwed than before unless they decide to use their actual patchlevel as the teeny-version value. They seem pretty set in their ways using official patches to major.minor going back at least as far as the 4.y series.

dmacks:
okay, weirdnesses I was seeing were not related to readline7 being +/- static.

PR #183, opened 2018-06-15
Gcrypt updates

nieder:
Updates to gcrypt libraries and programs. Ironically, this first started with an update to libgcrypt20, but version 1.8.3 hangs during the tests, so I didn't update it. All others tested on 10.11.


dmacks:
gnupg2 looks fine to me (I had been looking at it as part of an openssl100 sweep).

PR #186, opened 2018-06-17
nano: update to new release version of git (2.9.8)

kamischi:
Due to the merge of the current git version into the stable version, nano-cur can be removed.
Also taking over maintainer-ship.

Sorry. I forgot to upload the new tar ball and adjust the source in the .info file. It needs a new file.

Upload with fixed source URL finished. Ready to be tested for a merge.


kamischi:
Close this pull request and open a new after uploading an updated .info file.

kamischi:
I do not think that the package nano-cur will be needed in the future, at least in short time, since the current setup in nano is good enough for git releases and i am not going to follow the development any closer, such as on every commit. Is your proposal really worth the extra clutter and possible trouble from two differently named packages, which are actually the same? I would favor to end the mess with some trouble over no trouble, but never ending mess, in particular since no further package depends on nano-cur.

dmacks:
How about a Type:bundle package for nano-cur, with a PreInstScript the echos an instruction to install one of the other variants and then exits with a failure code? That way users with nano-cur will see that they need to update, and when they try they will learn how to get onto the new package.

kamischi:
I have committed updated files for nano.info and nano-cur.info. nano-cur is my first version of an "obsolete" package. So, check it carefully and let me know.

kamischi:
dmacks: ping

kamischi:
superseded by nieder's update to version 4.1.

PR #187, opened 2018-06-17
lz4: update to new upstream version (1.8.2)

kamischi:
Also taking maintainer ship.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/lz4.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/lz4.info
@@ -11,7 +11,13 @@ Source-Checksum: SHA1(dc24ee207db0b4481ff5e74b608c1af1c18c5a7c)
 
 BuildDependsOnly: true
 Depends: %N-shlibs (= %v-%r)
+BuildDepends: fink-package-precedence
 
+# prevents dragging in old libs from /sw/include
+NoSetCPPFLAGS: true
+NoSetCXXFLAGS: true
+
+# flags from gecko
 SetCFLAGS: -g -O2 -fstack-protector -Wformat -Werror=format-security -fPIE
 SetCPPFLAGS: -D_FORTIFY_SOURCE=2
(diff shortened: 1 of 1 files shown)


dmacks:
The INSTALL file is just instruciotns about how to compile and install the package from the source tarball. Not really useful to install it for users (this is generally true for "INSTALL" files).

dmacks:
One of the goals of fink-package-precedence is to detect use of headers from previously installed versions of this package. Using ```--no-headers``` defeats that feature. If you pass ```-MD``` (probably via ```SetCPPFLAGS```) you can use ```--depfile-ext='\.d'``` to get the headers tracking. The makefiles look hackable to get flags in the correct order (i.e., not needing to NoSet* them) but since there are no external libs being used anyway, your approach is fine.

gecko had some standard Set* entries he put in all his packages, but some of them are not needed in some cases and/or obsolete overall. Here, I don't see any c++ in this package, so you can scrap ```SetCXXFLAGS```.  In ```SetCFLAGS```, the ```-W*``` flags should probably go to ```SetCPPFLAGS``` (since they are only for the source compiling, not the later linking stage). And ```-fPIE``` should either only be in ```SetCFLAGS``` or in both ```SetCPPFLAGS``` and ```SetLDFLAGS``` instead. For C compiling, CFLAGS is used in both stages whereas CPPFLAGS is only used during compiling and LDFLAGS only during linking (I have no idea if -fPIE is relevant any more).

dmacks:
The 'lz4' package, which is ```BuildDependsOnly:true```, contains both userland programs and the compiler files for using liblz4. That is a policy and usability problem because it means no other package that needs (for example) /sw/bin/lz4 could Depends on it. The runtime programs (and their manpages) should be in a separate package (maybe 'lz4-bin'?).

kamischi:
Sounds like valid points to me. It will take some time to rework things. I will have to read about some stuff, such as the flags, but I am ready to go for them.

nieder:
The new package lz4-dev needs to `Replaces: lz4 (<< 1.8.2-1)`. Otherwise you'll get a file collision because something like %p/include/lz4.h from the old lz4 (bin and headers) could be installed and then the new lz4-dev will also try to install %p/include/lz4.h.

PR #188, opened 2018-06-20
ghostscript-9.23-1

remkos:
This is an update for the ghostscript, from upstream version 9.21 to 9.23.
This removes the dependency on libcms2.
Maintainer switched from "none" to myself.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/text/ghostscript.info
+++ b/10.9-libcxx/stable/main/finkinfo/text/ghostscript.info
@@ -8,7 +8,7 @@ Description: Interpreter for PostScript and PDF
 Source: https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/download/gs923/ghostscript-%v.tar.xz
 Source-MD5: c5cd025f17e0d87f812bcdb76a61ba3a
 PatchFile: ghostscript.patch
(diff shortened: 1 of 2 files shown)


dmacks:
Just a note (since I was one who worked on switching to using internal libs), that switching away from external lcms2 is required by the package itself now. They need a patched version that is not available on its own.

remkos:
One further note: I actually installed lcms, lcms-shlibs, lcms2, and lcms2-shlibs to check that none of them were accidentally picked up during configuration or compilation. And indeed, they were not.

dmacks:
Thanks for the brute-force testing of lcms (and avoiding BuildConflicts that aren't needed)! Here is [upstream fix](http://git.ghostscript.com/?p=ghostpdl.git;a=commitdiff;h=0a4ef8b01c2ea146b9dce7b80f21f6cd65420099) for the doc symlink, though I don't object to keeping it around.

remkos:
Thanks for the link to upstream fix. I incorporated that in the latest commit (6a8ddec) and also updated the DescPackaging information to spell out the patches.
I hope this is now satisfactory. Ghostscript 9.23 comes with important rendering improvements (that were faulty in 9.21), so this is a valuable update.

PR #192, opened 2018-07-01
New upstream setuptools 39.2.0 and add py37 variant

danielj7:
I just added python 3.7 and virtually every pymod will need this before they can be updated.


dhomeier:
Nice. I've created `py37` versions of `nosetests` and `numpy` and successfully built them against this (although it appears they would also build with `32.3.0`).
As a general question (pinging @nieder), is it preferred to submit all dependencies of an updated package into a single PR (would be just `nosetests` and `numpy` in this case, since `cython` is already in master), or create a separate PR for each package?

nieder:
@dhomeier: it's fine to submit multiple packages in one PR if their purpose is related (eg, a bunch of py37 additions)

PR #197, opened 2018-07-08
update converatll to latest upstream version

khorton:
built with fink -m rebuild.  tested on macOS 10.13

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/convertall.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/convertall.info
@@ -1,6 +1,6 @@
 Info2: <<
 Package: convertall
-Version: 0.7.3
+Version: 0.7.5
 Revision: 1
 Type: python (3.4)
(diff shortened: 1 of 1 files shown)


khorton:
Yes, I did quickly wonder about that for this update, but didn’t have time to research which version might be preferred, and do the required testing.  

What is the preferred python version now?

Thanks,

Kevin

nieder:
Not sure if there's a preferred 3.x version.  python27 is supported upstream until 2020, but I couldn't find any thing online listing when support for the various python3 versions would end. 3.4 is still supported, however. I also don't know if there's any functional benefit in bumping py3x version in this case.

dhomeier:
I think we had like half-agreed to retire 3.4, and even 3.5 in #140, but for the time being most of the `py34` variants have stayed in, so there's no particular urgency.
Newer Python might bring some performance improvements, but I doubt it will be noticeable in a GUI app. But that aside, 3.6 should be the best and most future-proof choice right now (will probably take some time for the `py37` variants of `pyqt5-mac` to appear).

PR #119, opened 2018-03-29
Sphinx depencies for 1.7.2

dhomeier:
Additional packages and updates required for a `Sphinx`-update (see also SF package submission [#4723](https://sourceforge.net/p/fink/package-submissions/4723/)), now to upstream `1.7.2`.
Provides dependencies not included in #82 - the Python3.6 variant of `html5lib-py` and update to `1.0.1` + a new dependency, and the `packaging` module required for all `Sphinx` tests to run successfully.

PR #199, opened 2018-07-10
Pytest/tox and dependencies updates and Python 3.7 variants

dhomeier:
Updates to `pytest-3.6.2` and `tox-3.0.0` with added `py37` variants, plus their `Depends`, `TestDepends` and their respective `TestDepends`.

Unfortunately `3.6.2` brings circular dependency madness to a new level, requiring a new module `invoke` which depends on a `pytest-relaxed` extension, which in turn explicitly requires `pytest<3.3`! I have taken the hard way out of this by hacking `pytest-relaxed-py` to accept `pytest-py-3.6`; this introduces some failures in its own tests, but allows the `invoke-py` test suite to complete successfully.

Builds with tests enabled without prior installation naturally are largely impossible; a working update path with subsequent testing should be
```
fink --validate update tox-py37 pytest-mock-py37
fink -m rebuild atomicwrites-py37 attrs-py37 invoke-py37 more-itertools-py37 pip-py37 pluggy-py37 py-py37 setuptools-scm-py37 six-py37 virtualenv-py37
fink -m rebuild pytest-py37 tox-py37 pytest-mock-py37
```
(tested on 10.12 and 10.13).
The `tox-py` update is largely taken from the updated package description in [#119](https://github.com/fink/fink-distributions/pull/119#issuecomment-393193908).

The whole package also requires the `setuptools-tng-py` update from #192.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4 -*-
 Info4: << 
 Package: pip-py%type_pkg[python]
-Type: python (2.7 3.5 3.6 3.7)
+Type: python (2.7 3.4 3.5 3.6 3.7)
 Version: 10.0.1
 Revision: 1
 
--- a/10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
@@ -1,17 +1,19 @@
-Info2: << 
+# -*- coding: ascii; tab-width: 4 -*-
+Info4: << 
 Package: pip-py%type_pkg[python]
-Type: python (2.7 3.4 3.5 3.6)
-Version: 9.0.1
+Type: python (2.7 3.5 3.6 3.7)
+Version: 10.0.1
 Revision: 1
+
 Source: https://pypi.io/packages/source/p/pip/pip-%v.tar.gz
-Source-MD5: 35f01da33009719497f01a4ba69d63c9
+Source-Checksum: SHA256(f2bd08e0cd1b06e10218feaf6fef299f473ba706582eb3bd9d52203fdbd7ee68)
 
 Depends: python%type_pkg[python], setuptools-tng-py%type_pkg[python]
 BuildDepends: setuptools-tng-py%type_pkg[python]
 
 Description: Python package management
 Maintainer: Kurt Schwehr <goatbar@users.sourceforge.net>
-DocFiles: AUTHORS.txt LICENSE.txt docs
+DocFiles: AUTHORS.txt LICENSE.txt NEWS.rst README.rst docs
 CompileScript: echo Skipping compile stage
 InstallScript: <<
   #!/bin/bash -ev
@@ -29,12 +31,12 @@ PreRmScript: <<
  if [ $1 != "upgrade" ]; then
    update-alternatives --verbose --remove pip %p/bin/pip-py%type_raw[python]
  else
-   echo "Skipping update-alternatives during an upgrage."
+   echo "Skipping update-alternatives during an upgrade."
  fi
 <<
 
 LICENSE: OSI-Approved
-HomePage: http://pypi.python.org/pypi/pip
+HomePage: https://pip.pypa.io/
 
 DescDetail: <<
 pip is a tool for installing and managing Python packages, such as
@@ -46,5 +48,5 @@ bad bad bad idea.  Use --target with install to install packages
 to the side someplace.  Or better yet, use virtualenv.
 << 
 
-# Info2
+# Info4
 <<
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
@@ -31,9 +31,9 @@ Source: https://pypi.io/packages/source/p/py/py-%v.tar.gz
 Source-Checksum: SHA256(3fd59af7435864e1a243790d322d763925431213b6b8529c6ca71081ace3bbf7)
 
 Depends: python%type_pkg[python]
-BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python]
+BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python], setuptools-scm-py%type_pkg[python]
 
-CompileScript: %p/bin/python%type_raw[python] setup.py build
+CompileScript: %p/bin/python%type_raw[python] setup.py build --offline
 
 InstallScript: <<
 	%p/bin/python%type_raw[python] setup.py install --root=%d
@@ -47,14 +47,14 @@ InfoTest: <<
 	# does not help).
 	TestScript: <<
 		#!/bin/bash -ev
-		if [ $(fink list -i pytest-py%type_pkg[python] | wc -l) -lt 1 ]; then
-			echo "Warning: unable to run tests without package pytest-py%type_pkg[python] installed!"
-			exit 1
+		if [ -x %p/bin/pytest-%type_raw[python] ]; then
+			perl -pi.bak -e 's|(py.test)( )|$1-%type_raw[python]$2|;' tox.ini
+			#TOX_TESTENV_PASSENV="PYTHONPATH=%b/build/lib" %p/bin/tox-py%type_pkg[python] -e py%type_pkg[python]
+			cd testing
+			%p/bin/pytest-%type_raw[python] code io_ log process root path/test_*.py
+		else
+			echo "pytest-py%type_pkg[python] is not installed. Skipping tests."
 		fi
-		perl -pi.bak -e 's|(py.test)( )|$1-%type_raw[python]$2|;' tox.ini
-		#TOX_TESTENV_PASSENV="PYTHONPATH=%b/build/lib" %p/bin/tox-py%type_pkg[python] -e py%type_pkg[python]
-		cd testing
-		%p/bin/pytest-%type_raw[python] code io_ log process root path/test_*.py
 	<<
 <<
 
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
@@ -1,9 +1,9 @@
 # -*- coding: ascii; tab-width: 4 -*-
-Info2: <<
+Info4: <<
 Package: py-py%type_pkg[python]
-Version: 1.5.3
+Version: 1.5.4
 Revision: 1
-Type: python (2.7 3.4 3.5 3.6)
+Type: python (2.7 3.5 3.6 3.7)
 
 Description: Python development support library
 DescDetail: <<
@@ -19,14 +19,13 @@ DescDetail: <<
 	NOTE: prior to the 1.4 release this distribution used to contain
 	py.test which is now its own package, pytest-py%type_pkg[python].
 <<
-DescPackaging: Can't run tests since it needs pytest and pytest needs py.
+DescPackaging: Needs pytest to run tests and pytest needs py; build just with --validate first.
 Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
 License: BSD
-Homepage: https://pypi.python.org/pypi/py
+Homepage: https://pypi.org/project/py
 
 Source: https://pypi.io/packages/source/p/py/py-%v.tar.gz
-Source-MD5: 667d37a148ad9fb81266492903f2d880
-Source-Checksum: SHA256(29c9fab495d7528e80ba1e343b958684f4ace687327e6f789a94bf3d1915f881)
+Source-Checksum: SHA256(3fd59af7435864e1a243790d322d763925431213b6b8529c6ca71081ace3bbf7)
 
 Depends: python%type_pkg[python]
 BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python]
@@ -37,5 +36,22 @@ InstallScript: <<
 	%p/bin/python%type_raw[python] setup.py install --root=%d
 <<
 
+InfoTest: <<
+	# Needs to be run inside 'testing' subdir when package is already installed
+	# to avoid conflicts with files in install location.
+	# Tries to run path/{common,svntestbase}.py with tons of errors about missing
+	# fixture 'path1', therefore running test_* files explicitly ("-k 'test_'"
+	# does not help).
+	TestDepends: pytest-py%type_pkg[python]
+	TestScript: <<
+		#!/bin/bash -ev
+		perl -pi.bak -e 's|(py.test)( )|$1-%type_raw[python]$2|;' tox.ini
+		#TOX_TESTENV_PASSENV="PYTHONPATH=%b/build/lib" %p/bin/tox-py%type_pkg[python] -e py%type_pkg[python]
+		cd testing
+		%p/bin/pytest-%type_raw[python] code io_ log process root path/test_*.py
+	<<
+<<
+
 DocFiles: CHANGELOG LICENSE PKG-INFO README* doc
+# Info4
 <<
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info
@@ -3,7 +3,7 @@ Info4: <<
 Package: py-py%type_pkg[python]
 Version: 1.5.4
 Revision: 1
-Type: python (2.7 3.5 3.6 3.7)
+Type: python (2.7 3.4 3.5 3.6 3.7)
 
 Description: Python development support library
 DescDetail: <<
@@ -19,7 +19,10 @@ DescDetail: <<
 	NOTE: prior to the 1.4 release this distribution used to contain
 	py.test which is now its own package, pytest-py%type_pkg[python].
 <<
-DescPackaging: Needs pytest to run tests and pytest needs py; build just with --validate first.
+DescPackaging: <<
+	Needs pytest to run tests and pytest needs py; test script running
+	depending on installation status of pytest-py%type_pkg[python].
+<<
 Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
 License: BSD
 Homepage: https://pypi.org/project/py
@@ -42,9 +45,12 @@ InfoTest: <<
 	# Tries to run path/{common,svntestbase}.py with tons of errors about missing
 	# fixture 'path1', therefore running test_* files explicitly ("-k 'test_'"
 	# does not help).
-	TestDepends: pytest-py%type_pkg[python]
 	TestScript: <<
 		#!/bin/bash -ev
+		if [ $(fink list -i pytest-py%type_pkg[python] | wc -l) -lt 1 ]; then
+			echo "Warning: unable to run tests without package pytest-py%type_pkg[python] installed!"
+			exit 1
+		fi
 		perl -pi.bak -e 's|(py.test)( )|$1-%type_raw[python]$2|;' tox.ini
 		#TOX_TESTENV_PASSENV="PYTHONPATH=%b/build/lib" %p/bin/tox-py%type_pkg[python] -e py%type_pkg[python]
 		cd testing
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytest-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytest-py.info
@@ -3,7 +3,7 @@ Info4: <<
 Package: pytest-py%type_pkg[python]
 Version: 3.6.2
 Revision: 1
-Type: python (2.7 3.5 3.6 3.7)
+Type: python (2.7 3.4 3.5 3.6 3.7)
 
 Description: Cross-project testing tool for Python
 DescDetail: <<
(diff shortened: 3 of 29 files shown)


danielj7:
Thanks for working on this. I do have some concerns.

1. You dropped all py34 variants. While I don't have an issue with getting rid of 34, by only doing it with some packages lots of others will break and that's not the Fink Way. For instance, lots of things depend on six and py. Now all those packages will just stop working for py34.

2. Checksums. I've been including both md5 and sha256 since sha256 requires an extra package to be installed, otherwise those packages will not build. I'm not sure what the official policy is, but that seems bad to me.

3. py-py tests. I appreciate the work getting tests to work in py-py, but just putting a note in the package saying to build with –validate first isn't a good choice. For example, that will break auto builds since they're just going to run in –maintainer mode. Unfortunately, there's just no good way around the circular dependency; fink isn't smart enough. I suppose you could put logic in TestScript that checks for the presence of pytest and only running the tests in that case, but it seems like a lot of work. It looks like pytest-mock also has this issue.

4. Is there any particular reason for changing all the Info2's to Info4's? You're not actually using any Info4 features that I can see.

5. Some of the packages belong to Kurt Schwehr. Have you asked him if it's ok? I'm adding him to the PR.

10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
```diff
@@ -1,17 +1,19 @@
-Info2: << 
+# -*- coding: ascii; tab-width: 4 -*-
```
schwehr: Should these emacs settings really be in the files?

dhomeier: I've found them in most of the newer `.info` files, so assumed it was policy now. Don't really care, since I am using vi for editing them most of the time (and personally prefer 2-4 spaces indents...), so I am flexible in that respect.

schwehr:
I'm fine with all of the packaging that has me tagged as the owner.  @danielj7 , Thanks for looping me in.

10.9-libcxx/stable/main/finkinfo/languages/pip-py.info
danielj7: I add them to my packages. BBEdit uses them. I like forcing ascii since info files should always be ascii and I like 4 space indents. :) Feel free to use it or not in your own packages. It's not policy.

danielj7:
1. Yeah that's going to have to be a core decision.

2. After looking at fink's source, it looks like sha256 is fine to use alone as long as either /usr/bin/openssl or /usr/bin/shasum are present and current OS versions have both.

3. It is unsatisfying to not run tests but an increasing number of pymods are clearly designed to be built and tested as a unit which is kinda hard with a package manager. The upstream developers clearly only care about testing themselves. I try to make tests work as much as possible but if upstream makes things too difficult I can only do so much.

4. That's fine and I don't really care, I just usually use the lowest version required. It's a matter of taste. :)

dhomeier:
2. OK, I leave it at that then (thought I had seen some packages with `Checksum SHA256` only before).

3. And they seem to increasingly rely on pip simply downloading and installing anything required by their package as needed (that's what e.g. the `pytest-relaxed` tests would have done before the patch, installing `pytest-3.2` locally inside their test environment. Which let all tests pass, but was somewhat meaningless, because that was not the environment the installed package would run in).

Which also is a mess for package maintainers and has even been identified as a potential security leak.

I made the affected `TestScripts` now checking for the presence of `pytest-py` or `tox-py` before running the tests, checked this in the `python3.4` environment I had previously wiped clean. Seems to work, but I may have overlooked something, because probably not all the .debs had been newly built...

This might still be contentious as the testing process now depends on installed packages (but the package created should not).

dhomeier:
I have added updates for `pygments`, `jinja2` and dependencies here - they are required by 2 or 3 tests of the `numba-py` package in #207, but depend heavily on the `pytest` infrastructure in this PR.

nieder:
1) Thanks for restoring py34 packages. Until python34 is retired (either as we move to a new dist and decide to trim variants), upstream stops supporting it, or maintainer axes it (prob for one of the 2 previous reasons), there's no reason to remove existing packages. **New** packages can definitely be created w/out a py34 variant but there's no reason to leave existing users behind (unless upstream stops supporting py34 and then we can make a py34-only variant).
3) For tests that cause circularity, another option is to check for the presence of a file from the required installed package instead of calling `fink list`. Either one is fine. File check might be slightly less resource intensive. If testing is turned on (not commented out) but uses an IFTHEN check to get around circular TestDepends, please make sure to list the package in question in a comment near the InfoTest block (like you've been doing).

nieder:
````
+	TestScript: <<
+		#!/bin/bash -ev
+		if [ $(fink list -i tox-py%type_pkg[python] | wc -l) -lt 1 ]; then
+			echo "Warning: unable to run tests without package tox-py%type_pkg[python] installed!"
+			exit 1
+		fi
+		%p/bin/tox-py%type_pkg[python] -e py%type_pkg[python] || exit 2
+	<<
````
Might it be better to flip the logic here and only run the tests if the needed package is installed, otherwise exit 0 cleanly (with perhaps a comment about skipping tests because tox-py is not installed?

```bash
if -f %p/lib/python%type_raw[python]/site-packages/tox/__init__.py; then
    %p/bin/tox-py%type_pkg[python] -e py%type_pkg[python] || exit 2
else
    echo "tox-py%type_pkg[python] is not installed. Skipping tests."
    echo 0
fi

**Comment by dhomeier** [comment]
The tests are not run either way; I just thought a warning about the incomplete test environment might be in place. But a clean exit is fine with me, too - it is not an issue with the package itself after all.

And checking for the installed file is indeed much simpler and cleaner.

**Comment by nieder** [comment]
I don't follow. The code chunk I pasted from your patch will have the tests run if tox-py is present (but not as an enforced TestDepends.
The way I understood your process was:
1) test for presence of tox-py through fink. If absent, exit TestScript.
2) Continue with TestScript and run tox-py.

My proposal was:
1) test for presence of tox-py through file check. If present, run tox-py
2) Else, skip tests and exit test.

Either way is fine. The 2nd method makes more sense to me, but it doesn't matter. It really does stink that so many python mods have suddenly decided they all need each other for self-tests.

File: 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/tox-py.info [hunk 2]
```diff
@@ -1,50 +1,95 @@
...
 
 PostInstScript: <<
-  update-alternatives --verbose --install %p/bin/tox tox-py %p/bin/tox-py%type_pkg[python] %type_pkg[python] --slave %p/bin/tox-quickstart tox-quickstart %p/bin/tox-quickstart%type_raw[python]
+    update-alternatives --verbose --install %p/bin/tox tox-py %p/bin/tox-py%type_pkg[python] %type_pkg[python] --slave %p/bin/tox-quickstart tox-quickstart %p/bin/tox-quickstart-py%type_pkg[python]
+    echo ""
+    echo "If you have just built and installed the tox dependencies"
+    echo "coverage-py%type_pkg[python], hypothesis-py%type_pkg[python], pytest-py%type_pkg[python], pytest-mock-py%type_pkg[python]"
+    echo "for the first time, they will have been build with their tests disabled."
+    echo "You may now test their builds using the `fink -m rebuild` command."
+<<
```
nieder:
````
$ fink validate tox-py.info 
Validating package file tox-py.info...
WARNING: Unable to parse the line "<<" at line 95 of "tox-py.info".
Field(s) outside info4 block! Skipping tox-py.info
````
One too many `<<` after `PostInstScript.

File: 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/py-py.info [hunk 3]
```diff
@@ -19,14 +19,16 @@ DescDetail: <<
...
-Homepage: https://pypi.python.org/pypi/py
+Homepage: https://pypi.org/project/py
 
 Source: https://pypi.io/packages/source/p/py/py-%v.tar.gz
-Source-MD5: 667d37a148ad9fb81266492903f2d880
-Source-Checksum: SHA256(29c9fab495d7528e80ba1e343b958684f4ace687327e6f789a94bf3d1915f881)
+Source-Checksum: SHA256(3fd59af7435864e1a243790d322d763925431213b6b8529c6ca71081ace3bbf7)
 
 Depends: python%type_pkg[python]
 BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python]
```
nieder: Needs setuptools-scm-pyXX:
````python
/sw/bin/python2.7 setup.py build
Download error on https://pypi.org/simple/setuptools-scm/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726) -- Some packages may not be found!
Couldn't find index page for 'setuptools-scm' (maybe misspelled?)
Download error on https://pypi.org/simple/: [SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed (_ssl.c:726) -- Some packages may not be found!
No local packages or working download links found for setuptools-scm
  File "setup.py", line 41, in <module>
    main()
  File "setup.py", line 37, in main
    zip_safe=False,
  File "/sw/lib/python2.7/site-packages/setuptools/__init__.py", line 128, in setup
    _install_setup_requires(attrs)
  File "/sw/lib/python2.7/site-packages/setuptools/__init__.py", line 123, in _install_setup_requires
    dist.fetch_build_eggs(dist.setup_requires)
  File "/sw/lib/python2.7/site-packages/setuptools/dist.py", line 514, in fetch_build_eggs
    replace_conflicting=True,
  File "/sw/lib/python2.7/site-packages/pkg_resources/__init__.py", line 770, in resolve
    replace_conflicting=replace_conflicting
  File "/sw/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1053, in best_match
    return self.obtain(req, installer)
  File "/sw/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1065, in obtain
    return installer(requirement)
  File "/sw/lib/python2.7/site-packages/setuptools/dist.py", line 581, in fetch_build_egg
    return cmd.easy_install(req)
  File "/sw/lib/python2.7/site-packages/setuptools/command/easy_install.py", line 667, in easy_install
    raise DistutilsError(msg)
distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('setuptools-scm')
````

File: 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/setuptools-scm-py.info [hunk 4]
```diff
@@ -28,10 +28,17 @@ CompileScript: <<
 
 # pytest now depends on this package so we can't run tests anymore
 # without a circular dependency.
-#InfoTest: <<
-#   TestDepends: pytest-py%type_pkg[python], mercurial, git
-#   TestScript: PYTHONPATH=. %p/bin/py.test-%type_raw[python] testing || exit 2
-#<<
+InfoTest: <<
+    TestDepends: mercurial, git
```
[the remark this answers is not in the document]
dhomeier: It probably wants this [lz4-py](https://pypi.org/project/lz4/0.9.0), that's a fairly dated version. And then tries subsequently to install `setuptools==0.9.8` and `setuptools==12.0.1`, so I don't see how to satisfy this with a package dependency, even if I tried.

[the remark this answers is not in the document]
nieder: From Debian, changing the test command to this skips the download tests:
````
TESTS_TO_EXCLUDE="not setuptools_support and not test_pip_download"
TESTS_TO_EXCLUDE="$TESTS_TO_EXCLUDE and not test_pkginfo_noscmroot and not test_use_scm_version_callable"
PYTHONPATH=. %p/bin/python%type_raw[python] -m pytest -k "$TESTS_TO_EXCLUDE"
````

(but I think the 2nd set of exclusions can be skipped for us)
Results before:

````python
plugins: hypothesis-3.65.1
collected 102 items                                                                                                 

testing/test_basic_api.py .s........                                                      [  9%]
testing/test_file_finder.py ..............................                                [ 39%]
testing/test_functions.py .................                                               [ 55%]
testing/test_git.py ..............                                                        [ 69%]
testing/test_main.py .                                                                    [ 70%]
testing/test_mercurial.py ...............                                                 [ 85%]
testing/test_regressions.py ..F.s                                                         [ 90%]
testing/test_setuptools_support.py FFF                                                    [ 93%]
testing/test_version.py .......                                                           [100%]
````
After
````python
plugins: hypothesis-3.65.1
collected 102 items / 6 deselected                                                                                  

testing/test_basic_api.py .s........                                                       [ 10%]
testing/test_file_finder.py ..............................                                 [ 41%]
testing/test_functions.py .................                                                [ 59%]
testing/test_git.py ..............                                                         [ 73%]
testing/test_main.py .                                                                     [ 75%]
testing/test_mercurial.py ...............                                                  [ 90%]
testing/test_regressions.py .s                                                             [ 92%]
testing/test_version.py .......                                                            [100%]
````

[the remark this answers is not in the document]
Comment (nieder) [reply 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/setuptools-scm-py.info]: If updating to 3.0.2 will take some work, let's work on fixing our current 2.1.0 here, and then updating to 3.0.2 since nothing else in the current tree is requiring 3.0.2.

File: 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytest-py.info [hunk 5]
```diff
@@ -43,11 +50,17 @@ PatchScript: <<
 
 CompileScript: %p/bin/python%type_raw[python] setup.py build
 
+# tox will download and pip-install a gazillion of extra deps, even those already installed as packages.
```
nieder: This is for future reference: tox will try to run the generic executable `pytest` when running these tests, so it'll grab whatever version is installed via update-alternatives. This patch is needed to make it try to use the already installed pytest command for the right variant:
````
perl -pi -e 's|pytest --lsof|%p/bin/pytest-%type_raw[python] --lsof|g' tox.ini
perl -pi -e 's|pytest -n3|%p/bin/pytest-%type_raw[python] -n3|g' tox.ini
perl -pi -e 's|pytest -n8|%p/bin/pytest-%type_raw[python] -n8|g' tox.ini
perl -pi -e 's|pytest -ra|p%p/bin/ytest-%type_raw[python] -ra|g' tox.ini
perl -pi -e 's|pytest --doctest|%p/bin/pytest-%type_raw[python] --doctest|g' tox.ini
perl -pi -e 's|sphinx-build|%p/bin/sphinx-build%type_raw[python]|g' tox.ini
perl -pi -e 's|coverage run|%p/bin/coverage%type_raw[python] run|g' tox.ini
perl -pi -e 's|coverage report|%p/bin/coverage%type_raw[python] report|g' tox.ini
````
This avoids downloading some of the pip-install dependencies, but it needs pytest itself to already be installed (wonder if you can play with symlinks, PATH, and PYTHONPATH during TestScript to make an appropriate `pytest` in %b...)
!lart bad test setup.

Comment (dhomeier) [reply 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytest-py.info]: As far as I can see too is already doing this - using the `pytest` executable in `%b/.tox/py%type_pkg[python]/bin`, which I believe is installed from its own distribution previously created in `%b/.tox/dist/pytest-3.6.2.zip`.
Compared to this default
``` 
py36 installdeps: hypothesis>=3.56, nose, mock, requests
py36 inst: /scratch.noindex/fink.build/pytest-py36-3.6.2-1/pytest-3.6.2/.tox/dist/pytest-3.6.2.zip
py36 installed: atomicwrites==1.1.5,attrs==18.1.0,certifi==2018.4.16,chardet==3.0.4,coverage==4.5.1,hypothesis==3.66.14,idna==2.7,mock==2.0.0,more-itertools==4.2.0,nose==1.3.7,pbr==4.2.0,pluggy==0.6.0,py==1.5.4,pytest==3.6.2,requests==2.19.1,six==1.11.0,urllib3==1.23
py36 runtests: PYTHONHASHSEED='3592075579'
py36 runtests: commands[0] | pytest --lsof -ra testing
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.6.6, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
rootdir: /scratch.noindex/fink.build/pytest-py36-3.6.2-1/pytest-3.6.2, inifile: tox.ini
plugins: hypothesis-3.66.14
collected 2115 items

testing/acceptance_test.py .................................................x................                                                            [  3%]
testing/deprecated_test.py ..............                                                                                                                [  3%]
...
============================================== 2024 passed, 80 skipped, 11 xfailed, 3 warnings in 243.87 seconds ===============================================

```
which can n.b. not be using `/sw/etc/alternatives/pytest@ -> /sw/bin/pytest-3.7`, since there is no `hypothesis-py37` installed,
forcing `tox.ini` to use the already installed `/sw/bin/pytest-py%type-raw[python]` lets it load a number of additional plugins, collection more tests but with many more failing:
```
py36 installdeps: hypothesis>=3.56, nose, mock, requests
py36 inst: /scratch.noindex/fink.build/pytest-py36-3.6.2-1/pytest-3.6.2/.tox/dist/pytest-3.6.2.zip
py36 installed: atomicwrites==1.1.5,attrs==18.1.0,certifi==2018.4.16,chardet==3.0.4,coverage==4.5.1,hypothesis==3.66.14,idna==2.7,mock==2.0.0,more-itertools==4.2.0,nose==1.3.7,pbr==4.2.0,pluggy==0.6.0,py==1.5.4,pytest==3.6.2,requests==2.19.1,six==1.11.0,urllib3==1.23
py36 runtests: PYTHONHASHSEED='2390453554'
py36 runtests: commands[0] | /sw/bin/pytest-3.6 --lsof -ra testing
===================================================================== test session starts ======================================================================
platform darwin -- Python 3.6.6, pytest-3.6.2, py-1.5.4, pluggy-0.6.0
PyQt5 5.10.1 -- Qt runtime 5.7.1 -- Qt compiled 5.7.1
Matplotlib: 2.2.2
Freetype: 2.6.1
rootdir: /scratch.noindex/fink.build/pytest-py36-3.6.2-1/pytest-3.6.2, inifile: tox.ini
plugins: remotedata-0.2.0, relaxed-1.1.4, qt-2.3.1, openfiles-0.2.0, mpl-0.9, mock-1.10.0, doctestplus-0.1.2, arraydiff-0.2, hypothesis-3.65.1
collected 2654 items

testing/acceptance_test.py ......F.......................................FF.x............F.........F......................F................FF.x........F [  2%]
...F...                                                                                                                                                  [  2%]
testing/deprecated_test.py ............................
```

I don't see the patch to `tox.ini` reducing the number of downloaded packages, but it introduces loads of test failures (and is using whatever pytest version is already installed, not what has just been built).

File: 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/setuptools-scm-py.info (hunk 4 above)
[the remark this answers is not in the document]
Comment (dhomeier) [reply 10.9-libcxx/stable/main/finkinfo/libs/pythonmods/setuptools-scm-py.info]: It's rather 2.1.0 causing troubles. 3.0 works by itself, and making 1.15 a `BuildConflict` lets `fink -m update` progress smoothly. So I prefer the latter, although I am not 100 % comfortable not knowing why 2.1.0 passes on your system, but not for me...

**Comment by nieder** [comment]
These look good to me. Someone else should go and try to build the series to make sure I wasn't just lucky with my 10.11 setup.

It's a shame what's needed to deal with so many of the tests in these packages, but it's not something we can help.

Note: attrs-py.info now has a conflict with HEAD.

**Comment by danielj7** [comment]
I fixed the conflict with HEAD. I'm now testing the updates.

**Comment by danielj7** [comment]
Building py-py37 fails with:
/sw/bin/python3.7 setup.py build --offline
/sw/lib/python3.7/site-packages/setuptools_scm/version.py:191: UserWarning: meta invoked without explicit configuration, will use defaults where required.
  "meta invoked without explicit configuration,"
usage: setup.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
   or: setup.py --help [cmd1 cmd2 ...]
   or: setup.py --help-commands
   or: setup.py cmd --help

error: option --offline not recognized

**Comment by nieder** [comment]
Same here. The py-pyXX build works if I remove the newly added --offline flag in CompileScript.

**Comment by danielj7** [comment]
That works for me too.

I then ran into other issues. After installing pytest, I ran
```
fink -m rebuild atomicwrites-py37 invoke-py37 py-py37 setuptools-scm-py37
```
but invoke-py failed because it is missing a test dep on pytest-relaxed-py:
```
ModuleNotFoundError: No module named 'pytest_relaxed'
```
Then when I tried to build that, pytest-runner-py tried to download and install setuptools-scm even though it was already installed: 
```
/sw/bin/python3.7 setup.py build
Download error on https://pypi.org/simple/setuptools_scm/: [Errno 64] Host is down -- Some packages may not be found!
Download error on https://pypi.org/simple/setuptools-scm/: [Errno 64] Host is down -- Some packages may not be found!
Download error on https://pypi.org/simple/: [Errno 64] Host is down -- Some packages may not be found!
No local packages or working download links found for setuptools_scm>=1.15.0
```
Note that my firewall was set to block downloading during building which is why it failed.

danielj7:
Ok, so when setuptools-scm was built the first time, it's version number was set to 0.0.0 which is why it wasn't found when pytest-runner was building. When I rebuilt setuptools-scm, it had the correct version number of 3.0.5. Don't know what that's about.

danielj7:
Ok, I've confirmed that setuptools-scm only builds with the correct version number when it's already installed. Removing it and rebuilding always has a version of 0.0.0. This is not a good thing. Probably due to `/sw/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'use_scm_version'`. So it's using itself to determine it's version.

danielj7:
Seems it's fixed in setuptools-scm 3.0.6.

danielj7:
I merged this with a couple of fixes for the tests. Everything builds fine and tests work.

PR #203, opened 2018-07-16
asteval-py: new numpy related python package

kamischi:


dhomeier:
Thanks for the contribution. Some comments:
- when you find you need to make some corrections, it would be far easier and more organised to commit the updated .info to the existing branch and push to #202 ;-)
- `setup.py test` does not perform any tests; `./tests/` require pytest.
Changing `InfoTest` to
```
    pytest-py%type_pkg[python],
    six-py%type_pkg[python]
  <<
  TestScript: <<
    PYTHONPATH=%b/build/lib %p/bin/pytest-%type_raw[python] tests || exit 2
    find build/lib -name '*.pyc' -delete
  <<
```
lets it (successfully) run 53 tests.
- (pinging core maintainers) is it a policy or just a recommendation to run `setup.py` with the full path to Fink's Python executables `%p/bin`...?
- The Compile/Install/TestScripts don't need an extra shell `#!/bin/bash -ev` (but it does not exactly do harm of course).
- `INSTALL` and `MANIFEST.in` don't contain any info useful or relevant to using the fink package and can probably stripped from the `DocFiles`.
- I was also able to build a `py37` variant, but this will need #199 for the tests.

nieder:
* In theory, all builds will have %p/bin in front of /usr/bin in $PATH, so the full path to Fink's python exe is not needed. However, using the full path will help prevent any accidental usage of the wrong python interpreter.
* In TestScript, does running pytest-pyXX with the `-B` flag help get rid of the .pyc files? That works when tests are run directly with the pythonX.Y interpreter, but I don't know if pytest-pyXX also honors that flag.

kamischi:
New version basically following your suggestions.
- Testscript changed. -B is not available. I did not find any equivalent.
- '#!/bin/bash -ev' removed
- full path to Fink's Python executables %p/bin/python...
- Files INSTALL and MANIFEST.in removed from docfiles

dhomeier:
I found one can use `setup.py pytest` when using the `--addopts` flag. Thus
```
  TestScript: <<
    PYTHONPATH=%b/build/lib python%type_raw[python] -B setup.py pytest --addopts tests || exit 2
  <<
```
does not create any .pyc files (besides this always seems to load the module from the source directory regardless of the setting of `PYTHONPATH`, for better or worse).

kamischi:
Doesn't seem to work:
...
error: invalid command 'pytest'

dhomeier:
Ah, that would explain why it didn't run under Python 3.7.

kamischi:
dhomeier: btw, any chance to update uncertainties-py to 3.0.2?

kamischi:
Great work, guys. Thanks for your help.

dhomeier:
Submitted a new `uncertainties-py` in #204.

kamischi:
Another big thank you.

Issue #162, opened 2018-05-12
libnewt: whiptcl.so not installed on 10.13


nieder:
On 10.13, whiptcl.so is not being installed and causes a failure when populating the `newt-tcl` SplitOff
````
mv /sw/src/fink.build/root-libnewt0-shlibs-0.52.19-1.1/sw/lib/whiptcl.so /sw/src/fink.build/root-newt-tcl-0.52.19-1.1/sw/lib/whiptcl/
mv: rename /sw/src/fink.build/root-libnewt0-shlibs-0.52.19-1.1/sw/lib/whiptcl.so to /sw/src/fink.build/root-newt-tcl-0.52.19-1.1/sw/lib/whiptcl/whiptcl.so: No such file or directory
````
whiptcl.so is built, just not installed. This is the relevant part of the install process:
````
ln -sf libnewt.0.52.19.dylib /sw/src/fink.build/root-libnewt0-shlibs-0.52.19-1.1//sw/lib/libnewt.0.dylib
[ -n "" ] && install -m 755 whiptcl.so /sw/src/fink.build/root-libnewt0-shlibs-0.52.19-1.1//sw/lib || :
````
On 10.11, that initial comparison is and it succeeds in putting whiptcl.so into %i/lib
````
[ -n "whiptcl.so" ] && install -m 755 whiptcl.so /sw/build.build/root-libnewt0-shlibs-0.52.19-1.1//sw/lib || :
````

**Comment by dmacks** [comment]
Earlier on,  I see:
```
./autogen.sh
configure.ac: error: no proper invocation of AM_INIT_AUTOMAKE was found.
configure.ac: You should verify that configure.ac invokes AM_INIT_AUTOMAKE,
configure.ac: that aclocal.m4 is present in the top-level directory,
configure.ac: and that aclocal.m4 was recently regenerated (using aclocal)
automake: error: no 'Makefile.am' found for any configure output
```
which means any patches against configure.ac and Makefile.am are likely being ignored.

But also:
```
checking for tcl configuration... NOT found
```
It makes a bit of sense that failing to detect some piece of tcl would inhibit complete installation of some tcl thing. The configure script has all kinds of hardcoded guessing of tcl paths that seem likely to mix up apple's and fink's. One of those looks for tclConfig.sh, which *until 10.13* was in /usr/lib, but now is not (/System/Library/Frameworks/Tcl.framework is a backward-compatible location of it).

dmacks:
[...]

PR #204, opened 2018-07-19
Additional updates and py36/py37 variants for pythonmods

dhomeier:
In the update #140 a single (non-deprecated) package was left without a `py36` variant, `newt-py`, which is added here together with its `py37` counterpart. Unfortunately this is still part of the monolithic `libnewt0` build, so required two additional `SplitOff`s. Should this rather be restricted to a single Python3 variant?
The `pysvn` update to the latest upstream resolves a conflict `1.9.0` had with our current `svn-1.10`.
`uncertainties` update thrown in on request.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
@@ -1,6 +1,6 @@
 Package: libnewt0-shlibs
-Version: 0.52.19
-Revision: 1.3
+Version: 0.52.20
+Revision: 1
 ###
 BuildDepends: <<
 	fink (>= 0.32),
@@ -13,6 +13,8 @@ BuildDepends: <<
 	tcltk-dev,
 	python27,
 	python35,
+	python36,
+	python37,
 	autoconf2.6,
 	automake1.15-core,
 	docbook-utils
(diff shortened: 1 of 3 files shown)


dhomeier:
@TheSin-, would you be able to test the `libnewt0` update on 10.11, then I could merge this and hopefully close #162?

TheSin-:
make the revisions 5.1 please, and I'd rather if you just picked one version of python3. But it's a huge flaw in fink sadly that isn't easily over come.  So just change the revision so that it denotes the Debian level please and I"m okay with it.

dhomeier:
Seems more a limitation of the build setup - moving the Python modules to their own newt-py.info would likely require a complete rebuild of the library for each variant.
But I have not found any versioned Python packages depending on this, so sticking with only `py27` and `py36` should be fine (or maybe, since there are no dependencies [yet], go to `py37` right away?)
Which revision 5.1 do you mean, or where would I find the "Debian level"?

TheSin-:
yeah sticking with 2 variants would be great I think, what ever is the current stable 3.x and 2.7.

on the Info file, I make sure when I'm using Debian patches to include it, currently you have it set to 1

Revision: 5.1

Revision: <debian patch level>.<fink patch level>

10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
```diff
@@ -1,6 +1,6 @@
 Package: libnewt0-shlibs
-Version: 0.52.19
-Revision: 1.3
+Version: 0.52.20
+Revision: 1
```
TheSin-: make this 5.1 please

dhomeier:
The Python 3.x are all stable. Thinking of it as a UI provided for user scripts, 3.6 is probably still the better option, as 3.7 still has a much smaller module base (50-60 packages vs. nearly 300).

Had to check that decimal `Revision` numbers are actually standard package syntax, but if it helps to keep the Debian patch sets organised, fine with me.

danielj7:
I committed the pysvn update.

10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
nieder: @dhomeier ping on @TheSin- 's request to change the Revision.

PR #205, opened 2018-07-20
mutt 1.10.1

bcully:


nieder:
* 1.10.1 seems to be libidn2 capable (`--with-idn2`).
* can a fink-package-precedence check be added to confirm library header use?

Otherwise, 1.10.1 as submitted looks OK.

PR #206, opened 2018-07-20
give up maintainership of packages

mroi:
Unfortunately I am no longer actively using Fink, so I am no longer able to test these packages or provide updates. I would like to document this situation publicly so that someone else will hopefully see the opening and take over. I am of course always available for questions.


dmacks:
Thank your for your years of work on these! I see you added yourself as former maintainer that in some cases. Let me know if you want to generally be listed (and with or without email) and we can update it on all of them when committing.

mroi:
You are very welcome. It feels a bit sad to turn away from Fink as it has served me well over the years. But lately I used it increasingly less, so it felt more of a burden to keep my installation going.

Listing me with email for questions would be fine, but better use reactorcontrol@icloud.com, because I am phasing out my SourceForge address. If you prefer to leave the old dists untouched, that would also be fine with me.

Thanks for all the wonderful work on Fink.

PR #214, opened 2018-08-05
gir2pascal: new package

kamischi:
gir2pascal converts gobject-introspection (*.gir) xml files into Pascal files.


nieder:
Need a LICENSE file. Since upstream doesn't include a standalone one, can use something like
`head -n 17 gir2pascal.lpr | tail -n 15 > LICENSE` to generate one and then stick it in the `DocFiles` field.

Issue #121, opened 2018-03-30
fink-bld SHELL conflicts with numpy.distutils


dhomeier:
Hi,

this problem has come up a couple of times in the past months, but tends to reappear (see links below)...
The issue, briefly, is that numpy's own `distutils` module, and any package using `numpy.distutils`, by default spawns all subprocesses in the shell `$SHELL` (in `numpy/distutils/exec_command.py`):
```
    if os.name == 'posix' and use_shell:
        # On POSIX, subprocess always uses /bin/sh, override
        sh = os.environ.get('SHELL', '/bin/sh')
        if is_sequence(command):
            command = [sh, '-c', ' '.join(command)]
        else:
            command = [sh, '-c', command]
```
Any suggestions how to deal with this in a more sustainable way (and if it's just adding a new `SetSHELL` field for convenience)? It's hard to track down if you are not aware of the problem, as even
```
DISTUTILS_DEBUG=1 %p/bin/python2.7 -d setup.py build
```
only gives you a vague pointer to the source of the failure.
I am considering reporting the issue upstream so numpy will at least test explicitly if `os.environ.get('SHELL')` produces a working command, but they might respond that building with `SHELL=/usr/bin/false` is a really bad idea™...

- [numexpr-py34-2.6.4-1 failed](https://sourceforge.net/p/fink/mailman/message/36280059/)

- [broken python packages](https://sourceforge.net/p/fink/mailman/message/36104712/)

- [numpy test failures](https://sourceforge.net/p/fink/package-submissions/4930/#467b)

dmacks:
It's not just us. FreeBSD has the same problem and patched in in their distro. See [FreeBSD Bug #225156](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=225156)

nieder:
I locally patched my numpy-py35 to include the SHELL fix from FreeBSD and then tried to build numexpr-py35 (all on 10.13/Xcode9.4) against the patched numpy-py. The numexpr-py build succeeds with the patched numpy-py. This patch seems like the correct way to go.
````patch
--- numpy/distutils/exec_command.py.orig	2017-09-29 20:10:10 UTC
+++ numpy/distutils/exec_command.py
@@ -231,7 +231,7 @@ def _exec_command(command, use_shell=Non
 
     if os.name == 'posix' and use_shell:
         # On POSIX, subprocess always uses /bin/sh, override
-        sh = os.environ.get('SHELL', '/bin/sh')
[...]

PR #218, opened 2018-08-10
fix numpy.distutils and SHELL issue (#121)

nieder:
Fixes the problem with the environment SHELL from fink-bld (/usr/bin/false) not being a real shell.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/matplotlib-basemap-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/matplotlib-basemap-py.info
@@ -2,17 +2,19 @@ Info2: <<
 
 Package: matplotlib-basemap-py%type_pkg[python]
 Version: 1.0.2
-Revision: 2
+Revision: 3
 Maintainer: Kurt Schwehr <goatbar@users.sourceforge.net>
 Depends: <<
-  python%type_pkg[python], matplotlib-py%type_pkg[python] (>= 1.4.3),
-  httplib2-py%type_pkg[python],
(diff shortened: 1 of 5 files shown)


dmacks:
Could you document in numpy-py.info itself why the special version-dependency on numpy-py? Otherwise lgtm. matplotlib-basemap-py.info could also benefit from that tweak.

dhomeier:
Same for cython-py.info and my numba-py submission #207 (I could take care of that of course).
Alternatively I had thought about a patch that will first check if `$SHELL` is a functioning shell and only afterwards revert to `/bin/sh`; something like
```
        sh = os.environ.get('SHELL', '/bin/sh')
        if subprocess.run([sh, '-c', 'exit 0'], timeout=100).returncode != 0:
            sh = '/bin/sh'
```
would of course not change anything for the fink builds, but restore the original behaviour in the (remote?) case a user wants to build some extension using `numpy.distutils` with their own `$SHELL`. It's still not clear to me if that would be POSIX-compliant, so might not be worth the hassle...

nieder:
matplotlib-basemap-py and mayavi2-py (@costabel) now also patched to build.

dmacks:
@dhomeier, upstream numpy has a ticket where they are discussing various solutions. Feel free to join them at numpy/numpy/issues/9970

nieder:
Pushing since it cleans up an issue that propagates as FTBFS for several packages.

dmacks:
Upstream switched to handling it in numpy/numpy/issues/11824

PR #220, opened 2018-08-10
c-blosc: New package. Compression library.

kamischi:
Hi.

I am not really sure about the license, as there is a mix for different parts. My best guess is BSD, but OSI-approved might also be right. Any opinions?

This packages aims to be used in pytables.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libblosc1-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libblosc1-shlibs.info
@@ -1,0 +1,95 @@
+Package: libblosc1-shlibs
+Version: 1.14.4
+Revision: 1
+Description: Blocking, shuffling compression library
+License: BSD
+
+Source: https://github.com/Blosc/c-blosc/archive/v%v.tar.gz
+Source-MD5: e80dfc71e4cba03b8d01ed0876547ffe
+SourceRename: c-blosc-%v.tar.gz
+
+BuildDepends: <<
+  cmake,
+  fink-package-precedence
+<<
+
+CompileScript: <<
+#!/bin/sh -ev
+  mkdir -p build
+  cd build
+  cmake \
+    -DCMAKE_BUILD_TYPE=Release \
+    -DCMAKE_C_FLAGS:STRING=-MD \
+    -DCMAKE_INSTALL_PREFIX=%i \
+    -DCMAKE_INSTALL_NAME_DIR=%p/lib \
+    -DCMAKE_PREFIX_PATH=%p \
+    ..
+
+  cmake --build .
+
+# replace install path by prefix
+  sed -i.bak 's|%i|%p|g' blosc.pc
+
+  cd ..
+  fink-package-precedence --depfile-ext='\.d' .
+<<
+
+InfoTest: <<
+  TestScript: <<
+  #!/bin/sh -ev
+    cd build
+    PATH=%b/build/bin:%p/bin:/usr/bin:/bin \
+      DYLD_LIBRARY_PATH=%b/build/lib \
+      ctest || exit 2
+  <<
+<<
+
+InstallScript: <<
+#!/bin/sh -ev
+  cd build
+  cmake --build . --target install
+
+  mkdir %i/bin
+  install bench/bench %i/bin
+<<
+
+DocFiles: ANNOUNCE.rst LICENSES README* RELEASE_NOTES.rst THANKS.rst
+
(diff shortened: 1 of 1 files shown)


dhomeier:
`pytables` ships with it (currently one minor version behind) and thus comes with the same licenses for the most part - thus BSD seems to be safe.

kamischi:
Ups. I completely missed that `pytables` has blosc included. I only saw the linker error about libblosc. So, is it at all helpful for `pytables`, or does it only interfere with building?

dhomeier:
What linker error did you get? I haven't tested the package yet, just checked if `pytables` might install conflicting files, but afaics it links the blosc stuff into its `tableextension.cpython-*.so` extension.
Not sure how easy or difficult it would be to have it link an external `cblosc` installation, or if it's worth the effort. But this could still be very useful to make `blosc` compression generally available at the HDF5 level for using it with e.g. `h5py`, outlined in the docs on [third-party filters](http://portal.hdfgroup.org/display/support/Contributions) - but I've never seriously tried to get this to work.

kamischi:
Without `libblosc1-shlibs` and `libblosc1-dev`, `fink rebuild pytables-py36` gives this non-fatal error:

```
...
* Skipping detection of LZO 1 since LZO 2 has already been found.
* Found bzip2 headers at ``/sw/include``, library at ``/sw/lib``.
/private/tmp/blosc_list_compressorseq81wzx3.c:1:1: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
main (int argc, char **argv) {
^
/private/tmp/blosc_list_compressorseq81wzx3.c:2:5: warning: implicit declaration of function 'blosc_list_compressors' is invalid in C99 [-Wimplicit-function-declaration]
    blosc_list_compressors();
    ^
2 warnings generated.
ld: library not found for -lblosc
* Could not find blosc headers and library; using internal sources.
Setting compiler flag: -march=native
running install
...
```
After the installation of `libblosc1-shlibs` and `libblosc1-dev`, the rebuild is without these errors, but in maintainer mode, it fails in the test phase with:
```
...
...................................................................../sw/src/fink.build/pytables-py36-3.2.2-3/PyTables-v.3.2.2/build/lib.macosx-10.13-x86_64-3.6/tables/atom.py:570: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  for arg in inspect.getargspec(self.__init__)[0]
.............................................................................................ssss........./sw/src/fink.build/pytables-py36-3.2.2-3/PyTables-v.3.2.2/build/lib.macosx-10.13-x86_64-3.6/tables/atom.py:570: DeprecationWarning: inspect.getargspec() is deprecated, use inspect.signature() or inspect.getfullargspec()
  for arg in inspect.getargspec(self.__init__)[0]
/tmp/fink.jyNQ6: line 4:  7327 Segmentation fault: 11  /sw/bin/python3.6 -B -c 'import tables, sys; r=tables.test(); sys.exit(r)'
### execution of /tmp/fink.jyNQ6 failed, exit code 139
### execution of /tmp/fink.vCzcs failed, exit code 139
Removing runtime build-lock...
Removing build-lock package...
...
```

dhomeier:
Well, the linker message is really a warning as the build falls back to the internal sources, but that means the external library will be used automatically, if present.
But I see that Fink's current `pytables` 3.2.2 actually comes with blosc 1.4.4, so the version mismatch might be too large. I already had a 3.4 version for testing on my system; the latest release 3.4.4 builds and tests successfully against your package. I could commit it to this PR.

Is there a reason you named the package `libblosc1-*`? The installed library name is `libblosc.%v.dylib`, the source directory `c-blosc-%v`, so `libblosc` would seem the natural choice to me...

kamischi:
I somehow find`libblosc1-*` nicer, when there will be a compatibility version 2.* in the future. But that is all. I would not really mind to change it to `libblosc-shlibs` and `libblosc-dev`.

dmacks:
Including the number portion of the install_name in the package-name for a "-shlibs" package and its companion headers-package (either plain-name or "-dev") is standard for future possible other versions. It is part of fink's packaging policy.

nieder:
%v1.14.4 builds and tests fine on 10.13.
It also worked when building pytables from #372:
Pytables-3.5.1 build output w/out our libblosc1: `Blosc version:       1.16.3 (2019-03-08)`
With this libblosc installed: `Blosc version:       1.14.4 (2018-07-30)`

So clearly finding it and using it. I don't see the test failures @kamischi saw in Aug 13 w/ pytables-3.2.2

I think this is probably good to commit.

PR #221, opened 2018-08-12
Fix ddd build on Xcode9

nieder:
ddd fails to build on Xcode9 with unknown type name a_class.
This is the upstream submission to fix it.


kamischi:
Successful build, install and run on 10.13. Nice.

PR #226, opened 2018-08-14
binutils: Update to new upstream version (2.31.1)

kamischi:


nieder:
Do you want to bump the version for `x86-64-binutils-default` ?

kamischi:
`x86-64-binutils-default` added.

PR #227, opened 2018-08-15
py2app, altgraph, modulegraph, macholib: update to upstream version

kamischi:
Update of version, source and homepage and extended to python 3 versions.  Includes the dependencies. Removal of even older version (0.6.3) in category devel. Dependencies are fine with 3.7. py2app still misses sphinx-py37.


kamischi:
Thanks for review and merge.

PR #231, opened 2018-08-19
eccodes-2.8.2-1: Upstream update

remkos:
This pull request is to introduce the new upstream update version 2.8.2 (from 2.7.3 as in Fink).
(I skipped 2.8.0 because some tests failed).

This info-file also improves the testing. In the past the package needed to be installed in %p first before it could be tested. Now with a bit of use of install_name_tool, it can be tested in place.

It compiled and tested correctly with "fink -m build", even after first removing the old package. This shows that it tested the just compiled/not yet installed version.

I hope I can get your approval.


remkos:
Thanks, @nieder for your suggestion. Indeed that runs significantly faster. I hope to remember this for future builds that involve cmake.
So now `fink-package-precedence` is now run in the source directory.
Please approve so it can be merged.

PR #232, opened 2018-08-20
Upstream updates to NetCDF tools CDO and NCO

remkos:
cdo-1.9.5-1: upstream update
nco-4.7.6-1: upstream update
All tested successfully with "fink -m build"


dmacks:
There are are few confusing things in the documentation of nco. It is described as "The NetCDF Operators, using netcdf-4", but its dependencies are now netcdf-c13. Later, "build nco with a static library only", but the build is ```ConfigureParams:--enable-shared --disable-static```. I understand the perl one-liner to fix ```-lnco```. I recognize the numbers-change one-liner as to avoid flat namespace; might want to annotate that (and maybe separate it out as its own one-liner for clarity?). Look for "Puma" in tons of other .info for example).

dmacks:
Looking at the build of nco on 10.13:
```
checking for XML_ParserCreate in -lexpat... yes
checking for /sw/share/udunits/udunits2.xml... no
configure: WARNING: file /sw/share/udunits/udunits2.xml was not found but will still attempt to build NCO with UDUnits2 support.
configure: NB: Following tests of UDUnits2 library and header must succeed for UDUnits2 support:
checking for ut_read_xml in -ludunits2... yes
checking for /sw/include/udunits2.h... yes
configure: UDUnits2 support enabled
```
but there is no ```BuildDepends:expat1``` to make sure the -lexpat test passes for everyone. There is a ```Depends:expat1-shlibs``` and I do see linkage against that library during the build process.

The WARNINGS message is because that .xml file is in the "udunits2" package, but there are only dependencies on udunits2-{dev,shlibs}. Should it be added to Depends: ?

remkos:
Thank you @dmacks for your thorough review. I'm always amazed you find things that have been in these .info files but were wrong for quite some time.

* cdo:
-- Thank you for pointing out that I needed to repeat the switched, I was totally unaware.
-- Once I fixed that, one of the Gradsdes.test failed with standard gcc, because of a bug in the cdo module "Gradsdes". I fixed that in the patch and reported it upstream.

* nco:
-- It *is* using netcdf-4 (i.e. NetCDF version 4), which is package "netcdf-c13" in Fink.
-- I've added "expat1" in BuildDepends
-- I had not realised that udunits2.xml was in the udunits2 package, not in udunits2-shlibs. I cannot imagine how that library works without the .xml file, so I expected it there. In any case, udunits2 is added in Depends.
-- The message in DescPort about using static library was old. In fact @akhansen and I discussed this in the past and realised the configure file was forcing the creating of a static library for Darwin, and then creating many large executables. So we decided then on a private shared library, as we have been doing since nco-4.7.0-1. Thus, no more ncoXXX-dev and ncoXXX-shlibs since then.
-- Removed the Conflict/Replace of ncoXXX-dev since that version.
-- Removed the optional texinfo-legacy from BuildDepends
-- Split and annotated the "perl -pi" line as per your suggestion.

All tested again on 10.13.

10.9-libcxx/stable/main/finkinfo/sci/nco.info
```diff
@@ -91,21 +92,24 @@ Replaces: <<
...
+Source-MD5: 84296a0ec44b9a58dab198fda9fac7f0
+Source-Checksum: SHA1(e39bdb3d50df0c7bb78c808e5b69907f9d2b3f75)
 
 # Patch Phase:
 PatchScript:  <<
 #!/bin/sh -ev
+	# Make sure to link against newly built libnco.dylib
 	perl -pi -e 's|-lnco|%b/src/nco/.libs/libnco.dylib|' src/nco++/Makefile.in
-	perl -pi -e "s/(apple.*)enable_shared='no'/\1/; s/(10\.\[012\])(\*)/\1\,.\2/" configure
+	# Patch configure to force enable_shared, since it suggests that Darwin cannot create shared linraries
```
schwehr: linraries → libraries

10.9-libcxx/stable/main/finkinfo/sci/nco.info
```diff
@@ -91,21 +92,24 @@ Replaces: <<
...
+	perl -pi -e "s/(apple.*)enable_shared='no'/\1/" configure
+	# Patch configure to not link like Puma on Yosemite
+        perl -pi -e 's|10\.\[012\]\*|10.[012][,.]*|g' configure
```
schwehr: "not link like Puma on Yosemite"  What does this mean?

dmacks: Thanks for all the fixes and explanations!

"Puma" is OS X 10.1. There is a string-regexp bug in an old version of libtool: OS X version string "10.10" (10.10 is "Yosemite") and later versions all match "10.1", so it uses the type of linking that was correct on 10.1 rather than what is correct on modern platforms. The patch changes the string-regexp to match "10.1." instead. This was fixed long ago in libtool, but the upstream source distributors have to upgrade their libtool as part of their distribution process (it is statically included in the tarball, not automatic on the machine where one compiles).

10.9-libcxx/stable/main/finkinfo/sci/nco.info
remkos: That you for pointing out the typo. This is fixed now in the latest commit (52426b2)

10.9-libcxx/stable/main/finkinfo/sci/nco.info
remkos: @schwehr: I hope the explanation of @dmacks is sufficient.

remkos:
Looking forward to approval with the above change.

PR #233, opened 2018-08-21
expat1: new version

dmacks:
Per https://sourceforge.net/p/fink/mailman/message/36396563/

Should we scrap the static lib? I've left it due to inertia, but .a are generally unused when there is a .dylib (and fully static linking isn't even possible regardless)


danielj7:
I would say get rid of .a. Nothing should be using it anyway.

PR #236, opened 2018-08-25
invoke-py: update to newer upstream version (1.1.1)

kamischi:
no further change.


kamischi:
Having in mind to reduce the number of branches, I assumed that the three commits could be treated separately, but the simplest way seems to be pull requests from 3 separate branches. Is this right?

kamischi:
closing and later creating 3 separate branches and pull requests.

PR #238, opened 2018-08-29
New upstream release paramiko 2.4.1

bcully:


nieder:
Builds fine.

However, the dependencies `bcrypt-pyXX` and `nacl-pyXX` both need `cffi-pyXX` as a Depends: in their own packages:
````
distutils.errors.DistutilsError: Could not find suitable distribution for Requirement.parse('cffi>=1.4.1'
````

bcully:
Added #265 for bcrypt-py. I was going to update nacl-py as well but that will depend on #263

PR #240, opened 2018-08-29
invoke-py: update to new upstream version (1.1.1)

kamischi:


nieder:
Can we document the PatchScript and add a note to TestScript that the pytest_relaxed conditional is to avoid circular deps? These changes can be done quickly in the browser.
Otherwise, looks fine.

PR #253, opened 2018-09-06
pexpect-py: update and add py37 variant

kamischi:
also move from libs to libs/pythonmods.


kamischi:
Thanks guys. You're great ;-)

PR #258, opened 2018-09-06
tox-py: update to newer upstream version (3.2.1)

kamischi:
Changes:
- patches in Patchfile adjusted
- moved some dependencies to TestDepends. I saw no reason for Depends: pytest-py and BuildDepends: setuptools-scm-py. So, I removed them. But, is that ok?
- Added LICENSE to Docfiles:
- Corrected/added the version patches for tox.ini

One problem is that the test do not succeed as long as tox is not yet installed. No idea, how to fix that.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/filelock-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/filelock-py.info
@@ -1,0 +1,33 @@
+Info2: <<
+
+Package: filelock-py%type_pkg[python]
+Version: 3.0.10
+Revision: 1
+License: OSI-Approved
+Type: python (2.7 3.4 3.5 3.6 3.7)
+
+Depends: <<
+  pv-py%type_pkg[python],
+  python%type_pkg[python]
+<<
+BuildDepends: setuptools-tng-py%type_pkg[python]
+
+Source: https://files.pythonhosted.org/packages/source/f/filelock/filelock-%v.tar.gz
+Source-MD5: df0006d2b1ec96473bfc0927de123aa6
+
+CompileScript: %p/bin/python%type_raw[python] setup.py build
+
+InstallScript: %p/bin/python%type_raw[python] setup.py install --root=%d
+
+DocFiles: LICENSE.rst README.rst PKG-INFO
+
+InfoTest: <<
+  TestScript: %p/bin/python%type_raw[python] setup.py test || exit 2
+<<
+
+Description: Platform independent file lock
+
+Homepage: https://github.com/benediktschmitt/py-filelock
+Maintainer: Karl-Michael Schindler <karl-michael.schindler@web.de>
+# Info2
+<<
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
@@ -1,0 +1,37 @@
+Info2: <<
+
+Package: pv-py%type_pkg[python]
+Version: 0.1.3
+Revision: 1
+License: OSI-Approved
+Type: python (2.7 3.4 3.5 3.6 3.7)
+
+Depends: <<
+  webob-py%type_pkg[python],
+  python%type_pkg[python]
+<<
+BuildDepends: setuptools-tng-py%type_pkg[python]
+
+Source: https://files.pythonhosted.org/packages/source/p/pv/pv-%v.tar.gz
+Source-MD5: 733f34c529ec5632db5b3049140cdde9
+
+CompileScript: %p/bin/python%type_raw[python] setup.py build
+
+InstallScript: %p/bin/python%type_raw[python] setup.py install --root=%d
+
+DocFiles: PKG-INFO
+
+InfoTest: <<
+#  TestDepends: <<
+#    webob-py%type_pkg[python],
+#    simplejson-py%type_pkg[python]
+#  <<
+#  TestScript: %p/bin/python%type_raw[python] setup.py test || exit 2
+<<
+
+Description: Minimal wsgi framework
+
+Homepage: https://github.com/tetsuo/pv
+Maintainer: Karl-Michael Schindler <karl-michael.schindler@web.de>
+# Info2
+<<
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
@@ -22,10 +22,7 @@ InstallScript: %p/bin/python%type_raw[python] setup.py install --root=%d
 DocFiles: PKG-INFO
 
 InfoTest: <<
-#  TestDepends: <<
-#    webob-py%type_pkg[python],
-#    simplejson-py%type_pkg[python]
-#  <<
+#  TestDepends: simplejson-py%type_pkg[python]
 #  TestScript: %p/bin/python%type_raw[python] setup.py test || exit 2
 <<
 
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pv-py.info
@@ -15,6 +15,11 @@ BuildDepends: setuptools-tng-py%type_pkg[python]
 Source: https://files.pythonhosted.org/packages/source/p/pv/pv-%v.tar.gz
 Source-MD5: 733f34c529ec5632db5b3049140cdde9
 
+PatchScript: <<
+  perl -pi -e 's|except Exception, e|except Exception as e|g' pv/web.py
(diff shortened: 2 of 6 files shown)


kamischi:
tox-py: update to newer upstream version (3.3.0). Also added toml-py, which is needed for tests. The tests still only work if tox-py is already installed.

nieder:
When testing this and future releases, at least on 3.0.0, pip-py is needed to pass 11 different tests, so that should be added to TestDepends.

kamischi:
Is it also required to put the changes in docs/changelog.rst to a patch file?

dhomeier:
I don't think there is a strict rule for this, but given the number and nature of changes having them in `tox-py.patch` might be cleaner. But is it actually necessary to purge all the non-ASCII chars? I would have thought this should not be an issue in .rst.

kamischi:
If i comment the first perl -pe -e ... replace, the test with tests/unit/test_docs.py fails with:

E       UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 26244: ordinal not in range(128)

Just to see, whether this is something specific to my setup, can you reproduce this:

After commenting the first perl replace, can you do a fink rebuild tox-py36; fink reinstall tox-py36; fink -m rebuild tox-py36 ?

Thanks.

nieder:
````
Can't resolve dependency "filelock-py36" for package "tox-py36-3.7.0-1" (no matching
packages/versions found)
````

**Comment by nieder** [comment]
* webob-py* builds fine, but the Homepage field needs to be fixed (not point to toml)
* pv-py* builds fine. To have on the record, enabling tests fails looking for wsgi, which needs PYTHONPATH=%b/build/lib/pv, but then the tests fail with several syntax errors. Those are fixed with this:
````
	perl -pi -e 's|except Exception, e|except Exception as e|g' pv/web.py
	perl -pi -e 's|lambda \(l, s\)|lambda l, s|g' pv/web.py
````
But the package is looking for module `htmlentitydefs`, which is a python builtin module removed from python35 and up. But even for py27/py34 that has htmlentitydefs, no tests to actually run. So seems fine to just leave them commented out. Maybe add the PatchScript above anyway to avoid propagating old syntax?
* filelock-py* builds fine.
* toml-py builds fine, but doesn't actually run tests. Using pytest fails with looking for `toml-test` which seems to be a separate dist of testing specific files. Can deal with that later.

**Comment by nieder** [comment]
For tox-py:
Without any tox-py* installed, tests fail at very beginning with `E   ModuleNotFoundError: No module named 'tox'`. This is fixed by `export PYTHONPATH=%b/build/lib` in TestScript.
Tests then complain about `No such file or directory: '/sw/bin/tox-quickstart-py36`. So it needs to be installed to run its own buildtime tests.
If I install old tox-py36-3.0.0-1, then tests fail with `module 'tox.session' has no attribute 'run_main'`, so it needs to find current tox apparently.
If instead of installing tox, I remove the absolute %p/bin to the tox-* scripts at the end of tox-py.patch (and rely on $PATH) to find the tox scripts in %b/build/bin, the failure is: `The 'tox==3.7.0' distribution was not found and is required by the application`. So looks like some 

If I install tox-3.7.0 w/out tests and then try rebuilding with tests, then the first set of tests pass. I see two later test failures with tests/unit/test_docs.py if I comment out all the ascii related perl-pies. But those are fixed by setting `export LANG=en_US.UTF-8` in TestScript. So that's the fix for the non-ASCII characters in changelog.rst.

There are 4 more test failures in test_z_cmdline.py and 2 in test_package_builder_isolated.py. I'm still looking through to see what can be fixed easily.

tox-3.0.0 also has a bunch of test failures, so it's not like we're regressing.

**Comment by kamischi** [comment]
I have the same 4 fails in test_z_cmdline.py and 2 in test_package_builder_isolated.py. My naive tries to fix them did not get me anywhere.

**Comment by dhomeier** [comment]
It's not just the test call that's broken with `pv-py3x` - just `import pv` from python3.6 and -3.7 fails with the same path error for `wsgi` and, having worked around that, `htmlentitydefs` and `urlparse` import errors. The entire module looks quite horrible to me in terms of documentation, tests, structuring and maintenance, but a kind of brute-force way to fix it for me was setting the patch script instead to
```
PatchScript: <<
  #!/bin/sh -ev
  # Almost nothing in this package conforms with Python 3!
  if [ %type_pkg[python] -ge 34 ]; then
    %p/bin/2to3-%type_raw[python] -w pv/__init__.py pv/escape.py pv/web.py pv/wsgi.py pv/template.py
  fi
<<
```
- `setup.py test` then also works with `PYTHONPATH=%b/build/lib`, but still does nothing.

**Comment by dhomeier** [comment]
I also ran into pytest collections errors in `freezegun` (possibly related to `setuptools-scm` being installed iirc), which I needed to fix by
```
PYTHONPATH=%b/build/lib %p/bin/pytest-%type_raw[python] tests/test*.py || exit 2
```
On the plus side, the `webob-py` tests can also be activated by calling `pytest` directly:
```
InfoTest: <<
  TestDepends: pytest-py%type_pkg[python]
  TestScript: <<
    PYTHONPATH=%b/build/lib %p/bin/pytest-%type_raw[python] tests/test*.py || exit 2
    find build -name "*.pyc" -delete
  <<    
<<
```

**Comment by dhomeier** [comment]
As for the `test_z_cmdline.py` failures, I suspect the whole `envtmpdir` business is not properly setup when running through pytest instead of tox. Other than disabling that test with `-k 'not _script and not test_test_usedevelop'` I don't have a good idea how to get around this.

**Comment by kamischi** [comment]
With "and not test_test_usedevelop" the failing test in the file tests/unit/test_z_cmdline.py is skipped, but the 2 fails in tests/unit/package/builder/test_package_builder_isolated.py are still not skipped. Any hint for that?

**Comment by dhomeier** [comment]
I've hacked `_pytestplugin.py` a bit to track the exception further down and found that apparently this
```
     def test_verbose_isolated_build(initproj, mock_venv, cmd):
        initproj(
            "example123-0.5",
            filedefs={
               "tox.ini": """
                        [tox]
                        isolated_build = true
                        """,
                "pyproject.toml": """
                        [build-system]
                        requires = ["setuptools >= 35.0.2"]
                        build-backend = 'setuptools.build_meta'
                                    """,
            },
        )
        result = cmd("--sdistonly", "-vvv")
```
produces this command call:
```
E         setting PATH=/sw/bin:/Users/derek/bin:/bin:/sw/bin:/usr/bin:/sbin:/usr/sbin:/sw/sbin:/Users/derek/bin:/sw/lib/perl5/ExtUtils:/usr/local/bin:/opt/X11/bin:/usr/X11R6/bin:/bin:.:/opt/N1GE6/bin/darwin-x86
E         ERROR: invocation failed (errno 2), args: /sw/bin/python -c 'import setuptools.build_meta
```
with an unversioned `python` - ditto for `test_package_builder_isolated.py`.
Testing in a clean tox environment is apparently the purpose of these tests, but I am lost as to where the default for `cmd` is set in this case, or how to change it (tried to set `basepython` etc. in the `filedefs`, to no avail).

nieder:
Cleaner pytest call so that you don't have to remove .pyc afterwards:
````
InfoTest: <<
  TestDepends: pytest-py%type_pkg[python]
  TestScript: PYTHONPATH=%b/build/lib %p/bin/python%type_raw[python] -B -m pytest tests || exit 2
<<
````
Will look at the tox updates tonight.

kamischi:
> Cleaner pytest call so that you don't have to remove .pyc afterwards:
> 
> ```
> InfoTest: <<
>   TestDepends: pytest-py%type_pkg[python]
>   TestScript: PYTHONPATH=%b/build/lib %p/bin/python%type_raw[python] -B -m pytest tests || exit 2
> <<
> ```
> Will look at the tox updates tonight.

At least TestDepends also needs freezegun-py. Many more tests fail than with old version.

nieder:
Interesting. I only have freezegun-py27 and -py37, but all tests passed on all py variants (and no LANG set). And building webob-py27 looks identical whether or not I have freezegun installed.

kamischi:
Maybe LANG=en_US.UTF-8 is needed because my system is german.

dhomeier:
Note that the pytest collection errors I reported above for `freezegun-py` (and `webob-py`) were due to `pytest-relaxed-py` being installed on my system - for allowing the `invoke-py` tests to run, but that package screws up virtually every other Python package's test suite by trying to collect eery single Python source file instead of just the `test_*.py` one!
Might have to make this a `TestConflicts` for all of them, but I am wondering if there is another way of keeping pytest-relaxed out unless absolutely needed (i.e. when testing invoke).

kamischi:
Feel free to make according changes. After all, tox-py is Kurt Schwehr's package and not mine.

nieder:
I've cherry picked out the patches to filelock/pv/toml/webob and committed those. I did not touch the tox patches since there have been changes to master (and upstream is now at 3.20.1) and it might be easier to track what's been discussed here in a different PR for tox specifically. I'm going to close this since the other stuff is complete. Please reopen if you think continuing here is needed.

PR #259, opened 2018-09-06
psutil-py: new package

kamischi:
The tests fail, but no idea what to do about it.


kamischi:
Some tests still fail, but fink is no different than a plain build. Bug report is submitted to upstream.

Otherwise the package seems to be operational.

MiSchi.

nieder:
I had to add a py37 variant to @danielj7's funcsigs-py to satisfy deps. But I couldn't find any point in the code that used funcsigs-py anyway and the build logs looked identical with and without it. Might be worth verifying in future updates if funcsigs-py is actually needed.

PR #260, opened 2018-09-07
sync with changes from Mojave and additional fixes

jwhowarth:
The additional changes to the extutils-makemaker-pm.patch port the changes made to the system ExtUtils-MakeMaker-6.66 to cope with the removal of the SDK from / as well as similar changes for the new xs section in 7.34 not present in 6.66. Tested with 'fink -m' on 10.14 without the fall back header package installed.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.info
@@ -1,0 +1,20 @@
+# -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
+Package: extutils-makemaker-pm
+Version: 7.34
+Revision: 1
+Distribution: 10.14
+Depends: system-perl5182, %n5182 (>= %v)
+Type: bundle
+Description: ExtUtils::MakeMaker for /usr/bin/perl
+DescDetail: <<
+	Use BuildDepends:extutils-makemaker-pm for packages that do not
+	otherwise need to be perl-version varianted so that they do
+	not need to be perl-version varianted on account of this build
+	dependency.
+
+	For packages that *are* perl-version varianted, use
+	BuildDepends:extutils-makemaker-pmXXX for the appropriate perlXXX.
+<<
+License: Artistic
+Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
+Homepage: http://search.cpan.org/dist/ExtUtils-MakeMaker/
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm.info
@@ -46,7 +46,7 @@ Source-MD5: 073c0fb4eebf3953de2a1e94fa189bac
 Source-Checksum: SHA256(95f1eb44de480d00b28d031b574ec868f7aeeee199eb5abe5666f6bcbbf68480)
 
 PatchFile: %{ni}.patch
-PatchFile-MD5: 4e690e03f0f654f0922c2261038f1c4e
+PatchFile-MD5: c10cd9bf709b76df975cf23da5c28973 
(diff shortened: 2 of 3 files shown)


jwhowarth:
In theory, the python27 info file changes could be replaced with a patch to setup.py. However, I haven't had any luck finding a permutation which prevents the two python modules from failing to build in the absence of the fall back header SDK package.

danielj7:
With this patch, extutils-makemaker-pm doesn't build for me on 10.13 with Xcode 9.

```
	export BUILDING_AS_PACKAGE=1
	ARCHFLAGS="" /usr/bin/arch -x86_64 perl5.18 Makefile.PL PERL="/usr/bin/arch -x86_64 perl5.18" PREFIX=/sw INSTALLPRIVLIB=/sw/lib/perl5/5.18.2 INSTALLARCHLIB=/sw/lib/perl5/5.18.2/darwin-thread-multi-2level INSTALLSITELIB=/sw/lib/perl5/5.18.2 INSTALLSITEARCH=/sw/lib/perl5/5.18.2/darwin-thread-multi-2level INSTALLMAN1DIR=/sw/share/man/man1 INSTALLMAN3DIR=/sw/share/man/man3 INSTALLSITEMAN1DIR=/sw/share/man/man1 INSTALLSITEMAN3DIR=/sw/share/man/man3 INSTALLBIN=/sw/bin INSTALLSITEBIN=/sw/bin INSTALLSCRIPT=/sw/bin
Checking if your kit is complete...
Looks good
Generating a Unix-style Makefile
Writing Makefile for ExtUtils::MakeMaker
Writing MYMETA.yml and MYMETA.json
/usr/bin/make CC=gcc CXX=g++
make: *** No rule to make target `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Perl/5.18/darwin-thread-multi-2level/CORE/config.h', needed by `Makefile'.  Stop.
```

jwhowarth:
So we may have to conditiionalize these changes to apply only on darwin18 or later.

dmacks:
Do these python or perl changes propagate into the installed files in a way that affects runtime? For example, do packages built by MakeMaker or python27 use the -I flags that they themselves were compiled with when compiling other packages, or are the changes only to compile them themselves? If the changes propagate to their runtime, there needs to be logic control *in the runtime* to Do The Right Thing on all platforms, or else separate two separate .info (with widely separate Revision) so that each platform gets the platform-correct flavor.

tl,dr: probably need to offload 10.14-specific stuff into new .info with new Revision, with Distribution tags.

dmacks:
Regarding python27, I think CFLAGS is not needed. The build system appears to use flags properly, which means compiler calls use that variable along with CPPFLAGS when compiling sources and also use it when linking binaries. That means it's always redundant or unused.

On 10.13, python27 has a bunch of differences in autodetection with using the SDK.../usr flags because it uses system frameworks not just system headers. The normal location /S/L/F are obviously compatible with normal location /usr/include but are seemingly not compatible with SDK.../usr/include. They use different preprocessor tokens, so I assume one would need to pass a flag for SDK.../S/L/F along with SDK.../usr/include.

dmacks:
Okay, there are a ton of weird effects in the flag-passing from the autotools Makefile and then munging in setup.py. I don't see an easy way to get the -F flag passed deeply enough and in the correct order. "Not deep enough" and we get frameworks and regular headers out-of-sync. And setup.py tends to strip out -F or re-order it vs -I, giving incorrect header precedence of system vs fink vs srcdir. Fink's flag-sort can't cope with -F yet either. What worked for me on 10.13 was *just*
```
export CFLAGS="$CFLAGS -isysroot `xcrun --show-sdk-path` -iframework `xcrun --show-sdk-path`/System/Library/Frameworks"
```
That gives all the needed -I and -F flags but tagged as being system so not given precedence regardless of their ordering on the command-line. Could someone with 10.14 please check?

danielj7:
I don't have 10.14 so I'd also be interested in someone testing it.

remkos:
I'll volunteer to install 10.14 on an external drive so that I can always revert. Then check the suggestions by @jwhowarth

PR #271, opened 2018-09-18
pypubsub-py: new package

kamischi:
Version 4 does not build on python 2.7. Therefore, version 3 for that.


nieder:
Building pypubsub4-py3X:
````
running build_ext

----------------------------------------------------------------------
Ran 0 tests in 0.000s

OK
cd tests
/sw/bin/python3.4 perf.py || exit 2
  File "perf.py", line 15, in <module>
    from pubsub import pub
ImportError: No module named 'pubsub'
````
Probably need to set PYTHONPATH to include the freshly built module.

kamischi:
Issue of failing tests fixed. Hint about PYTHONPATH was very helpful.

PR #272, opened 2018-09-18
pyqt4-mac-py: minor update and add py37 variant

kamischi:


nieder:
Build failure:
````
/sw/bin/python2.7-config --ldflags
Querying qmake about your Qt installation...
Determining the details of your Qt installation...
This is the GPL version of PyQt 4.12.3 (licensed under the GNU General Public
License) for Python 2.7.15 on darwin.

Type 'L' to view the license.
Type 'yes' to accept the terms of the license.
Type 'no' to decline the terms of the license.

Do you accept the terms of the license? Found the license file pyqt-gpl.sip.
Error: This version of PyQt4 requires sip 4.19.12 or later.
````
I have sip-py27-4.19.8-2

kamischi:
This will only work after the commit of sip-py and pyqt5-mac-py. I missed the unchanged MD5sum, because i expected an error from the validator, which did not show up, probably because the shasum ist correct.

PR #278, opened 2018-10-01
gevent-py fix; libuv1 update

dmacks:
gevent-pyXX FTBFS for me (same as on bindist) where I have libuv1-1.8.0-1 installed with errors related to unknown API in libuv1, but gevent-pyXX has no dependencies on libuv1. Looks like it autodetects, and if so actually needs a newer version. Pushing libuv1 to its latest version allows gevent-pyXX to build against it, so I set versioned dependencies accordingly.

I did not investigate what version first added the needed API to be able to set a slightly looser dependency (but since fink has no intermediate versions, who cares). I also did not investigate whether there is a way to disable libuv detection.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libuv1.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libuv1.info
@@ -1,6 +1,6 @@
 # -*- coding: ascii; tab-width: 4 -*-
 Package: libuv1
-Version: 1.8.0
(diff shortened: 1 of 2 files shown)


dmacks:
That points to a quite different underlying cause: global -I masking local. Python is quite horrible about that IIRC. Rather than trying to keep in sync, I used flag-sort to make sure the compiler uses the local version.

PR #279, opened 2018-10-02
libxapian update

maartanis:
Update libxapian22 to version 1.2.25; add new libxapian30 with version 1.4.7.; add new python wrappings. libxapian22 is needed for py27, libxapian30 for py34 and higher.

The simplest way to resolve conflicts between libxapian*-doc and libxapian*-bin was to rename the packages version specific and replace/conflict them. I think renaming the docs and bins is not really worth the effort. 

nieder: do you want to take over maintainership of the new libxapian30 packages?


nieder:
lib/pkgconfig/xapian-core.pc was previously incorrectly in libxapian22-shlibs. Because it's been moved to libxapian22-dev (correctly), libxapian22-dev now needs to `Replaces: libxapian22-shlibs (<< 1.2.25-1)`. libxapian30-dev is correct from the start, so it doesn't needs this change.

nieder:
Why do the -bin SplitOffs need to conflict with each other? Are the package contents really different (removed programs, completely changed syntax, etc...)?

nieder:
Do the tests still fail (see `DescPort` and patch) in libxapian30? If not, they should be re-enabled.

10.9-libcxx/stable/main/finkinfo/libs/libxapian22-shlibs.info
```diff
@@ -43,29 +43,36 @@ Splitoff: <<
 	Description: Search engine library, dev pkg
 	Depends: %N (= %v-%r)
 	BuildDependsOnly: true
+    Conflicts: libxapian22-dev, libxapian30-dev
```
nieder: Whitespace fix (standardize on tabs) needed here and in all C/R fields.

kamischi:
summary of last commits:
- fixed indentation with tabs
- replace of libxapian22-shlibs with libxapian22-dev added
- patch of libxapian30-shlibs removed as all tests succeed.
- separate libxapian*-bin packages kept, since the libxapian30-bin package has some binaries removed.

nieder:
The xapian bindings use c++ and need a `GCC: 4.0` field (at least py27, haven't been able to build py3X).
Also, xapian30 fails tests:
````
PASS: stemtest
Running test: closefrom1...readdir failed: Bad file descriptor                 
FAIL: unittest
========================================
1 of 4 tests failed
Please report to https://xapian.org/bugs
========================================
````

dmacks:
An alternate solution to the -doc package-naming and Conflicts/Replaces is to merge that package into whichever package they document. Looks like they are all about the API, so that would be -dev. So whenever you have a -dev installed, you automatically have the correctly-versioned -doc for it rather than possibly having xapian22-dev but xapian30-doc.

kamischi:
dmacks: clever suggestion. I hope my implementation is correct.

nieder: 
- GCC field added. 
- I can't reproduce the test failure of libxapian30-***. The only difference might be, that i have some packages updated locally. Can this be?
- Can you give details of the fail when building xapian-py3*? Here, the build of 34, 35, 36 goes well, although the tests have numerous warnings, but no failures.

kamischi:
nieder: In order to have a clean as possible setting I purged python34 and all dependencies and rebuild xapian-py34 without any updated python mods. It was built without any problem. Which did python version did you try? I could do the same trick with 35. 36 would be hard, but if need be ... 

After updating sphinx to 1.7.7 and adding the py37 variant, i could also build xapian-py37.

nieder:
It's the libxapian30-shlibs package that fails the test, not the corresponding pymod. I can't get to the pymod until the library builds successfully.

kamischi:
nieder: OK. Finally, got the problem. I rebuilt libxapian30-shlibs with `fink -mkK --build-as-nobody rebuild libxapian30-shlibs`. All tests, including unittest, passed. My only idea is to compare complete logs. I attached my log.
[libxapian30-shlibs.log](https://github.com/fink/fink-distributions/files/2453853/libxapian30-shlibs.log)

nieder:
How many build jobs are you using in Fink? For me it fails reproducibly with -j4 in TestScript, but not with -j1.

kamischi:
My 'fink configure' reports 4. I also tried 'make -j4 check' in the TestScript. It still ran through. Since I see no big harm in using 'make -j1 check', I did the change and left a note in DescPort.

nieder:
The test failure is due to BSD make not being threadsafe for that test, while GNU make works all the time.  Passing -j1 is the best solution.

PR #280, opened 2018-10-04
pipdeptree-py: new package

kamischi:
pipdeptree is a utility to show dependency trees of packages.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pipdeptree-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pipdeptree-py.info
@@ -1,0 +1,52 @@
+Info2: <<
+
+Package: pipdeptree-py%type_pkg[python]
+Version: 0.13.0
+Revision: 1
+License: OSI-Approved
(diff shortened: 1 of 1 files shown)


nieder:
The wording on DescPort is very unclear. I had to find the README.rst to find out you meant graphviz was needed for pipdeptree to output as an image. If graphviz adds functionality, should be added to the `Suggests` or `Recommends` field, or even `RuntimeDepends` if you feel it's a critical functionality.

However, it fails for me even when graphviz is installed:
````
 i   graphviz                   2.40.1-2             Graph visualization software (binaries)
$ /sw/bin/pipdeptree2.7 --graph-output png > dependencies.png
graphviz is not available, but necessary for the output option. Please install it.
````

nieder:
I just committed graphviz-py.info

kamischi:
Thanks. Works.

nieder:
Almost forgot: need Depends on `python%type_pkg[python]` as well. The other pymods will bring it in but is customary to make sure you have the interpreter that will actually be running the pymod. Good otherwise.

kamischi:
I wasn't aware of that. So, I'll change my proxies to this. Thanks for your feedback.

PR #281, opened 2018-10-04
pip-py: Update to newer upstream version (18.0)

kamischi:
Source URL updated as well.


kamischi:
newer version is already committed.

PR #282, opened 2018-10-05
pytz-py37: tests fixed

kamischi:
I adjusted the expected values for the tests for py37.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytz-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pytz-py.info
@@ -2,7 +2,7 @@ Info2: <<
 Package: pytz-py%type_pkg[python]
 Version: 2018.5
 Revision: 1
-Type: python (2.7 3.4 3.5 3.6)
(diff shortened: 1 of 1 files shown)


dhomeier:
I had a go at the PatchScript, but so far was only able to push it to my own copy of the branch in f9b715f - feel free to fetch it from there, which is probably safer than me trying to force-push something here (if I even have the access rights).

I removed the backup flag from the last two `sed` calls so the original version would be preserved in README.txt.bak.

dmacks:
Sorry, I have no idea how to pull a local copy of a remote alternate working tree. The kamischi-pytz-py-update branch is part of the main fink-distributions repo, so I think you have write access to it. But otherwise, could you make your commit into the kamischi-pytz-py-update branch of your dhomeier/fink-distributions fork (cherry-pick it all into a new clean branch), and then make a new PR for that branch, so you get attributed for your fix?

nieder:
@dmacks ping. pytz-py37 is needed to fill in some dep holes through matplotlib. thanks

dmacks:
My fink sandbox machine is almost back in service, but "looks fine by eye"

PR #283, opened 2018-10-11
ghostscript-9.25-1

remkos:
This is the update for ghostscript from 9.24 to 9.25.
Built and test on 10.14 with "fink -m build"


nieder:
v9.25 fixes a problem building doxygen-doc from the latest upstream. With 9.24 present, the docs build process uses, epstopdf, which then calls gs, but it can't find the GS icc profiles. With 9.25, building the doxygen docs successfully finishes.

remkos:
@schwehr: Duly noted. But I don't like it when `patch` reports it needed to be sloppy. Better safe than sorry.

PR #285, opened 2018-10-11
grib-api-1.27.0-1 (and grib-api-fortran-1.27.0-1)

remkos:
This is a simple upstream update from 1.26.1 to 1.27.0.
(I missed that update on grib-api-fortran though so that jumps one version).
Built and tested with "fink -m build"

Note 1:
This does not need the complicated install_name_tool stuff as in eccodes (see PR #231)
This is because GRIB-API is not built with cmake.

Note 2:
I removed the old packages first to be sure the test was done with the just built executables (before installing in %p)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/grib-api-fortran.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/grib-api-fortran.info
@@ -15,17 +15,15 @@ PatchFile-MD5: 0448a7a625b66581baab7cc3e5834e71
 
 SourceDirectory: grib_api-%v-Source
 BuildDependsOnly: true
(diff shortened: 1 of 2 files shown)


remkos:
Thanks for catching that @danielj7. I'll research. I have both installed, so it did not bite me.

remkos:
`libopenjp2.7` was actually not supported by GRIB-API in the first place, only up to version 1.5.2, as in `libopenjpeg1`. However, in order to avoid this older version and the additional flags needed to find the rather buried header and libraries, I selected what GRIB-API considers their JPEG library of choice: Jasper. So we now use `libjasper.1` instead.

PR #289, opened 2018-10-13
wxpython400-py: new package.

kamischi:
updated wx library and python wrapper.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/wxpython400-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/wxpython400-py.info
@@ -1,0 +1,186 @@
+Info2: <<
+
+Package: wxpython400-py%type_pkg[python]
+Version: 4.0.3
+Revision: 1
+Type: python (2.7 3.4 3.5 3.6 3.7)
+License: GPL
+
+BuildDepends: <<
+  doxygen (>= 1.8.8),
+  fink-package-precedence,
+  expat1,
+  libiconv-dev,
+  libjpeg9,
+  libpng16,
+  libtiff5,
+  setuptools-tng-py%type_pkg[python]
+<<
+
+Depends: <<
+  %N-shlibs
+<<
+
+Source: https://files.pythonhosted.org/packages/source/w/wxPython/wxPython-%v.tar.gz
+Source-MD5: fc5e87be879f0855740c5d15b7586be5
+
+GCC: 4.0
+
+PatchScript: <<
+#!/bin/sh -ev
+  sed -i.bak 's|10.6|10.9|g' buildtools/build_wxwidgets.py
+  sed -i.bak 's|if not self.IsDisplayAvailable():|if False :|g' wx/core.py
+<<
+
+SetCFLAGS: -MD
+
+CompileScript: <<
+#!/bin/sh -ev
+  export DOXYGEN=%p/bin/doxygen
+  %p/bin/python%type_raw[python] build.py dox etg --nodoc sip build
+  fink-package-precedence --depfile-ext='build.*\.d' .
+<<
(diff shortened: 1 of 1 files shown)


nieder:
Each dylib exists with these versioned names:
````
3.0.0.4.0.dylib
3.0.0.dylib
3.0.dylib
````
The install_name is `FOO_3.0.0.4.0.dylib` (and that's the actual file). The fully versioned dylib ends up in the -shlibs SplitOff and the 2 lesser versioned files are in the base package.
Are other packages supposed to be able to link to the dylibs, or are they for internal use only? If there's no linkage expected to happen, the we don't need the lesser versioned dylib symlinks. If linkers are supposed to use them, then those should be moved to a -dev SplitOff. Are there no public headers for other programs to compile against wxpython400?
The `install_name -id` chunk in InstallScript is just to point to the correct final `%p/lib/...` of the library and not to fix the file name, correct?

Why are you setting `PYTHONPATH=/sw/lib/wxpython400-py27` to some .py files during InstallScript? I don't see that path anywhere in either of the two current SplitOffs. Add a comment to this perl-pie why this is needed.

Should the files in %p/bin be moved to their own -bin SplitOff? Or are they intrinsically tied to the site-packages/wx/*.py that they should always be kept together?

The contents of `%p/share/doc/wxpython400-py27/demo/data/locale` are not readable.

kamischi:
> Are other packages supposed to be able to link to the dylibs, or are they for internal use only? If there's no linkage expected to happen, the we don't need the lesser versioned dylib symlinks. If linkers are supposed to use them, then those should be moved to a -dev SplitOff. Are there no public headers for other programs to compile against wxpython400?

Dylib symlinks moved. No header files.

> The `install_name -id` chunk in InstallScript is just to point to the correct final `%p/lib/...` of the library and not to fix the file name, correct?

Correct.
 
> Why are you setting `PYTHONPATH=/sw/lib/wxpython400-py27` to some .py files during InstallScript? I don't see that path anywhere in either of the two current SplitOffs. Add a comment to this perl-pie why this is needed.

Removed

> Should the files in %p/bin be moved to their own -bin SplitOff? Or are they intrinsically tied to the site-packages/wx/*.py that they should always be kept together?

I thought, that with -dev and -shlibs splitoffs, the bins can stay in the base package. No?

> The contents of `%p/share/doc/wxpython400-py27/demo/data/locale` are not readable.

Can you check again?

nieder:
%p/bin can stay with the site-packages directory in the base package, especially if they're intrinsically linked.

Last thing would be to document both lines in PatchScript, but is good now I think.

kamischi:
I could not find out the details for the second patch in a brief test, although I remember hunting down the issue.

nieder:
One thing I noticed: only some of the pyXX variant runs tests. For me, just py27 and py34 ran tests. The other variants just have the output below during TestScript. Could this be a missing TestDepends?
````
/sw/bin/python3.6 setup.py test || exit 2
running test
running egg_info
writing wxPython.egg-info/PKG-INFO
writing dependency_links to wxPython.egg-info/dependency_links.txt
writing entry points to wxPython.egg-info/entry_points.txt
writing requirements to wxPython.egg-info/requires.txt
writing top-level names to wxPython.egg-info/top_level.txt
reading manifest file 'wxPython.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
no previously-included directories found matching 'docs/sphinx/build'
warning: no previously-included files matching '.git' found under directory 'wx'
warning: no previously-included files matching '*.pyc' found under directory 'wx'
warning: no previously-included files matching '__pycache__' found under directory 'wx'
writing manifest file 'wxPython.egg-info/SOURCES.txt'
running build_ext

USING VERSION: 4.0.3

phase test: passed

````

PR #292, opened 2018-10-18
New package skhd

bcully:
This is handy on its own but I've packaged it to support keyboard control of chunkwm, coming shortly.


nieder:
* Please add LICENSE.txt to DocFiles.
* Should the function of examples/com.koekeishiya.skhd.plist be explained (like you do for skhdrc)? Or is it self-explanatory to anyone using this?

PR #293, opened 2018-10-18
New package chunkwm 0.4.6

bcully:
The closest you can get to i3 on a mac, it seems.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/chunkwm.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/chunkwm.info
@@ -1,0 +1,58 @@
+Package: chunkwm
+Version: 0.4.6
+Revision: 1
+Source: https://codeload.github.com/koekeishiya/%n/tar.gz/v%v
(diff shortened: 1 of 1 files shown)


10.9-libcxx/stable/main/finkinfo/utils/chunkwm.info
```diff
@@ -0,0 +1,67 @@
...
+
+Description: Tiling window manager
+DescDetail: <<
+chunkwm is a tiling window manager for macOS that uses a plugin
+architecture, successor to kwm. It represents windows as the leaves of
+a binary tree, and supports binary space partitioned, monocle and
+floating desktops.
+<<
+DescUsage: <<
+Copy chunkwmrc from %p/shard/doc/%n/examples into ~/.chunkwmrc to get
```
bcully: typo: shard -> share. I'll fix this.

nieder:
This is an upstream bug, but the plugins are being built as `-shared`, which makes them dylibs, not plugins/bundles. If they fix that (build as .dylib, or change to `-bundle`, then you can get rid of the Shlibs: field and not have to fix the install_name.

PR #295, opened 2018-10-20
doxygen 1.8.14

nieder:


dmacks:
By eye in doxygen-doc, the build now uses cmake, but is missing BuildDepends:cmake. Also it has -MD but no fink-package-precedence call (possible that it doesn't actually *compile* andthing for a pure -doc package, just seems inconsistent).

nieder:
-MD and no f-p-p was not really an issue because even though it previously compiled stuff, we didn't care what got compiled since the -doc package only installed documentation. But now it doesn't compile anything, so it's been removed.

kamischi:
I downloaded the 3 files and did a maintainer rebuild and reinstall. All looks fine to me.
MiSchi.

Issue #300, opened 2018-10-28
datrie-pyXX mixes external and internal libdatrie code


dmacks:
Instrumenting the build with ```SetCPPFLAGS:-MD```, I see that both the internal libdatrie (C library) and the python glue code (C code wrapping the C library) are using fink's libdatrie1 headers rather than the ones supplied as part of the internal libdatrie library (but still using the internal *.c). Debian's package has a patch to use the external lib. I filed an upstream datrie-py bug about it (https://github.com/pytries/datrie/issues/56).

The reason for the header misuse is that setuputils appends flags rather than prepending, so the default CPPFLAGS (-I%p/include) take priority over anything setup.py adds. One could also argue that this is a long-standing bug in setuputils.

danielj7:
Gah, you're right. Stupid python. I tried the debian patches and they do allow it to build using libdatrie1, but there are 6 test failures, while all tests pass with the internal libdatrie. I'll have to see if its version is different from our package.

danielj7:
Ok https://github.com/fink/fink-distributions/pull/301 seems to fix it.

danielj7:
https://github.com/fink/fink-distributions/pull/301 now merged to fix this.

PR #301, opened 2018-10-28
Update libdatrie1 and fix datrie-py to use it

danielj7:
Fix for https://github.com/fink/fink-distributions/issues/300


danielj7:
I fixed the HypothesisDeprecationWarning and removed the fink BDep. I'm working on a PR for datrie for other things so I'll see about doing one for the warning. Thanks.

PR #302, opened 2018-10-30
bison 3.2

nieder:
sets several ac_cv_prog macros to be deterministic. GNU-gawk was needed at some point in the past, but it's not needed any more (at least as of 10.11).

Tested on 10.11.


nieder:
Those 2 skips here as well and all others passed.

Issue #305, opened 2018-11-03
pyqt5-mac-py27-5.11.2-2 fails to build


nieder:
````
#! /bin/sh -ev
export PATH=/sw/lib/qt5-mac/bin:$PATH
export QTDIR=/sw/lib/qt5-mac

/sw/bin/python2.7 configure.py \
	--confirm-license \
	-d /sw/lib/python2.7/site-packages \
    -b /sw/lib/qt5-mac/lib/python2.7/bin \
    -q /sw/lib/qt5-mac/bin/qmake \
    --designer-plugindir /sw/lib/qt5-mac/plugins/python2.7/designer \
    --qml-plugindir /sw/lib/qt5-mac/plugins/python2.7/PyQt5 \
    -v /sw/share/sip-py27/pyqt5-mac \
    -n /sw/lib/qt5-mac/include/python2.7 \
    --no-python-dbus \
    --no-dist-info \
    INCDIR_OPENGL+=/usr/X11R6/include \
    LFLAGS_PLUGIN="-bundle `/sw/bin/python2.7-config --ldflags`" \
    LIBS="`/sw/bin/python2.7-config --ldflags`"
/sw/bin/python2.7-config --ldflags
/sw/bin/python2.7-config --ldflags
Querying qmake about your Qt installation...
Error: Unable to import PyQt5.sip. Make sure you have configured SIP to create
a private copy of the sip module.
### execution of /tmp/fink.dHnBa failed, exit code 1
````

py34/35/36 all build with this output during the configure phase
````
#! /bin/sh -ev
export PATH=/sw/lib/qt5-mac/bin:$PATH
export QTDIR=/sw/lib/qt5-mac

/sw/bin/python3.5 configure.py \
	--confirm-license \
	-d /sw/lib/python3.5/site-packages \
    -b /sw/lib/qt5-mac/lib/python3.5/bin \
    -q /sw/lib/qt5-mac/bin/qmake \
    --designer-plugindir /sw/lib/qt5-mac/plugins/python3.5/designer \
    --qml-plugindir /sw/lib/qt5-mac/plugins/python3.5/PyQt5 \
    -v /sw/share/sip-py35/pyqt5-mac \
    -n /sw/lib/qt5-mac/include/python3.5 \
    --no-python-dbus \
    --no-dist-info \
    INCDIR_OPENGL+=/usr/X11R6/include \
    LFLAGS_PLUGIN="-bundle `/sw/bin/python3.5-config --ldflags`" \
[...]

PR #308, opened 2018-11-12
wxmplot-py: new package

kamischi:
Plotting widgets for wxPython with matplotlib. No py37 variant, since matplotlib-py37 is still missing.


nieder:
Waiting on #305, which is needed for matplotlib-py.

kamischi:
since [#305](https://github.com/fink/fink-distributions/issues/305) is resolved, this should be ok.

kamischi:
nieder: no problem. go ahead.

PR #316, opened 2018-12-12
clamav 0.101.0

mommsen:
Update clamav to the latest upstream version. Increment versioned split-offs due to new libraries.


nieder:
This breaks Shlibs policy because it removes the package clamav8-shlibs (though nothing else uses clamav8-shlibs).

Ping @dmacks about whether OK or not to lose clamav8-shlibs package.

mommsen:
I know that it breaks the strict shlibs policy, but it has been handled like this since years to avoid keeping around many out-dated clamav info files.

remkos:
If nothing depends on the shared libraries, then make them "private", i.e. by starting the Shlibs line with a "!", put that in the clamav package and eliminate the clamav9-shlibs split off.

PR #318, opened 2018-12-14
libcurl4: fix include prefix for 10.14

kamischi:
The files for the copy in the patch script are no longer in /usr/include.


danielj7:
This looks fine. I don’t have 10.14 so I’ll trust that it works. 🙂

PR #320, opened 2018-12-18
doxygen: fix cmakefile for Mojave

PovlAbrahamsen:
doxygen refuses to build on 10.14, since it specifies "-mmacosx-version-min=10.5" as a compiler option. This patch changes the minimum version to 10.9, in line with changes to doxygen's source code (though they haven't published a new version yet). Tested on 10.14.2.


kamischi:
Thank you sooo much. This was a major road block for a number of packages I use.

PR #333, opened 2019-01-24
gmt5-5.4.5-1: Upstream update

remkos:
This is an update to the gmt5 package because of an upstream update from version 5.4.4 to 5.4.5. No further changes needed.
Tested on 10.14.2 with Command Line Tools.
fink -m build gmt5 successful.


kamischi:
How did you get gcc7 for fftw3 on 10.14?

PR #334, opened 2019-01-27
aquaterm: fix building on 10.9+

kamischi:
Fix as learned from macPorts. Also cleaned up outdated arch treatment.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/aquaterm.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/aquaterm.info
@@ -1,6 +1,6 @@
 Package: aquaterm
 Version: 1.1.1
-Revision: 4
+Revision: 5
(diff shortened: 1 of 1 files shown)


nieder:
On a VM with 10.13/Xcode10, the above change does allow me to build aquaterm because that system does have `/Library/Developer/PrivateFrameworks`. So the `-UseNewBuildSystem=NO` change probably needs to be conditionalized on Xcode version.

nieder:
ahh. Based on https://stackoverflow.com/questions/46463791/xcode-9-missing-coresimulator-framework I have to run Xcode.app once to have it install some modules. Once I did that, /Library/Developer/PrivateFrameworks was populated, and aquaterm built against Xcode9 with the current .info file w/ or w/out @kamischi's patch.

nieder:
Should be fixed with 3614892. I had to conditional the flag, but is otherwise the patch proposed here. No rev up because the conditional change only applies to non-building systems.

PR #339, opened 2019-02-02
On Xcode10, the existing hardwired MACOSX_DEPLOYMENT_TARGET=10.3 sett…

nieder:
…ing breaks the build.

Use upstream's patch to not set M_D_T for 10.6+ and instead use the correct value of macosx-version-min to match the system.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/perl5162-deployment_target.patch
+++ b/10.9-libcxx/stable/main/finkinfo/languages/perl5162-deployment_target.patch
@@ -1,0 +1,155 @@
+--- a/hints/darwin.sh
++++ b/hints/darwin.sh
+@@ -143,7 +143,7 @@ esac
+ 
+ # Shared library extension is .dylib.
+ # Bundle extension is .bundle.
+-ld='cc';
++
+ so='dylib';
+ dlext='bundle';
+ usedl='define';
+@@ -167,28 +167,137 @@ case "$ccdlflags" in		# If passed in from command line, presume user knows best
+ ;;
+ esac
+ 
++# Allow the user to override ld, but modify it as necessary below
++case "$ld" in
++    '') case "$cc" in
++        # If the cc is explicitly something else than cc (or empty),
++        # set the ld to be that explicitly something else.  Conversely,
++        # if the cc is 'cc' (or empty), set the ld to be 'cc'.
++        cc|'') ld='cc';;
++        *) ld="$cc" ;;
++        esac
++        ;;
++esac
++
++# From http://ftp.netbsd.org/pub/pkgsrc/current/pkgsrc/mk/platform/Darwin.mk
++#
++# OS, Kernel, Xcode Version
++# Note that Xcode gets updates on older systems sometimes.
++# pkgsrc generally expects that the most up-to-date xcode available for
++# an OS version is installed
++#
++# Codename        OS      Kernel  Xcode
++# Cheetah         10.0.x  1.3.1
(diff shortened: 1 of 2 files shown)


dmacks:
This is obviously the "right" solution for fixing perl itself. But IIRC the result is hardcoded into the recipe that the resulting perl uses when perlmods are built. That means this change would entail rev-up and a BDep on this perl for all binary -pm5162 packages. That's *really* hard to do because perl5162-core is virtual ("Provides" from system-perl) on on 10.9--no version string, so the BDep:perl>= would have to be only for 10.10+ and there is no %token for distro to use for a conditional dependency.

How about killing perl5162 on 10.14 (or even further back is fine too:) and having a private patched copy buried somewhere (like for a few other "-legacy" packages) for special cases rather than a normal public package? Or else have texi2html (the only actual consumer of this specific perl) contain its own private copy of perl (again, with the patch)?

nieder:
This is a compiler and linker command when building encode-pm5162 on 10.13/Xcode9 using our current perl5162:
````
gcc -c -I./Encode  -I../Encode -I/sw/include -fno-common -DPERL_DARWIN -no-cpp-precomp -fno-strict-aliasing -pipe -fstack-protector -O3   -DVERSION=\"2.03\" -DXS_VERSION=\"2.03\"  "-I/sw/lib/perl5-core/5.16.2/darwin-thread-multi-2level/CORE"   ma_03_t.c
env MACOSX_DEPLOYMENT_TARGET=10.3 cc  -L/sw/lib -bundle -undefined dynamic_lookup -fstack-protector  Symbol.o symbol_t.o  -o ../blib/arch/auto/Encode/Symbol/Symbol.bundle  \
````
This is the compiler command when building with the patched perl5162:
````
gcc -c -I./Encode  -I../Encode -I/sw/include -fno-common -DPERL_DARWIN -no-cpp-precomp -mmacosx-version-min=10.13 -fno-strict-aliasing -pipe -fstack-protector -O3   -DVERSION=\"2.03\" -DXS_VERSION=\"2.03\"  "-I/sw/lib/perl5-core/5.16.2/darwin-thread-multi-2level/CORE"   ma_03_t.c
gcc  -L/sw/lib -mmacosx-version-min=10.13 -bundle -undefined dynamic_lookup -fstack-protector  Symbol.o symbol_t.o  -o ../blib/arch/auto/Encode/Symbol/Symbol.bundle 
````

So flags are being changed when building with old vs new perl5162.

Instead of creating a private or legacy perl5162: apply the patch only for macos >= 10.14. Currently, perl5162 doesn't build on 10.14, so there's no %r to bump. And perlmods built on earlier macOS (and using hardcoded M_D_T=10.3) should still run after the system is upgraded to 10.14 irrespective of how the perl5162 was built on the 10.14 machine, right?

akhansen:
I guess the question isn't whether perlmods built without the patch on < 10.13 and brought over to 10.14 will _work_, but rather whether the patch results in different binaries?

nieder:
A hidden or private perl5162 is out of my range.

dmacks:
PR #347 is a "private perl5162" implementation. Not really "onboard" (was harder to have it in the actual texi2html package and .info itself). But it uses the upstream patches for perl and avoids the propagated binary change (we could drop the whole actual perl5162/*-pm world on 10.14; see issue #6) and could even push this back further to 10.13 if we wanted. The M_D_T issue is still a bug on that older platform, just no internal linking error as a result of it.

nieder:
Taken care of by #347

PR #341, opened 2019-02-04
fftw3-3.3.8.1: upstream update and make things work on Mojave

remkos:
This update is three-fold:
- Make things work on Mojave: this included adding `xcrun` to get the SDK path for `/usr/include`
- Removed a switch for the support of powerpc, which does not apply to the `10.9-libcxx` tree, as powerpc is no longer supported.
- Update to the latest FFTW3 version (3.3.8 from May 2018)

Run and tested on 10.14.2 with Command Line Tools 10.1 and `fink -m build`

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/fftw3.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/fftw3.info
@@ -1,13 +1,14 @@
 Info2: <<
 Package: fftw3
-Version: 3.3.7
-Revision: 2
+Version: 3.3.8
+Revision: 1
 Maintainer: Remko Scharroo <remkos@users.sourceforge.net>
 
 Type: gcc (7)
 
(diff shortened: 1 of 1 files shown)


10.9-libcxx/stable/main/finkinfo/sci/fftw3.info
```diff
@@ -28,11 +29,7 @@ CompileScript: <<
  #!/bin/sh -ev
  gcclib=`%p/bin/gfortran-fsf-%type_raw[gcc] --print-lib`
  FLIBDIR=`dirname $gcclib`
- if [ x%m == x"powerpc" ]; then
```
schwehr: Hurrah for removing powerpc legacy cruft!

remkos: 👍 You're welcome!

PR #344, opened 2019-02-05
sphinx-py-1.8.4-1: Add missing dependency on typing-py and upstream update

remkos:
This PR is twofold:
- Add the dependency on typing-py (for python < v3.5). It was added to the dependencies for the test suite, but not for the package itself.
- En passant, updated to upstream version 1.8.4, released 3 Feb 2019.


remkos:
I forgot to mention:
- Compiled and tested with `fink -m build` and Python 2.7 on Mac OS 10.14.2 with Command Line Tools 10.1

Issue #19, opened 2017-12-22
Some small things for wget


dmacks:
1. There are some obsolete items in the user-facing packaging notes of wget-1.19.2-2:
- DescDetail talks about upstream versions 1.13.x, and especially mentions use of fink's openssl vs apple's because apple's is missing a symbol (vs now being missing altogether)--see also DescPort.
- DescUsage talks about certificate support on OS X 10.4 and 10.5, and nothing else. Are these instructions still valid on 10.9 and beyond?

2. I filed an upstream bug-report about the format-string warning in src/utils.c line 2822. Not critical to fix (just in a diagnostic output). The message is not appearing in their list archives yet:
```
-            logprintf (LOG_VERBOSE, _ ("Skipping key with wrong size (%d/%d): %s\n"),                  
+            logprintf (LOG_VERBOSE, _ ("Skipping key with wrong size (%lu/%d): %s\n"),                  
                        (strlen (begin_pos + 8) * 3) / 4, SHA256_DIGEST_SIZE,                            
                        quote (begin_pos + 8));                                                          
```

Update: see [bug-wget message](http://lists.gnu.org/archive/html/bug-wget/2017-12/msg00022.html)

3. It currently uses openssl100. You can probably hold off testing with openssl110...upstream says at best this release only builds when openssl110 is built with certain options. But if you want to, they do have a [recent git commit to support it](https://gitlab.com/gnuwget/wget/commit/6aa6b669efbef62f60471c18e7fdef0206f92337)  (or could just wait for next release:)

akhansen:
I think the certificate option may still hold true.

akhansen:
Updated in commit  1704410

remkos:
As far as I know, the following issues with wget were covered:

1. Regarding openssl versus Apple's: @akhansen already removed the reference to Apple before I took over as maintainer. See commit 1704410. Now two SSL options are available: openssl and gnutls.
Regarding the certificates: That too @akhansen already modified, removing the reference to 10.4 and 10.5 by commit 1704410.

2. This bug was dealt with in a different way upstream: adding `(int)` in front of `(strlen`.

3. The issue with openssl110 was also tackled upstream by a [commit to support it](https://gitlab.com/gnuwget/wget/commit/6aa6b669efbef62f60471c18e7fdef0206f92337) on 26 Nov 2017. It has been working since wget 1.19.5, at least.

Thus, I think, all matters are closed.

dmacks:
[...]

PR #345, opened 2019-02-06
wget-1.20.1-2: Avoid linking libpcre2 (use libpcre1 always)

remkos:
This is an update since a user found that the binary package was linked against libpcre2 rather than the intended libpcre1.
This is now avoided by adding `--enable-pcre --disable-pcre2` to `configure`.

Built and tested on Mac OS 10.14.2 with Command Line Tools 10.1, with both libpcre1 and libpcre2 installed. Only libpcre1 was now linked as required.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/web/wget.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/wget.info
@@ -1,6 +1,5 @@
 Info2: <<
 Package:  wget%type_pkg[-gnutls]%type_pkg[-idn]
-# OPENSSL110 FTBFS; on hold pending issue #19
 Version: 1.20.1
 Revision: 2
 Type: -gnutls (boolean), -idn (boolean)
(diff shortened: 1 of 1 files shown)


dmacks:
Please put a DescPackaging note about this pcre linkage issue (that it would try to link both and you chose not pcre2 because...pcre1 is good enough and more widely used? Or whatever reason).

Unrelated to this issue, but a buglet from 446b6909054ec44f46c384ac4955efa6e1db2e21#diff-4d5cc0e353ef4218661591c31beb3fd6: in that commit, you moved up to openssl110 (thanks!). So please remove line 3 of the .info:
````
# OPENSSL110 FTBFS; on hold pending issue #19 
````
I think there might still be items from issue #19 left unresolved.

remkos:
@dmacks: Thanks for pointing that out. Both the update of the DeskPackaging note and the removal of the buglet are done in commit 10657c5.
Feel free to continue the review.

PR #347, opened 2019-02-09
Texi2html onboard perl

dmacks:
Creating a private perl5162 interpretter (hidden in lib/texi2html) for texi2html to use on platforms where apple doesn't supply perl5162 and we don't feel like having to continue supporting a full user-land perl5162 suite. See PR #339.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/perl5162.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/perl5162.info
@@ -1,6 +1,7 @@
 Package: perl5162
 Version: 5.16.2
 Revision: 103
+# sync changes from here into texi2html-perl5162 pkg if appropriate
 Distribution: 10.10, 10.11, 10.12, 10.13, 10.14
 Depends: %n-core (>= %v-%r)
 Conflicts: perl5123, perl5124, perl5162, perl5182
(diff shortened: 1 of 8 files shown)


dmacks:
Note, this isn't actually *tested* on 10.14 as I only have a 10.13 fink sandbox at the moment...

kamischi:
with an unmodified 10.14, building breaks with:
...
ld: library not found for -lgcc_s.10.4

nieder:
Is it texi2html that fails building or the private perl (texi2html-perl5162)?

kamischi:
private perl.

dmacks:
grr. Depending on weather and everyone's health in my house, I might be able to get to a 10.14 machine starting tomorrow to work on it for real.

nieder:
@kamischi can you look in the build output for `What command should be used to create dynamic libraries?` and paste the output from that query and the next two?

nieder:
On my 10.14 VM, texi2html-perl5162 built fine/passed tests. And the linker is passed `-mmacosx-version-min=10.14` as expected with the patch. texi2html-1.82-203 also built, and perl5162-core is not installed.

nieder:
Another positive report for texi2html-perl5162 on Fink-beginners: https://sourceforge.net/p/fink/mailman/message/36590638/

dmacks:
Given a few positives and no details about the fail, and that it (at least sometimes) fixes a hard blocker for many 10.14 users, should we push this as (at worst) a stop-gap solution pending further improvement?

nieder:
Concur.

nieder:
I guess we now need to mark all 5162 perlmods as !10.14

dmacks:
Won't see me crying...

PR #348, opened 2019-02-11
sbcl v1.4.16

leighsmith:
- Builds sbcl (Steel Bank Common Lisp) on MacOS v10.14.3.
- Reduces warnings building python v3.7 with Xcode compiler.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/python37.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python37.info
@@ -52,7 +52,7 @@ Source: https://www.python.org/ftp/python/%v/Python-%v.tar.xz
 Source-Checksum: SHA256(d83fe8ce51b1bb48bbcf0550fd265b9a75cdfdfa93f916f9e700aef8444bf1bb)
 
 PatchFile: %n.patch
-PatchFile-MD5: a68f52a0d892c9b3d2628289c9fae0a2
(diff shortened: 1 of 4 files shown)


nieder:
Pinging danielj7 and @jessealama

jessealama:
This looks good. The texinfo changes should probably have been added long ago...

danielj7:
Oh, wait. Actually, that -w flag should be conditionalized to only be added on 10.13 and later. It’ll throw warnings on older Xcodes.

leighsmith:
Hmm, the extra warning occurs within the following conditional context:
```
	# for OS X 10.14, ensure full SDK is available, but include it after any fink specific versions.
	if [ "$(uname -r | cut -d. -f1)" -ge 18 ]; then
		SDK_PATH="$(xcrun --sdk macosx --show-sdk-path)"
		export CFLAGS="-Wno-nullability-completeness -Wno-expansion-to-defined -isysroot $SDK_PATH"
		export CPPFLAGS="-I%p/include -I%p/include/ncursesw"
		perl -pi -e "s|/usr/include|$SDK_PATH/usr/include|" setup.py
	fi
```
So that would seem to meet the request to ensure this flag is only applied to MacOS 10.14.

danielj7:
Ah, indeed. I missed that. Looks good.

akhansen:
You guys broke maxima.

PR #350, opened 2019-02-11
gdl-0.9.7-3: Make configuration work for Mojave (find zlib)

remkos:
This is a simple update that helps to find zlib on Mojave using `xcrun`.

Built and tested on 10.14.3 with Command Line Tools 10.1 and `fink -m build`.

PS: I was not able to get either upstream 0.9.8 or 0.9.9 compiled due to an incompatibility with `plplot` in `src/gdlgstream.hpp`. It is true that Fink's `plplot-5.9.6-5` is very much out of date: upstream is now at 5.14.0.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/gdl.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/gdl.info
@@ -1,6 +1,6 @@
 Package: gdl
 Version: 0.9.7
-Revision: 2
+Revision: 3
 Epoch: 1
 Description: GNU Data Language
 License: GPL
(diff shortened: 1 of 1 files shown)


10.9-libcxx/stable/main/finkinfo/sci/gdl.info
```diff
@@ -76,6 +76,7 @@ pushd build
 	-DNCURSESDIR=%p \
 	-DNETCDF=ON \
 	-DOPENMP=OFF \
+	-DZLIBDIR=`xcrun --sdk macosx --show-sdk-path`/usr/include \
```
schwehr: Can you add to the commit description what this line is doing?

schwehr: Maybe something like:

    Find zlib with xcrun show-sdk-path for Mojave.

    Also works on.... Or doesn't work before...

    What changed to make this necessary?

remkos: Done in DescPort in commit be14375 (below).
<pre>Need to define ZLIBDIR using xcrun since 10.14 (XCode 10) since this
does no longer include a directory /usr/include, but needs to find it
in the MacOS SDK. Works also for previous versions, at least back to
Mac OS 10.9.
</pre>

PR #357, opened 2019-02-19
cdo-1.9.6-1: upstream update

remkos:
One less patch is needed with this update.

Built and tested with `fink -m build` on MacOS 10.14.3 with Command Line Tools 10.1


remkos:
Sorry, does not fully work yet. Forgot to test cdo-openmp, and that one fails configuration.

PR #358, opened 2019-02-24
Limit pm5162 to <= 10.13

nieder:
Now that #347 has been checked in, we need to restrict all pm5162 packages to <= 10.13.


dmacks:
Can this all be squashed, unless there is some reason for several passes?

nieder:
Only to not have ~800 files in a single commit. But that's a minor reason

Issue #359, opened 2019-02-28
multiple gccX build problems on 10.14


nieder:
On 10.14, when Xcode CLI 10.1 is installed, it's possible to not have /usr/include. This leads to the error seen here on gcc5, but I see it also with gcc7:
https://sourceforge.net/p/fink/mailman/message/36582179/
````shell
if [ -d ../prev-gcc ]; then \
          cd ../prev-gcc && \
          /Applications/Xcode.app/Contents/Developer/usr/bin/make real-install-headers-tar DESTDIR=`pwd`/../gcc/ \
            libsubdir=. ; \
        else \
          set -e; for ml in `cat fixinc_list`; do \
            sysroot_headers_suffix=`echo ${ml} | sed -e 's/;.*$//'`; \
            multi_dir=`echo ${ml} | sed -e 's/^[^;]*;//'`; \
            fix_dir=include-fixed${multi_dir}; \
            if ! false && test ! -d `echo /usr/include | sed -e :a -e 's,[^/]*/\.\.\/,,' -e ta`; then \
              echo The directory that should contain system headers does not exist: >&2 ; \
              echo "  `echo /usr/include | sed -e :a -e 's,[^/]*/\.\.\/,,' -e ta`" >&2 ; \
              tooldir_sysinc=`echo "/sw/lib/gcc5/lib/gcc/x86_64-apple-darwin18.2.0/5.5.0/../../../../x86_64-apple-darwin18.2.0/sys-include" | sed -e :a -e "s,[^/]*/\.\.\/,," -e ta`; \
              if test "x`echo /usr/include | sed -e :a -e 's,[^/]*/\.\.\/,,' -e ta`" = "x${tooldir_sysinc}"; \
              then sleep 1; else exit 1; fi; \
            fi; \
            /bin/sh ../../gcc-5.5.0/gcc/../mkinstalldirs ${fix_dir}; \
            chmod a+rx ${fix_dir} || true; \
            (TARGET_MACHINE='x86_64-apple-darwin18.2.0'; srcdir=`cd ../../gcc-5.5.0/gcc; ${PWDCMD-pwd}`; \
              SHELL='/bin/sh'; MACRO_LIST=`${PWDCMD-pwd}`/macro_list ; \
              gcc_dir=`${PWDCMD-pwd}` ; \
              export TARGET_MACHINE srcdir SHELL MACRO_LIST && \
              cd ../build-x86_64-apple-darwin18.2.0/fixincludes && \
[...]

PR #361, opened 2019-03-03
GCC fixes for 10.14

nieder:
Separate .info files for 10.14 packages.
* removed 32bit libraries
* allow building when /usr/include is not present
Also update to the latest upstreams for all of them.

Fixes issue #359


remkos:
A further point may be to start switching the gcc7 dependencies over to gcc8, now that that is rather mature.

Issue #363, opened 2019-03-08
Package python27-2.7.16-1 fails to build


remkos:
The latest python27-2.7.16-1 builds python, but the package cannot be completed as nine of the modules fail to build. Apparently only 9 failures are expected.
Maybe there are some missing `BuildDepends`?
Or the CFLAGS should use `-I$SDK_PATH/usr/include` instead of `-isysroot $SDK_PATH`?

The error message:
<pre>Python build finished, but the necessary bits to build these modules were not found:
bsddb185           dl                 imageop
linuxaudiodev      nis                ossaudiodev
spwd               sunaudiodev        zlib
To find the necessary bits, look in setup.py in detect_modules() for the module's name.
(Fink package build should have 7 missing)
</pre>

Building on MacOS 10.14.3 with Command Line Tools 10.1

remkos:
The fix for this issue is indeed replacing  `-isysroot $SDK_PATH` by  `-I$SDK_PATH/usr/include` as suggested above.
See PR #364.

danielj7:
Fixed by https://github.com/fink/fink-distributions/pull/364

gaboentropy:
I have this problem with python27 1:2.7.16-2
SO it seems like the problem was not fixed. My mac has mojave.

emmomp:
I am also getting this error installing  python27-2.7.17-1 with the following 
Package manager version: 0.45.0
Distribution version: selfupdate-rsync Mon Oct 28 14:26:35 2019, 10.14.5, x86_64
Trees: local/main stable/main
Xcode.app: 11.1
Xcode command-line tools: 10.3.0.0.1.1562985497

sth0:
There is a mismatch between Xcode and the command line tools.

Try updating the command line tools to version 11.1, I think you need to download them explicitly from the developer site.
<https://developer.apple.com <https://developer.apple.com/>>

There are still some packages that do not build under Xcode 11.1 so you may have found one of them.

-Scott

PR #364, opened 2019-03-08
python-2.7.16-2 and python-3.7.2-3: Avoid failing module builds on 10.14

remkos:
Replaced `-isysroot $SDK_PATH` by `-I$SDK_PATH/usr/include` in CFLAGS setting.
This is because `-isysroot` appears to suppress `-I%p/include`.

The same approach was used in `python34.info`, `python35.info`, `python36.info`

This fixed issue #363.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/python27.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python27.info
@@ -2,7 +2,7 @@
 Info2: <<
 Package: python%type_pkg[python]
 Version: 2.7.16
-Revision: 1
+Revision: 2
 Epoch: 1
 Type: python 2.7
 Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
@@ -82,7 +82,7 @@ CompileScript: <<
 	# for OS X 10.14, ensure full SDK is available, but include it after any fink specific versions.
 	if [ "$(uname -r | cut -d. -f1)" -ge 18 ]; then
 		SDK_PATH="$(xcrun --sdk macosx --show-sdk-path)"
-		export CFLAGS="-Wno-nullability-completeness -isysroot $SDK_PATH"
+		export CFLAGS="-Wno-nullability-completeness -I$SDK_PATH/usr/local"
 		export CPPFLAGS="-I%p/include -I%p/include/ncursesw"
 		perl -pi -e "s|/usr/include|$SDK_PATH/usr/include|" setup.py
 	fi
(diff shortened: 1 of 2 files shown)


nieder:
This is using $SDK_PATH/usr/local (which seems wrong in and of itself), while #363 says to use $SDK_PATH/usr/include.

My local temp solution was to patch things with:
`perl -pi -e "s|path.startswith\(\'\/Library|$&-fake|g" setup.py` to short-circuit a check when building python that scanned for includes that began with /Library. Don't know which one is better.

remkos:
Hi @nieder,
You are right that I mistyped `-I$SDK_PATH/usr/local` instead of `-I$SDK_PATH/usr/include`. But either way, I think the problem is actually the original existence of `-isysroot`. Probably just removing that would work as well.
In any case, I believe the using `-I$SDK_PATH/usr/include` works as it should, and it is the same as was done for the other `pythonxx.info`.

Unless you would like to promote your patch, feel free to approve my change of the .info files.
Builds and runs on MacOS 10.4.3 with Command Line Tools 10.1.

danielj7:
Only python27 and 37 specify -isysroot but not the other pythons. Do they build on 10.14? I'd rather they all do the same thing if possible and I don't have 10.14 to test with.

remkos:
@danielj7:
Indeed, python27 and 37 used `-isysroot` and hence *did not* build on 10.14.
The approach in my patch here, i.e. use `-I$SDK_PATH/usr/include` is exactly the same as used for the other python packages.
With my patch they all build on 10.14.

remkos:
Can I ask @nieder or @danielj7 to re-review?
These python27 and python37 info files are now in-line with the others.

nieder:
@danielj7 this normalizes all pythons to use `-I$SDK_PATH/usr/include` instead of the `-isysroot $SDK_PATH` that python27 and -37 are currently using during their build failure. Switching the 27/37 builds to use `-I$SDK_PATH/usr/include` lets them both build.

Only thing I would change would be to put the -I$SDK_PATH in CPPFLAGS in 27/37 since that's where py34/35/36 have it (vs CFLAGS). I don't know which is more proper, but might as well have them in the same place across the board.

remkos:
@nieder: I followed your suggestion in commit ce2506b.
Ready to re-review and approval.

nieder:
Ping @danielj7 
Ping @akhansen or @dmacks @fingolfin for other -core approval. This standardizes the python builds on 10.14 and lets them all build there (not just 34-36 as currently).

danielj7:
Sorry, I've been out of the loop for a bit. I now have a new machine that can run Mojave and am getting caught up on things. I Should have this fixed soon as well as updates to new python versions.

PR #366, opened 2019-03-13
netcdf-c15-4.6.3-1: upstream update and new library version

remkos:
Library version number changed from `libnetcdf.13.dylib` to `libnetcdf.15.dylib`, hence new package.
When accepted, I'll forthwith update packages to link against `netcdf-c15-shlibs`, where appropriate.

Removed from configuration `--disable-cdf5` which was introduced in package version 4.6.1 because of bug; fixed since 4.6.2.

All versions built and tested with `fink -m build` on MacOS 10.14.3 with Command Line Tools 10.1.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/netcdf-c15.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/netcdf-c15.info
@@ -1,0 +1,138 @@
+Package: netcdf-c15
+Version: 4.6.3
+Revision: 1
+BuildDependsOnly: true
+Maintainer: Remko Scharroo <remkos@users.sourceforge.net>
+
+Depends: %n-shlibs (= %v-%r)
+BuildDepends: doxygen, fink-package-precedence, hdf5.10, hdf5-cpp11, libcurl4, szip
+Conflicts: <<
+	netcdf-absoft,
+	netcdf, netcdf7,
+	netcdf-c7,
+	netcdf-c11,
+	netcdf-c13,
+	netcdf-c15
+<<
+Replaces: <<
+	netcdf-absoft,
+	netcdf,
+	netcdf7,
+	netcdf-c7,
+	netcdf-c11,
+	netcdf-c13,
+	netcdf-c15
+<<
+
+Source: ftp://ftp.unidata.ucar.edu/pub/netcdf/netcdf-c-%v.tar.gz
+Source-MD5: ef0b4d24f2c5a2a424c769cbb91fa45f
+Source-Checksum: SHA1(f165d9d4120dca09dcd49a90f31957640734ad0c)
+SetCFLAGS: -O2
+SetLDFLAGS: -lsz -Wl,-dead_strip_dylibs
+
+ConfigureParams: <<
+  --enable-shared --disable-static \
+  --mandir='${prefix}/share/man' \
+  --docdir='${prefix}/share/doc/%n' \
+  --disable-pnetcdf
+<<
+
+CompileScript: <<
+	#!/bin/sh -ev
+	%{default_script}
+	fink-package-precedence .
+<<
+InfoTest: <<
+	TestConfigureParams: --enable-extra-example-tests --disable-dap-remote-tests
+  	TestDepends: sed
+	TestScript: make -j1 check || exit 2
+<<
+InstallScript: <<
+  #!/bin/sh -ev
+  make install DESTDIR=%d
+  # Remove unwanted files from test suite
+  rm -f %i/bin/ocprint %i/lib/libmisc.* %i/lib/libbzip2.*
+  # Add C and CDL examples
+  mkdir -p %i/share/doc/%N/examples
+  cp -pr examples/C* %i/share/doc/%N/examples
+  rm -rf %i/share/doc/%N/examples/C/.{deps,libs}
+  rm -rf %i/share/doc/%N/examples/C/*/.{deps,libs}
+  rm -f %i/share/doc/%N/examples/C/*/*.la
+<<
+SplitOff: <<
+  Package: %N-shlibs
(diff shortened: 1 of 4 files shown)


nieder:
Since you're going to be pushing things to newer APIs, is it worth updating the hdf5 deps on this to the latest libN hdf5.100.v1.10 and hdf5-cpp.100.v1.10 (assuming it netcdf works with those)

nieder:
For the record, current netcdf-c15 built for me on 10.13, so it's good as is. Pushing hdf5 dep is just a suggestion.

remkos:
@nieder: This should not be a new API since we are merely going from version 4.6.2 to 4.6.3. 
But for some reason or another, Unidata decided to update the library version nonetheless.
As for the updating to the latest hdf5: it makes sense to do that now, indeed. So I will not merge and resubmit for review when done.

remkos:
@nieder, @akhansen : please review again after bumping HDF5 version to the latest, as requested.

10.9-libcxx/stable/main/finkinfo/sci/netcdf-c15.info
```diff
@@ -0,0 +1,136 @@
...
+
+Depends: %n-shlibs (= %v-%r)
+BuildDepends: doxygen, fink-package-precedence, hdf5.100.v1.10, libcurl4, szip
+Conflicts: <<
+	netcdf-absoft,
+	netcdf, netcdf7,
+	netcdf-c7,
+	netcdf-c11,
+	netcdf-c13,
+	netcdf-c15
```
schwehr: Is this correct?  c15 conflicts with itself?

dmacks: The fink .info processor automatically removes "self" from Conflicts and Replaces lists. That's so you can have a bunch of libversions with identical C/R to reduce differences among the .info and possible typos if you were to have to remember to exclude the correct one from each manually.

remkos: According to the [reference manual](http://www.finkproject.org/doc/packaging/reference.php?phpLang=en)
<pre>
If a package is listed in its own Conflicts, it will be (silently) removed from that list.
</pre>
It is just practical to do it that way, because then you can copy this to all other (older) versions the same way.

nieder:
You lost the BDep on hdf5-cpp.100.v1.10. W/out that installed, hdf5.h is not found. Doesn't look like there's any linkage to the cpp-shlibs, just header usage. With that added, it's fine.

nieder:
You lost the BDep on hdf5-cpp.100.v1.10. W/out that installed, hdf5.h is not found. Doesn't look like there's any linkage to the cpp-shlibs, just header usage. With that added, it's fine.

remkos:
@nieder: That's weird. I thought that was a bug in the previous HDF5 version. I thought I tested it properly without any of the hdf5-cpp headers installed. Will try again.

kamischi:
I can confirm the missing dependence on hdf5-cpp.100.v1.10. Otherwise, everything builds fine.

remkos:
@nieder, @kamischi: The dependency is back, with notes in `DescPackaging`.
Thank you for your inputs!
Ready for re-review and approval.

dmacks:
Did fink-package-precedence not detect the hdf5-cpp.100.v1.10 headers dependency? That is one of the purposes of creating that tool, so you may have found a weakness in its detection algorithm or else your package's build is not fully using its features.


````
While there is a BuildDepends on hdf5-cpp.100.v1.10, this is only because
one header file in that package should have been in hdf5.100.v1.10 instead.
````

If that's a true bug in the hdf5*v1.10 suite, how about that bug gets fixed there (ping akhansen) and then you only need a versioned dependency here on the corrected (not -cpp) package?

remkos:
@dmacks: Indeed if @akhansen would like to fix the issue of the header in the wrong place on his side, that would be nice.
For the time being, let us approve this version of netcdf-c15, otherwise the implementation of this NetCDF version gets further delayed. When @akhansen finds a way to fix the HDF5 packages, then I can always redo the NetCDF packages.
Regarding the fpp tool. I had not noticed it saying
<pre>
Found use of headers from 2 fink packages:
	hdf5-cpp.100.v1.10
	hdf5.100.v1.10
</pre>
because after this a lot of compilation and running of tests follows. Since it does not yield an error, the package built successfully.

kamischi:
Builds nicely, now. Looks good to me. MiSchi.

akhansen:
It's totally unclear to me from this what needs to be moved in the hdf5*100.v1.10 packages.  "One header" is pretty vague given the sheer number of headers in those packages.

$ dpkg -S hdf5.h
hdf5.100.v1.10: /sw/opt/hdf5.v1.10/include/hdf5.h

nieder:
For me, when building netcdf-c15 w/out hdf5-cpp.100.v1.10 installed, config.log shows the failure to find hdf5.h:
````
configure:18005: checking hdf5.h presence
configure:18005: gcc -E -I/sw/opt/hdf5.v1.10/include conftest.c
In file included from conftest.c:111:
In file included from /sw/opt/hdf5.v1.10/include/hdf5.h:24:
/sw/opt/hdf5.v1.10/include/H5public.h:69:10: fatal error: 'H5api_adpt.h' file not found
#include "H5api_adpt.h"
         ^~~~~~~~~~~~~~
1 error generated.
````

akhansen:
H5api_adpt.h has been moved to the main hdf5.100.v1.{8.10} packages as of 1.10.0-4, and I'm no longer maintaining them.

remkos:
Thanks @nieder for finding the mislocated header file.
Thanks @akhansen for moving the file to the right package.
I'll update the dependencies to include (>> 1.10.0-4) and removing the hdf5-cpp BuildDepends in all the packages affected.

PR #370, opened 2019-03-18
gdal2-2.4.0-1: upstream and dependency update

remkos:
Updated from version 2.3.1 to 2.4.0. In addition, changed dependencies:
hdf5.10 -> hdf5.100.v1.10
hdf5-cpp1 -> hdf5-cpp.100.v1.10
netcdf-c11 -> netcdf-c15

Built and tested with `fink -m build gdal2` on MacOS 10.14.3 with Command Line Tools 10.1


kamischi:
It builds and installs fine on 10.14.3, but how about updating gdal-py as well?

remkos:
Good point, @kamischi. It is indeed worthwhile to update these in-sync.
Hence I added the commit 2a5c579 just now for gdal-py.
Tested also on MacOS 10.14.3.

remkos:
Now also removed the dependency on hdf5-cpp, as per the discussion here in PR #366 
This PR is ready for review and approval.

PR #371, opened 2019-03-18
netcdf-fortran7-4.4.5-1: upstream and dependencies update

remkos:
Updated from version 4.4.4 to 4.4.5. En passant updated dependencies:
gcc7 -> gcc8
netcdf-c13 -> netcdf-c15.

Built and tested with `fink -m build netcdf-fortran7` on MacOS 10.14.3 with Command Line Tools 10.1


remkos:
Thanks, @kamischi, for that comment. I'll review those difference for a future version.

PR #373, opened 2019-03-18
eccodes-2.12.0: upstream and dependencies update

remkos:
Updated from version 2.10.0 to 2.12.0. En passant changed dependencies:
gcc7 -> gcc8
netcdf-c13 -> netcdf-c15

Removed dependencies on hdf5: they were not used.

Built and tested with `fink -m build eccodes` on MacOS 10.14.3 with Command Line Tools 10.1

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
@@ -12,8 +12,8 @@ Source: https://software.ecmwf.int/wiki/download/attachments/45757960/%n-%v-Sour
 Source-MD5: a812b5837901526905b020d328d8007c
 Source-Checksum: SHA1(dff83137e95520462118a5cc65b58157d40e736d)
 PatchFile: eccodes.patch
-PatchFile-MD5: 43ec7477a96e37e7a951faae7019622b
-PatchFile-Checksum: SHA1(6d04866b20c96813524a46b3a365844d4866fb1f)
+PatchFile-MD5: 4fc2a180dcdf2c577ce213884a297fb4
+PatchFile-Checksum: SHA1(f9af183e1998bee05317c2f2e3a2d28a5e1fb51f)
 
 SourceDirectory: %n-%v-Source
 BuildDependsOnly: true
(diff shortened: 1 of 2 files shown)


remkos:
@nieder: I found that HDF5 is actually *not at all* used in the ecCodes code. No headers are included, nor are any library functions called.
Nonetheless, the CMake system calls to find HDF5 and adds the location of the headers and the library to the location of the NetCDF headers and library, which the compilation of `grib_to_netcdf` then uses. 
This is in fact the *only* place where the HDF5 header path or library is added. Yet `grib_to_netcdf.c` *does not* include anything related to HDF.

I looked if there would be the CMake equivalent of `--disable=hdf5` so that it will not be searched, nor is arbitrarily added to the path. But I have not found a way (yet). I may need to patch a CMake file to exclude the HDF5 header path and library irrespectively of whether it is found and make a clear note in the DeskPackaging.

remkos:
Searching for the HDF5 package has been disabled in the change to the patch file `eccodes.patch`.
`DescPackaging` updated accordingly. See commit 617add7.
This should solve the failure to compile `grib_to_netcdf`.
Let me know if it works now for you, @nieder

dhomeier:
@remkos, this still fails to build for me (10.12, Xcode 9.2) at least erratically
```
[ 65%] Linking C executable ../bin/grib_to_netcdf
ld: library not found for -lhdf5
```
which is, more precisely, this command
```
/sw/var/lib/fink/path-prefix-clang/cc -MD -O3 -DNDEBUG -Wl,-search_paths_first -Wl,-headerpad_max_install_names -headerpad_max_install_names -L/sw/lib CMakeFiles/grib_to_netcdf.dir/grib_to_netcdf.c.o  -o ../bin/grib_to_netcdf -Wl,-rpath,@loader_path/../lib libgrib_tools.a /sw/lib/libnetcdf.dylib -lhdf5 /sw/lib/libsz.dylib /usr/lib/libz.dylib /usr/lib/libdl.dylib /usr/lib/libm.dylib ../lib/libeccodes.0.dylib /sw/lib/libpng.dylib /sw/lib/libjasper.dylib /sw/lib/libjpeg.dylib /sw/lib/libopenjp2.7.dylib /usr/lib/libz.dylib /usr/lib/libm.dylib
```
where indeed the linker path to `libhdf5[.100].dylib` is missing.
It looks like as long as any of the hdf5 header packages are installed the `-lhdf5` still gets sneaked in - at least after removing all of `hdf5-cpp.100.v1.10` `hdf5.10-gfortran` `hdf5.100.v1.10`, the build succeeded again. Might be necessary to make them (and perhaps `hdf5.9`, `hdf5.10`, `hdf5.100.v1.8` and the other `hdf5-cpp` versions as well) `BuildConflicts`.

remkos:
@dhomeier: Yes, this seems to appear for other users as well. I hate to put in a `BuildConflicts` just because the CMake machinery want to find HDF. If I cannot find a better way, I'll follow your advice.

nieder:
Try adding `-LAH` as an extra flag to the cmake command. This will output all the things cmake found. You can then use that to figure out what bit found hdf5 and try to explicitly set it to something purposely broken to stop HDF5 detection (or there might even be a hidden NO_HDF5 flag or somesuch).

remkos:
See new PR #382

PR #377, opened 2019-03-19
Updated ncview to use netcdf-c15

PovlAbrahamsen:
This is an updated package for ncview, to use the latest version of the Netcdf libraries. Also, I spotted that when Fink's libxt is installed, the program is linked to these, so I have included them as a dependency.


PovlAbrahamsen:
I just spotted that @remkos has submitted a very similar patch, without asking/informing me. If he wishes to take over as maintainer of this package, I would appreciate if you could remove me as maintainer of ncview.

PR #378, opened 2019-03-19
jbig2dec 0.15 (retrieve from GitHub releases page)

krbullock:


akhansen:
No new revision is technically required, because changing the source location doesn't affect the final package build.

dmacks:
Noting also that upstream is now at version 0.15, vs our 0.11.

krbullock:
I also saw the update to 0.15, but was skeptical of updating to that because

1. It's not available on the SourceForge project, only from a dedicated domain, https://jbig2dec.com/ (which is puzzlingly different from the version 0.14 available from https://artifex.com/jbig2dec/).
2. It's been relicensed from GPL to AGPL, and I'm not sure if that affects anything in Fink that builds against it. It's possible it would introduce a license incompatibility.

krbullock:
@akhansen I've removed the revision bump from both trees.

akhansen:
Actually, both 0.11 and 0.15 seem to be available through https://github.com/ArtifexSoftware/jbig2dec/releases , though I haven't verified that they're identical.

krbullock:
Looks like the 0.11 release on GitHub has a different checksum (`cb2a3ee532587045e3b4c1e245bdd2e7` vs `1f61e144852c86563fee6e5ddced63f1`) and a not-trivially-different file list.

nieder:
Github releases (usually named vX.Y.Z.tar.gz) are often just tags of the repo without needed generated files.
How does the `jbig2dec-0.15.tar.gz` file from here look:
https://github.com/ArtifexSoftware/ghostpdl-downloads/releases/tag/gs924 ? That's identical to the file from https://www.jbig2dec.com/ and is what MacPorts is using.

krbullock:
Looks like 0.16 has dropped on https://github.com/ArtifexSoftware/ghostpdl-downloads/releases along with Ghostscript 9.27. Since 0.15 is the latest version that seems to go along with Ghostscript 9.26, I've updated this PR to install 0.15 from GitHub, and leave the upgrade of Ghostscript for a separate effort.

akhansen:
The 0.15 update works for me.  Let's give @danielj7 a couple more days to respond, and if we don't hear from him I'd say we invoke fink-core privilege and merge.

akhansen:
Since it's been a week and the change is a relatively small one, I'm going to step in and do this merge.

krbullock:
@akhansen What do we do with a branch once it's been merged? Should I go ahead and delete it or do you prefer to keep them around?

akhansen:
I normally delete them to minimize clutter.  I forgot to do that when I did the merge, but just did it.

PR #379, opened 2019-03-20
Ugly hack for Xcode 10

akhansen:
Xcode 10 seems no longer to honor the *ROOT variables from the CompileScript and only uses their default values (/tmp/telnet/*).


dmacks:
Can you ```mkdir /tmp/%n || exit 2``` in CompileScript prior to running apple's broken build system?

akhansen:
Yeah, makes sense.

PR #382, opened 2019-03-25
eccodes-2.12.0-2: Getting rid of searching for HDF5

remkos:
Despite eccodes not using HDF5, its CMake system tries to find it in two places. One place was already commented out, the second was missed and is added in this patch.
This should allow building eccodes without unnecessary dependencies, irrespective of whether HDF5 is already installed.

See e-mail "phase compiling: eccodes-2.12.0-1 failed" to fink-beginners@lists.sourceforge.net and comment by @dhomeier in PR #373 

Built and tested on MacOS 10.14.3 with Command Line Tools 10.1

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/eccodes.info
@@ -1,7 +1,7 @@
 Info2: <<
 Package: eccodes
 Version: 2.12.0
-Revision: 1
+Revision: 2
 Type: gcc (8)
 Description: Coding/encoding ECMWF files, C headers/docs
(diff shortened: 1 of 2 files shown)


dhomeier:
I haven't been able to reproduce the linking error since, even with the original revision and all of `hdf5-cpp.100.v1.10`, `hdf5.100.v1.10`, `hdf5.100.v1.10-bin`, `hdf5.100-gfortran.v1.10` installed, with `UseMaxBuildJobs` set to either true or false...
But the whole lot of `HDF5_*` defines previously showing up has disappeared from the build log now, so hopefully this will have done it.

remkos:
Thanks, @dhomeier, for double checking that.
This indeed was intended to work irrespective of the installation of any of the packages you mentioned.

PR #383, opened 2019-03-26
Updated dependencies gcc7 -> gcc8

remkos:
This updates four of the sci packages to use gcc8 instead of gcc7, i.e.:
fftw3
openmpi
pgplot
wip

pgplot also was patched to work with MacOS 10.14 (where there is no /usr/include).
Otherwise, it is a revision bump and replacement of `Type: gcc (7)` by `Type: gcc (8)` for all four packages.

Built and tested on MacOS 10.14.3 with Command Line Tools 10.1

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/openmpi.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/openmpi.info
@@ -1,12 +1,12 @@
 Info2: <<
 Package: openmpi
 Version: 1.10.7
-Revision: 2
+Revision: 3
 GCC: 4.0
 Description: MPI implementation for parallel computing
 License: BSD
 Maintainer: Remko Scharroo <remkos@users.sourceforge.net>
-Type: gcc (7)
+Type: gcc (8)
 Depends: <<
 	%N-shlibs (= %v-%r),
 	gcc%type_raw[gcc]-compiler,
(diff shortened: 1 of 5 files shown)


dhomeier:
All built and tested successfully on 10.12.7 with Xcode 9.2, Command Line Tools 9.0.

But I did get a segfault on executing the `cpgdemo` program as soon as switching to the second page, both with the `/aqt` and `/xwin` backends.

I was able to fix it by also changing the C compiler to gcc8:
```
perl -pi -e 's|g77|gfortran-fsf-%type_raw[gcc]|g;s|(CCOMPL=.)(gcc)|${1}gcc-fsf-%type_raw[gcc]|;'
```
-- reminiscent of similar problems I've seen when mixing gfortran with (Xcode's) clang, on moving `scipy-py` to gcc8 in #385.
Note that I did not change the Objective C compiler `MCOMPL`, as I could not get `gcc-fsf-8 -x objective-c` to handle the `aqdriv.m` source file (`clang-5.0 -ObjC++` did compile it, but then the final linking with gfortran failed). Nonetheless `cpgdemo` now runs without segfaults.

remkos:
Thanks, @dhomeier, for catching that and your easy patch.
I implemented it in commit decc0b0. 
Ready for re-review and approval.

dhomeier:
LGTM, unless someone wants to test them on 10.13 or <= 10.11 as well.

nieder:
Good on 10.13.

remkos:
Thanks @nieder for checking also on 10.13

PR #385, opened 2019-03-27
Scipy update to 1.2.1 and Python 3.7 built against new OpenBLAS package

dhomeier:
The packaging in this PR is not necessarily finalised, but I am putting this already up for discussion.

Rationale: updating the existing `scipy-py` package to upstream 1.1.0 introduces about a dozen segfaults in the test suite; from 1.2.0 on Scipy no longer officially supports building against Accelerate due to its outdated API - trying to do so introduced another bunch of segfaults in the `linalg` tests. Experimenting with reintroducing ATLAS supports produced rather worse results.
Thus I jumped into the water and packaged OpenBLAS 0.3.5 (the version current Scipy wheels are built against), largely copying from the [MacPorts package](https://github.com/macports/macports-ports/blob/master/math/OpenBLAS/Portfile).

The commits here build for Python versions 2.7-3.7 and pass all tests except for 5-6 failures marked to skip in the `TestScript`. Both builds have been based on `gcc-fsf-8` throughout, since the combination of `clang` and `gfortran` in either or both of `openblas` and `scipy-py` still produced some segfaults in the Scipy tests (in `test_blas`, `test_cython_blas`, `test_nonlin` and `test_arpack`).

Some other specific decisions in the `openblas` packaging:

- `TestScript` builds timing executables in the `benchmark` subdirectory, but does not run them - have not thought of a way to process the ~200 lines of timing output from each in a useful manner. Results are somewhat variable, but generally the OpenBLAS (`*.goto`) ones are slightly faster to 20-30 % slower than the `veclib` equivalents, and 50-100 % faster than the `atlas` ones`.

- Overriding the automatic optimisation flags to create a binary to run on most (all?) `x86_64` hardware still supported. This might carry a small performance penalty compared the build with `-avx2` that is otherwise created on my Core i7 MB Pro, but not very significant given the variation in the timings. Is there still any PowerPC support in Fink?

- The current `openblas.info` is installing directly to `%p/lib` and `%p/include`. This creates a conflict with the `cblas.h` header also installed by `atlas`. Several options to overcome the conflict:
-- MacPorts is renaming the file to `cblas_openblas.h`, but I don't know what their suggestion to build against the OpenBLAS version is.
-- The headers could be moved out into an additional `-dev` `SplitOff`.
-- The entire package could be installed into an `openblas` subdirectory.
-- The headers could be moved into their own `%p/include/%n` subdir.
If the conflict with `atlas` is indeed seen as a problem, I would prefer the last solution - pointing the scipy build to different `include_dirs` in `site.cfg` is trivial, and it would probably also be the least hassle for other packages.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
@@ -1,0 +1,94 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info2: <<
+Package: openblas
+Version: 0.3.5
+Revision: 1
+Type: gcc (8)
+Maintainer: Derek Homeier <dhomeie@gwdg.de>
+
+BuildDependsOnly: true
+BuildDepends: fink (>= 0.30.0), gcc%type_raw[gcc]-compiler
+Conflicts: atlas
+
+Source: https://github.com/xianyi/OpenBLAS/archive/v0.3.5.tar.gz
+Source-MD5: 579bda57f68ea6e9074bf5780e8620bb
+SourceRename: OpenBLAS-%v.tar.gz
+PatchFile: %n.patch
+PatchFile-MD5: ceb1e323762c1dc7ff1896787ef9140b
+
+PatchScript: sed 's|@PREFIX@|%p|g' < %{PatchFile} | patch -p1
+
+SetCC: gcc-%type_raw[gcc]
+
+CompileScript: make FC=gfortran-fsf-%type_raw[gcc] COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1 USE_THREAD=1 libs netlib re_lapack shared
+
+InstallScript: <<
+	make FC=gfortran-fsf-%type_raw[gcc] COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1 USE_THREAD=1 PREFIX=%i install
+	install_name_tool -id %p/lib/libopenblas-r%v.dylib %i/lib/libopenblas-r%v.dylib
+	perl -pi -e 's|%i|%p|;' %i/lib/pkgconfig/openblas.pc %i/lib/cmake/openblas/OpenBLASConfig.cmake
+<<
+
+InfoTest: <<
+	TestScript: <<
+		#!/bin/sh -ev
+		make FC=gfortran-fsf-%type_raw[gcc] COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1 USE_THREAD=1 tests | tee tests.log
+		grep -i fail tests.log
+		grep -c PASSED tests.log
+		cd benchmark
+		make FC=gfortran-fsf-%type_raw[gcc] COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1 USE_THREAD=1 veclib goto # atlas
+		# this creates a ton of *.veclib, *.goto, *.atlas executables to be compared in performance...
+	<<
+<<
+
+Splitoff: <<
+	Package: %N-shlibs
+	Depends: gcc%type_raw[gcc]-shlibs
+	DocFiles: LICENSE USAGE.md
+	Files: lib/libopenblas-r%v.dylib
+	Shlibs: %p/lib/libopenblas-r%v.dylib 0.0.0 %v (>= 0.3.5-1)
+<<
+
+DocFiles: LICENSE README.md USAGE.md Changelog.txt CONTRIBUTORS.md BACKERS.md GotoBLAS_0*.txt benchmark/scripts
+License: BSD
+Homepage: https://www.openblas.net
+
+Description: Optimized BLAS library based on GotoBLAS2
+DescDetail: <<
+BLAS stands for Basic Linear Algebra Subprograms. BLAS provides standard
+interfaces for linear algebra, including BLAS1 (vector-vector operations),
+BLAS2 (matrix-vector operations), and BLAS3 (matrix-matrix operations).
+In general, BLAS is the computational kernel ("the bottom of the food chain")
+in linear algebra or scientific applications. Thus, if BLAS implementation is
+highly optimized, the whole application can get substantial benefit.
+
+As BLAS is a standardized interface, you can refer to the documentation of its
+reference implementation at netlib.org. Calls from C go through its CBLAS
+interface, so your code will need to include the provided cblas.h in addition
+to linking with -lopenblas.
+A single-precision matrix multiplication will look like
+
+#include <cblas.h>
+...
+cblas_sgemm(CblasRowMajor, CblasNoTrans,
+            CblasNoTrans, M, N, K, 1.0, A, K, B, N, 0.0, result, N);
+
+where M,N,K are the dimensions of your data.
+
+OpenBLAS is an open source BLAS library forked from GotoBLAS2-1.13 BSD version.
+Since Mr. Kazushige Goto left TACC, GotoBLAS is no longer being maintained.
+Thus this project was created to continue developing OpenBLAS/GotoBLAS.
+<<
+DescPackaging: <<
+Setup with architecture-independent version and some patches from MacPorts
+https://github.com/macports/macports-ports/blob/master/math/OpenBLAS/Portfile
+Build with FSF gcc8 to ensure optimum compatibility with gfortran.
+cblas.h conflicts with header from atlas.
+TestScript also builds executables in benchmark/, but currently does not run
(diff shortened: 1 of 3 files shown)


dhomeier:
I made some more systematic benchmarks using the scripts in `/sw/share/doc/openblas/scripts/SCIPY`, comparing the old scipy to `1.2.1` built against an openblas complied with `USE_THREAD=1` + `COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1` and `COMMON_OPT='-O2 -mtune=intel'` (invoking `-mavx2`), respectively:
```
ssyrk.py From: 128 To: 2048 Step=128 Loops=6
 scipy-py36    1.0.0 (Veclib)      1.2.1 (-msse NO_AVX2=1)     (-mavx2)
        SIZE                                 MFlops      

     128x128 :     15787.184605          22837.108669          21358.380467  
     256x256 :     50862.843641          41507.320592          41527.733359  
     384x384 :     69947.333085          49728.391890          70181.593262  
     512x512 :     91835.228943          44202.050913          87645.952580  
     640x640 :     88130.140091          48080.094502          61219.455709  
     768x768 :     85340.144908          46516.436419          64641.213450  
     896x896 :     92022.343975          57544.167943          78616.014903  
   1024x1024 :     96675.233228          63659.767296          85488.119576  
   1152x1152 :    118775.849986          75882.078555         102815.323322  
   1280x1280 :    121391.000042          79855.467208         115715.481262  
   1408x1408 :    131785.698105          80979.514502         111001.794905  
   1536x1536 :    123116.779307          83437.381592         124799.204735  
   1664x1664 :    130561.202377          81952.161902         127523.352557  
   1792x1792 :    130716.169430          78741.857977         128051.775972  
   1920x1920 :    132310.590886          84072.067746         126570.583570  
   2048x2048 :    115587.067009          85607.524476         127974.983195 
```
```
dsyrk.py From: 128 To: 2048 Step=128 Loops=6
 scipy-py36    1.0.0 (Veclib)      1.2.1 (-msse NO_AVX2=1)     (-mavx2)
        SIZE                                 MFlops      

     128x128 :     11791.009413           2973.327219          16685.601686
     256x256 :     38985.453838          20809.919911          26973.261679
     384x384 :     57282.805499          35418.748002          37866.840360
     512x512 :     53747.370004          22835.873496          29924.780244
     640x640 :     52952.359968          27902.491474          35053.691925
     768x768 :     58184.482538          34207.351020          39538.486948
     896x896 :     55567.743721          34880.582217          43291.027106
   1024x1024 :     54435.018794          33447.952968          48026.784216
   1152x1152 :     57897.945090          42036.080168          51394.202145
   1280x1280 :     57906.049737          41198.855703          53151.435206
   1408x1408 :     58505.577222          44049.993055          54617.338904
   1536x1536 :     57497.727119          44590.160726          50059.800038
   1664x1664 :     60202.543208          41983.418182          57097.264316
   1792x1792 :     59100.385538          40499.101898          56646.589548
   1920x1920 :     61026.196972          44233.337820          55008.195816
   2048x2048 :     49881.207350          43630.776228          56814.418201
```
I.e. quite substantial speedup after all with the AVX2 instructions especially for single precision!
So maybe we should keep them enabled in the `openblas.info` file and just create a special .deb with `NO_AVX2=1` for the binary distribution...

nieder:
Need to fix the SetCC field:
````
make FC=gfortran-fsf-8 COMMON_OPT='-O2 -msse -mtune=intel' NO_AVX2=1 USE_THREAD=1 libs netlib re_lapack shared
make: gcc-8: Command not found
````
The dylib is currently installed as a fully versioned file. When %v gets bumped to 0.3.6, that filename will change and anything linking to it will be broken.
How stable is the API? If it's stable, can the file name be adjusted (say libopenblas.0.dylib) and then bump the 0 if the API changes? Otherwise, any new %v will need a new package (probably with %v in package name: `libopenblas-v%v-shlibs` or some such).

For the conflicting header with atlas: are the two libraries doing the same thing? Or are they different roles that just happened to name a header the same? Say openblas headers get installed into private include subdir and atlas is still installed. What are the chances that builds will first find the the naked atlas cblas.h header before the desired openblas/cblas.h? In short, should the `Conflicts: atlas` be changed to `Conflicts/Replaces: atlas`?

If you move the headers into a private subdir, 3rd party packages will still look for `#include <lapacke.h>`, so `openblas.pc` and `OpenBLASConfig.cmake` will probably need to be adjusted to include the private subdir in the output for -I flags (or patch them all to use `#include <openblas/lapacke.h>`.

**Comment by dhomeier** [comment]
Sorry about the glitch with `CC`; think I had made the first build tests with `%p/lib/gcc8/bin` in the path and then tried to put as much of the configuration into the general package setup as possible.

- [versioning] I had only adopted part of the MacPorts patch, which instead creates a more robust library name `libopenblas_r1.dylib`, so the API is probably rather stable. As a test I just built version 0.3.3 of the library (manually renamed to `r0.3.5`) and successfully tested the `scipy.linalg` built against 0.3.5 with it, so I expect it's safe for newer (minor) versions as well. Debian has a similar naming scheme for their package (`libopenblas.so.0`), so I'd go with your suggestion.

- [header name] they provide largely the same functions and define compatible APIs with small syntactic variations, e.g. for ATLAS, OpenBLAS, and the system's Veclib symbols like
```
float  cblas_sdsdot(const int N, const float alpha, const float *X,
                    const int incX, const float *Y, const int incY);
```
```
float  cblas_sdsdot(OPENBLAS_CONST blasint n, OPENBLAS_CONST float alpha, OPENBLAS_CONST float *x, OPENBLAS_CONST blasint incx, OPENBLAS_CONST float *y, OPENBLAS_CONST blasint incy);
```
```
float  cblas_sdsdot(const int __N, const float __alpha, const float *__X,
        const int __incX, const float *__Y, const int __incY)
```
although the OpenBLAS one has a couple more (169 vs. 149) defines than the older versions. Note that  the lapack-3.x sources also come with their own version of `cblas.h`, but none of the Fink packages are installing it. I've experimented with building the new Scipy as well as `openblas` versions of `arpack-ng`, `qrupdate` and `octave382` with the include file renamed to `/sw/include/cblas_openblas.h`, but none of them seemed to bother (Scipy actually includes its own copy of cblas.h for the only source file using it, the other packages are apparently using mostly or exclusively the Fortran interface).
Finally, the [Debian package](https://packages.debian.org/sid/i386/libopenblas-dev/filelist) has adopted the renaming approach as well.

- [atlas ] with `Replaces: atlas` atlas build-depends still ask for installing the real thing (and they have to, since they generally want to link to `-l[st]atlas`), so the main effect seems to be that Fink can smoothly install `openblas` over `atlas` without the user having to resort to `apt-get` - I assume for this to work both ways we would also need a new `atlas` revision which `Replaces: openblas`?

Still not sure whether to retain the `Conflicts`/`Replaces`, but I don't think there is much benefit in moving all the headers into their private subdirectory (only potential use I can think of is that one could create custom `cblas.pc` and `lapack.pc` files analogously to the `lapack3x` ones that would allow to get the `openblas` configuration via `PKG_CONFIG_PATH=/sw/lib/openblas/pkgconfig pykg-config --libs blas` etc.)

nieder:
* header name: everyone seems to be renaming to cblas_openblas.h, so might as well go with that in case it becomes some sort of unofficial standardization. If there's no more file conflicts, then we don't need standalone `Conflicts`. I think `Conflicts/Depends` would only be used if the plan is for openblas to be used as a direct replacement for atlas and that would allow easy swapping between the two. Smooth swapping would need %r bump in atlas to add matching `C/R` to it. 
* Hiding things in subdirs should be last option.
* dylib naming: moving the %v out of the libopenblas dylib name is probably better long term, and matches Debian. Do you know why Debian has several .so files?: https://packages.debian.org/sid/i386/libopenblas-base/filelist
* fix `SetCC` field.
* openblas package needs to Dep on openblas-shlibs.

dhomeier:
> * dylib naming: moving the %v out of the libopenblas dylib name is probably better long term, and matches Debian. Do you know why Debian has several .so files?: https://packages.debian.org/sid/i386/libopenblas-base/filelist

Not really. There are some different library names used in the Makefiles, but in the Darwin build they all point to `libopenblas`, and accordingly that includes all the lapack, lapacke functions etc.
I also kind of expected that `USE_THREAD=1` would create something like a `_p` or `_pt` version of the library, but that is not the case, although multithreading is supported with this build.

dhomeier:
This has AVX2 support back in on CPUs supporting it.

dhomeier:
Small but significant modification of the build configuration to avoid a problem that turned up with the new Scipy release candidate [here](https://github.com/scipy/scipy/pull/10075#issuecomment-489333442).

PR #397, opened 2019-04-16
Lazarus: Update to new upstream version (2.0.2)

kamischi:
Includes qt5pas. All builds now on 10.14.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/lazarus-doc.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/lazarus-doc.info
@@ -1,14 +1,14 @@
 Package: lazarus-doc
-Version: 1.8.4
+Version: 2.0.2
 Revision: 1
-Enhances: lazarus-aqua, lazarus-gtk2, lazarus-qt4, lazarus-qt5
+Enhances: lazarus-aqua, lazarus-cocoa, lazarus-gtk2, lazarus-qt4, lazarus-qt5
 Recommends: xchm
 
 License: GPL/LGPL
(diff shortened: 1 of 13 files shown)


nieder:
Does lazarus-lcl-cocoa-32bit actually build on 10.14? I thought 32bit libraries couldn't be built there anymore.

kamischi:
> Does lazarus-lcl-cocoa-32bit actually build on 10.14? I thought 32bit libraries couldn't be built there anymore.

It definitely builds, as does lazarus-aqua and lazarus-lcl-carbon, both of which are 32 bit. You get a warning about deprecation, when linking lazarus-aqua, but that is all. At least on my system.

nieder:
Seems fine with the possible issue about 32bit libraries on 10.14.

kamischi:
> Seems fine with the possible issue about 32bit libraries on 10.14.

The most critical test is lazarus-aqua. Since I am not 100% sure about the cleanliness of my system, I would definitely appreciate a test on a second, presumably clean system.

nieder:
I left lazarus-lcl-cocoa-32bit building on a 10.14 VM when I left this morning. Will check it tonight. I don't remember if I told it to use lazarus-aqua or another lazarus variant (think it was aqua since it was option#1).

kamischi:
> I left lazarus-lcl-cocoa-32bit building on a 10.14 VM when I left this morning. Will check it tonight. I don't remember if I told it to use lazarus-aqua or another lazarus variant (think it was aqua since it was option#1).

lazarus-lcl-cocoa-32bit builds with any variant of lazarus. The lazarus-lcl-... packages are just lazarus component libraries including some cross-platform ones. To the best of my knowledge they do not involve any linker action and contain no executables, contrary to the lazarus variants, which build the ide as a double clickable application as well as some other executables.

PR #399, opened 2019-04-18
New upstream jbig2dec source

akhansen:


krbullock:
Looks like the one I'd already submitted at #378? Are we still interested in updating to fetch from https://github.com/ArtifexSoftware/ghostpdl-downloads instead?

akhansen:
Ah, totally forgot about yours.

PR #404, opened 2019-04-23
Use a custom mirror to search all sodium source locations.

### LINKED ISSUE #403: libsodium (<1.0.16) sources moved
Upstream moved libsodium sources to an archive directory.
Previous: http://download.libsodium.org/libsodium/releases
Now: https://download.libsodium.org/libsodium/releases/old/unsupported/

This can be dealt with by using CustomMirror to search both directories (and the intermediate `releases/old` one that also hosts some sources).

nieder:
Fixes #403


kamischi:
The CustomMirror trick worked here. As expected, building and installing was no problem, so lgtm.

nieder:
Pushing because this is a non-change-of-function change that is needed to build.

PR #405, opened 2019-04-23
Compile CERNLIB with gcc8

mommsen:
Update CERNLIB and patchy to use gcc8 instead of gcc6.


nieder:
Does it matter if cernlib2006 is compiled with gcc8 while BuildDepending on an older patchy4-gfortran using gcc6? Or does the BuildDepends: patchy4-gfortran need to be versioned to make sure it's using %v=12?

mommsen:
It should not matter which gcc version was used for building patchy4 wrt cernlib.

Issue #393, opened 2019-04-02
jupyter/nbconvert/nbformat script setup


dhomeier:
I have just submitted PR #391 as first of a two-part attempt to replace the stale update in #164 to the IPython + Notebook system. The actual `jupyter-*`, `nb*` and `ipython-py` and `notebook-py` updates themselves with their mess of circular dependencies are to follow in a second PR.
In the course of this I have tried to understand the function of the various jupyter packages and scripts a bit better and found the current setup rather complicated and possibly confusing.

Very briefly, `jupyter-core-pyXX` provides the script `jupyter[XX]`, which acts as a launcher or (thin?) wrapper for the various commands provided by `jupyter-client-pyXX`, `nbformat-pyXX`, `nbconvert-pyXX` and `notebook-pyXX`. It can detect and launch any commands in `$PATH` under the name of `jupyter*`. E.g. with the current `py27` and `py35` versions of the packages installed, one would find
```
 jupyter2.7 -h                                                                                                                         
usage: jupyter2.7 [-h] [--version] [--config-dir] [--data-dir] [--runtime-dir]
...
Available subcommands: bundlerextension kernel kernelspec kernelspec2.7
kernelspec3.5 migrate migrate2.7 migrate3.5 nbconvert2.7 nbconvert3.5
nbextension run serverextension trust trust2.7 trust3.5

> jupyter2.7 kernel -h                                                                                                                  
jupyter: 'kernel' is not a Jupyter command
```
Here things begin becoming messed up, since some of the scripts are not managed under `update-alternatives` (and some currently are not working anyway). But even where they are, calling a versioned `jupyter` launcher with a command will execute the highest installed version of that command script, e.g. in the above example
`jupyter2.7 nbconvert` would launch the `jupyter-nbconvert3.5` script (under `python3.5`), since the underlying `command.py` is calling
```
    command = 'jupyter-' + subcommand
    try:
        _execvp(command, sys.argv[1:])
    except OSError as e:
        sys.exit("Error executing Jupyter command %r: %s" % (subcommand, e))
```
Maybe not a critical issue, since apparently all the subcommand scripts can be called directly, but I'd at deem it at least confusing.

[...]

PR #413, opened 2019-05-08
IPython update and Jupyter/Notebook infrastructure

dhomeier:
Finally bringing around the updated Jupyter ecosystem for Python up to 3.7.
Following the discussion in #393 I have patched the calling interface of `%p/bin/jupyter-pyNN` (provided by `jupyter-core-pyNN`) to only launch subcommands of the matching Python variant.
Call `jupyter -h` or `jupyter-pyNN -h` for a list of available/installed commands.
Scripts under `update-alternatives` are named in the `*-pyNN` scheme now.

Thanks to the still persistent mess of circular dependencies I could not come up with a straightforward path for updating *and* testing all packages; in particular both the doc build and tests for `ipython-py` depend on `ipykernel-py` and `nbsphinx-py`. The best way to check should be something like
```
fink -m update nbsphinx-pyNN
fink --validate update ipython-pyNN
fink -m update ipykernel-pyNN
fink -m rebuild ipython-pyNN ipywidgets-pyNN jupyter-core-pyNN jupyter-client-pyNN
fink -m update notebook-pyNN
```
The Ipython docs are built conditional on the availability of sphinx+nbsphinx+ipykernel, which means that different packages will be build depending on the installation status of these packages...
Python 2.7 variants of ipython and ipykernel come in their LTS versions.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/ipython-docs-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/ipython-docs-py.info
@@ -1,3 +1,4 @@
+# -*- coding: ascii; tab-width: 4; x-counterpart: ipython-docs-py.patch -*-
 Info2: <<
 Package: ipython-docs-py%type_pkg[python]
 Version: 7.5.0
@@ -6,27 +7,36 @@ Type: python (3.5 3.6 3.7)
 Maintainer: Derek Homeier <dhomeie@gwdg.de>
 
 Depends: <<
-  ipython-py%type_pkg[python] (= %v-%r)
+    python%type_pkg[python]
 <<
 
 BuildDepends: <<
-  python%type_pkg[python],
-  ipython-py%type_pkg[python] (= %v-%r),
-  ipykernel-py%type_pkg[python],
-  setuptools-tng-py%type_pkg[python],
-  sphinx-py%type_pkg[python] (>= 1.6.2-1)
+    python%type_pkg[python],
+    ipython-py%type_pkg[python] (= %v-%r),
+    ipykernel-py%type_pkg[python],
+    jupyter-client-py%type_pkg[python] (>= 5.2.4-1),
+    jupyter-core-py%type_pkg[python] (>= 4.4.0-1),
+    matplotlib-py%type_pkg[python],
+    nbconvert-docs-py%type_pkg[python],
+    nbformat-py%type_pkg[python] (>= 4.4.0-1),
+    setuptools-tng-py%type_pkg[python],
+    sphinx-py%type_pkg[python] (>= 1.6.2-1)
 <<
 
 Source: https://files.pythonhosted.org/packages/source/i/ipython/ipython-%v.tar.gz
 Source-Checksum: SHA256(e840810029224b56cd0d9e7719dc3b39cf84d577f8ac686547c8ba7a06eeab26)
 
+PatchFile: ipython-docs-py.patch
+PatchFile-MD5: a760e05c555e921e8fd271f758097cb1
+PatchScript: sed 's|@PREFIX@|%p|g;s|@PY_PKG@|%type_pkg[python]|g;s|@PY_RAW@|%type_raw[python]|g' < %{PatchFile} | patch -p1
+
 CompileScript: <<
- make -C docs LANG=en_US.UTF-8 PYTHON=%p/bin/python%type_raw[python] SPHINXBUILD=%p/bin/sphinx-build%type_raw[python] html
+    make -C docs LANG=en_US.UTF-8 PYTHON=%p/bin/python%type_raw[python] SPHINXBUILD=%p/bin/sphinx-build%type_raw[python] html
 <<
 
 InstallScript: <<
- mkdir -p %i/share/doc/ipython-py%type_pkg[python]
- mv docs/build/html %i/share/doc/ipython-py%type_pkg[python]
+    mkdir -p %i/share/doc/ipython-py%type_pkg[python]
+    mv docs/build/html %i/share/doc/ipython-py%type_pkg[python]
 <<
 
 Description: IPython HTML documentation
--- a/10.9-libcxx/stable/main/finkinfo/languages/ipython-docs-py.patch
+++ b/10.9-libcxx/stable/main/finkinfo/languages/ipython-docs-py.patch
@@ -1,0 +1,27 @@
+diff --git a/docs/source/conf.py b/docs/source/conf.py
+--- a/docs/source/conf.py       2019-05-20 23:53:07.000000000 +0200
++++ b/docs/source/conf.py       2019-05-30 01:55:00.000000000 +0200
+@@ -222,18 +222,11 @@
+ # Output file base name for HTML help builder.
+ htmlhelp_basename = 'ipythondoc'
+ 
+-intersphinx_mapping = {'python': ('https://docs.python.org/3/', None),
+-                       'rpy2': ('https://rpy2.readthedocs.io/en/version_2.8.x/', None),
+-                       'jupyterclient': ('https://jupyter-client.readthedocs.io/en/latest/', None),
+-                       'ipyparallel': ('https://ipyparallel.readthedocs.io/en/latest/', None),
+-                       'jupyter': ('https://jupyter.readthedocs.io/en/latest/', None),
+-                       'jedi': ('https://jedi.readthedocs.io/en/latest/', None),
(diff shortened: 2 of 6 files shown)


dhomeier:
- [] nbsphinx-py (new)
- [] ipywidgets-py (new)
- [] jupyter-core-py (%v, 37)
- [] jupyter-client-py (%v, 37)
- [] ipython-py (%v, 37)
- [] ipykernel-py (new)
- [] notebook-py (new)

nieder:
Missing jupyter-core-py.patch

dhomeier:
That's odd, it was committed in [1211da6](https://github.com/fink/fink-distributions/pull/413/commits/1211da63d382253125b187904f3f4c222a1ea960#diff-5acb029e655d6dcd5d98f720fcceaf47)...

nieder:
I just can't read.

nieder:
Your new nbsphinx-py37 needs nbconvert-py37 and nbformat-py37. But nbconvert-py.info and nbformat-py.info in stable don't have py37 variants (nor the older jupyter-core-py.info %v4.0.6 in stable).

nbconvert-py.info and nbformat-py.info at the least need to be adjusted for py37. If edited in stable tree, then the existing jupyter-core and jupyter-client also need to have py37 added. Otherwise, the py37 change to nbconvert and nbformat should be added here (and version bumped while at it?)

nieder:
Can the ipython docs be built without building the rest of the package? It's already a separate command, but do they need the built ipython mod in tree or is the installed ipython tree good enough?

dhomeier:
Uh, really should check my `git add`s a fourth or fifth time, if 3x isn't enough...
All the version-specific command versions are now renamed `jupyter-*-py%type_pkg[python]` with these commits.

dhomeier:
Re: ipython docs: yes, it's possible to build them on the installed ipython instead.
So your suggestion is to create a separate package `ipython-docs-py` from the same sources, which `BuildDepends` on `ipython-py%type_pkg[python] (== %v)`, `ipykernel-py%type_pkg[python]`, `[nb]sphinx-py%type_pkg[python]`?
I had only been thinking about making it a `SplitOff` somehow, but that might be the cleanest solution indeed.

nieder:
Yes, please make `ipython-docs-pyNN` as an entire separate package (from same sources), which can then BDep (or Depends if there's no point in having the docs w/out the runtimes) on `ipython-py%type_pkg[python] (= %v)` and whatever other packages it needs to build. Extra bonus if doing this can clear some of the (Build)Dependencies from `ipython-py` itself that were only needed to create its docs.

dhomeier:
Doc requirements should only be in `BuildDepends`, unless I had made some other silly mistake, so i does not look like there is much to get rid of. Whether it makes sense to have the docs installed without the actual software is a matter of taste, I guess. In principle the docs could be made `%type_pkg[python]`-independent, as there are virtually no differences apart from some version strings and date stamps. But then it would become tricky to decide which `ipython-py` to BDep on...

dhomeier:
I've removed the old `ipython-py` packages plus `ipykernel-py` and `sphinx-py` to test the `ipython-py` rebuild, and then the `ipython-docs-py` build. I noticed that Sphinx is downloading a number of files during build:
```
Running Sphinx v1.8.4
Adding Tag: ipystable
Initializing GitHub plugin
loading pickled environment... done
loading intersphinx inventory from https://docs.python.org/3/objects.inv...
loading intersphinx inventory from https://rpy2.readthedocs.io/en/version_2.8.x/objects.inv...
loading intersphinx inventory from https://jupyter-client.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://ipyparallel.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://jupyter.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://jedi.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://traitlets.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://ipykernel.readthedocs.io/en/latest/objects.inv...
loading intersphinx inventory from https://python-prompt-toolkit.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://ipywidgets.readthedocs.io/en/stable/objects.inv...
loading intersphinx inventory from https://pip.pypa.io/en/stable/objects.inv...
```
-- this can be avoided by removing the `intersphinx_mapping` dictionary from `docs/source/conf.py`, which still lets the `html` build succeed, but breaks or removes a couple of external links.
It might be possible to point the intersphinx entries to local paths by modifying `intersphinx_mapping` for a controlled download as `SourceN`, but from their paths I doubt that the files are immutable, so their md5sums would probably change frequently...

nieder:
ipython-py allows either qt4-mac or qt5-mac, but sets PYTHONPATH to qt4-x11/qt4-mac. Is that the right

Does this patch help with the downloading problems in ipython-docs? https://salsa.debian.org/python-team/modules/ipython4/blob/master/debian/patches/0002-Update-intersphinx-links-for-local-access.patch

nieder:
Building ipython-docs needs matplotlib:
````
reading sources... [ 84%] sphinxext                                                             

>>>-------------------------------------------------------------------------
Exception in /sw/build.build/ipython-docs-py35-7.5.0-1/ipython-7.5.0/docs/source/sphinxext.rst at block ending on line None
Specify :okexcept: as an option in the ipython:: block to suppress this message
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-16-b431797473b4> in <module>
----> 1 from matplotlib.pylab import *

ImportError: No module named 'matplotlib'
<<<-------------------------------------------------------------------------
````

**Comment by nieder** [comment]
Building ipykernel-py27 needs ipython-py27 as a BuildDep (to match the !py27 variants?):
```
ipykernel-4.10.0/setup.py
sudo -u fink-bld [ENV] sh -c /tmp/fink.aFtOo
python2.7 setup.py build
  File "setup.py", line 91, in <module>
    from ipykernel.kernelspec import write_kernel_spec, make_ipkernel_cmd, KERNEL_NAME
  File "/sw/build.build/ipykernel-py27-4.10.0-1/ipykernel-4.10.0/ipykernel/__init__.py", line 2, in <module>
    from .connect import *
  File "/sw/build.build/ipykernel-py27-4.10.0-1/ipykernel-4.10.0/ipykernel/connect.py", line 13, in <module>
    from IPython.core.profiledir import ProfileDir
ImportError: No module named IPython.core.profiledir
```

**Comment by dhomeier** [comment]
Thanks for the pointer to the intersphinx patch, @nieder - 

1. having the `objects.inv` directories available locally also requires Sphinx builds of those packages' docs

2. the style
```
'jupyterclient': ('https://jupyter-client.readthedocs.io/en/latest/', '@PREFIX@/share/doc/python-jupyter-client-py@PY_PKG@/html/objects.inv'),
```
will use the local `objects.inv` information, so no download required, to create links into the online docs. Nothing wrong with that per se, but the online docs _might_ change while the installed package remains the same. Since the local docs have to be built anyway, one could alternatively use the syntax
```
 'jupyterclient': ('@PREFIX@/share/doc/python-jupyter-client-py@PY_PKG@/html/', None),
```
to create purely local links, which should remain current until the package is updated. Although in principle, to ensure absolute synchronicity, I guess the docs would have to be rebuilt whenever any of the dependencies with their docs are updated (e.g. after an update to `jupiter-client-py` `ipython-docs-py` as well as any other package docs that link to the former's docs should be rebuilt) - I have no idea how to ensure this. But we are taking about a very small number of inter-module links in the first place, and even in the probably rare event that their path would change after an update, the worst that could result is a broken link in the HTML docs.

3. Of course this opens up an entire new dimension of circular dependencies, but I've now got the `jupyter-*` and `nb*` docs to build and set the intersphinx directories up so that only links "downward" in the dependency hierarchy are included (i.e. `ipython-docs` will include links to `nbformat-docs` etc., but not the opposite direction).


Looks like the `%p/lib/qt4-*` were left over from when there were `-mac` and `-x11` versions of the package with different paths; since both `pyqt4-mac` and `pyqt5-mac` are now installing to the default `site-packages` in fact no amendment to `$PYTHONPATH` is needed at all.

Might take a few days to push the updates, since I've only kept them on my primary computer's disk, which is currently in for repair...

dhomeier:
The `matplotlib-py37` `TestDepends` will require an update from the forthcoming mpl PR; actually I have not tried to test against the old `1.5.3` versions.
@schwehr I have taken over maintainership of the more extensively modified `nb*` and `jupyter` packages, if that's OK with you.

PR #416, opened 2019-05-17
Gettext 0.20.1

nieder:
update base packages to use gettext-0.20.1.


dhomeier:
Built with `-m` on 10.12, but only after tweaking the download paths -  I was unable to fetch `gettext-0.20.1.tar.gz` from any of the Sourceforge mirrors, and `libiconv` is listed as no longer hosted on SF.
```
Source2: mirror:gnu:gettext/gettext-0.20.1.tar.gz
Source3: mirror:gnu:gperf/gperf-3.0.4.tar.gz
```
did work after trying a few mirror sites.
Direct `fink -m update dpkg` did not work due to broken dependencies; had to update `libgettext2-dev` first.
Building `gettext` and 'dpkg' produced thousands of odd warnings à la
```
Warning: Failed to set locale category LC_NUMERIC to en_DE.
```
although my language environment is `en_GB`, however some things on the root level of my machine are still speaking German (e.g. the many compiler messages in fink builds), and I've never figured out how to change it even with `nvram` settings...

Otherwise `dpkg` builds and appears to work fine.

nieder:
gettext-0.20.1.tar.gz wasn't found in the SF downloads because I forgot to upload it. It's there now.

All bootstrap tarballs are available from Fink's SF space: https://sourceforge.net/projects/fink/files/miscellaneous/bootstrap/

Most packages use custom mirrors to check both upstream and our own SF space, but dpkg and libiconv can't do that for the gettext part because there can only be 1 custom mirror in a .info.

dhomeier:
Would something like
```
CustomMirror: <<
        Primary: https://ftpmirror.gnu.org/
<<
Source: mirror:custom:%n/libiconv-%v.tar.gz
Source-MD5: ace8b5f2db42f7b3b3057585e80d9808
Source2: mirror:custom:gettext/gettext-0.20.1.tar.gz
Source2-MD5: bb5b0c0caa028105f3ca1905ddc306e2
Source3: mirror:custom:gperf/gperf-3.0.4.tar.gz
```
not work for the `CustomMirror`?

After upgrading to `libgettext8-shlibs_0.20.1-1` I am now getting the
```
Warning: Failed to set locale category LC_NUMERIC to en_DE.
Warning: Failed to set locale category LC_TIME to en_DE.
Warning: Failed to set locale category LC_COLLATE to en_DE.
Warning: Failed to set locale category LC_MONETARY to en_DE.
Warning: Failed to set locale category LC_MESSAGES to en_DE.
Warning: Failed to set locale category LC_NUMERIC to en_DE.
Warning: Failed to set locale category LC_TIME to en_DE.
```
by the dozens in every fink build - not sure how nonstandard my installation is, but this could become a bit of a nuisance for following build logs...

nieder:
Because of the finkmirror problems, base packages are checking Fink's SF space alongside upstream (in case upstream has removed an old tarball). Just using `mirror:gnu` will not fix that.

By the time the bootstrap gets to libiconv and dpkg, gettext-0.20.1.tar.gz has been downloaded for libgettext8 and only gperf tarball remains. And if SF space goes down, then we have bigger problems.

For the locale warnings, that's something to file a bug with upstream.

nieder:
Because of the finkmirror problems, base packages are checking Fink's SF space alongside upstream (in case upstream has removed an old tarball). Just using `mirror:gnu` will not fix that.

By the time the bootstrap gets to libiconv and dpkg, gettext-0.20.1.tar.gz has been downloaded for libgettext8 and only gperf tarball remains. And if SF space goes down, then we have bigger problems.

For the locale warnings, that's something to file a bug with upstream.

dhomeier:
At least on my first build attempt, certainly all of the SF mirrors seemed to have failed. But indeed the `mirror:custom:othersourcedir` will probably not work for both `Primary` and `Secondary`, so I guess patience is simply required.

Finally teaching my MB proper English with 
```
sudo defaults write /Library/Preferences/.GlobalPreferences.plist AppleLocale en_GB
```
seems to have gotten me rid of the locale warnings (the previous setting `en_DE` indeed was not a supported combination).

remkos:
Closing this after merge.

PR #418, opened 2019-05-23
netcdf-c15-4.7.0-1: Upstream update

remkos:
Upstream update to version 4.7.0.
Also updated license, as changed upstream.
Removed unnecessary BuildDepends: doxygen
Moved CPPFLAGS and LDFLAGS from ConfigureParams to SetCPPFLAGS and SetLDFLAGS.

Successfully built and tested with "fink build -m" on Mac OS 10.14.5 with Command Line Tools 10.2.1

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/netcdf-c15.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/netcdf-c15.info
@@ -1,12 +1,11 @@
 Package: netcdf-c15
-Version: 4.6.3
+Version: 4.7.0
 Revision: 1
 BuildDependsOnly: true
 Maintainer: Remko Scharroo <remkos@users.sourceforge.net>
 
 Depends: %n-shlibs (= %v-%r)
 BuildDepends: <<
-	doxygen,
 	fink-package-precedence,
 	hdf5.100.v1.10,
 	hdf5-cpp.100.v1.10,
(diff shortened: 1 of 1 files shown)


nieder:
If you set the BuildDepends for hdf5.100.v1.10 to be >= 1.10.0-4, then you can get rid of the extra BDep on hdf5-cpp.100.v1.10, because that version moved the stray header to the right SplitOff.

nieder:
But this is otherwise fine as is. The above is just a dep cleanup to simplify things.

remkos:
Thanks for the suggestions @nieder. I'll slate that to include with the next upstream update.

dhomeier:
Will it help to avoid situations where even `apt-get` can't rescue you, like with this update of packages having both `netcdf-c15` and an older version as [Build]Depends?
```
While trying to install:
  hdf5-cpp.100.v1.10-1.10.0-4
  hdf5.100.v1.10-1.10.0-4

The following inconsistencies found:
  hdf5-cpp11 conflicts with hdf5-cpp.100.v1.10, but hdf5-cpp.100.v1.10 is installed

Trying to resolve dependencies...
Could not resolve inconsistent dependencies!

Fink isn't sure how to install the above packages safely. You may be able to fix things by running:

  fink scanpackages
  sudo apt-get update
  sudo apt-get install hdf5-cpp.100.v1.10=1.10.0-4 hdf5.100.v1.10=1.10.0-4

Failed: Fink::SysState: Could not resolve inconsistent dependencies
> fink scanpackages                                                                                                                                  
Updating the list of locally available binary packages.
Scanning dists/stable/main/binary-darwin-x86_64
New package: dists/stable/main/binary-darwin-x86_64/base/fink-mirrors_0.39.4.1-1_darwin-x86_64.deb
New package: dists/stable/main/binary-darwin-x86_64/base/fink-mirrors_0.41.0.1-1_darwin-x86_64.deb
> sudo apt-get update                                                                                                                                
Ign file: local/injected Release                       
Hit http://bindist.finkmirrors.net stable/main Packages
Hit http://bindist.finkmirrors.net stable/main Release
Reading Package Lists... Done              
Building Dependency Tree... Done
> sudo apt-get install hdf5-cpp.100.v1.10=1.10.0-4 hdf5.100.v1.10=1.10.0-4
Reading Package Lists... Done
Building Dependency Tree... Done
E: Version '1.10.0-4' for 'hdf5-cpp.100.v1.10' was not found
> fink list hdf5-cpp.100.v1.10                                                                                                                       
Information about 13570 packages read in 2 seconds.
     hdf5-cpp.100.v1.10               1.10.0-4                 HDF5 C++ headers
 i   hdf5-cpp.100.v1.10-shlibs        1.10.0-4                 HDF5 C++ libraries
> fink list hdf5-cpp11                                                                                                                               
Information about 13570 packages read in 2 seconds.
 i   hdf5-cpp11                       1.8.16-6                 HDF5 (v1.8 API)--C++ headers
 i   hdf5-cpp11-shlibs                1.8.16-6                 HDF5 (v1.8 API)--C++ libraries
```
Eventually got out of that by manually uninstalling `hdf5-cpp11` and then building `netcdf-c7` at a later time, but the solution was not really that obvious.

nieder:
Unfortunately, two related BuildDependsOnly packages frequently confuse the dependency engines. I've seen this sometimes happen with the libmagick*-dev packages when switching between versions. The only consistent solution I've found is to manually remove any of the -dev packages that belong together and are conflicting with what you want and then install their replacements. The dependency engine I think tries to remove liba1-dev and then replace it with liba2-dev, but libb1-dev blocks it somehow. So it should have first also removed libb1-dev, but it's not smart enough to know that.

PR #422, opened 2019-05-29
nco-4.8.0-1: Upstream update

remkos:
Also needed to update PERL5LIB env variable in order to find Locale::Messages perl module.


remkos:
I am cancelling this PR as the version is outdated already.

PR #423, opened 2019-06-10
pkg updates for use in 10.14.5 distribution

nieder:
Updates to some lower level packages that need systemperl-5184.
Also includes perl5184 for distributions << 10.14.5 so they can use the new pm5184 perlmod variants.

This needs PR fink/fink#190 to work.


dmacks:
Are Apple's 5.18.2 and 5.18.4 binary compatible? A few packages have .bundle that dyld link to libperl.dylib. Apple's pathname is only 5.18 not 5.18.{2 vs 4} so compiling against 5.18.2 and then upgrading to 5.18.4 will still have the .dylib present. But it will only work (rather than requiring a separate %r) if the .dylib is compatible.

dmacks:
Adding a new perl-version requires updating the install.sh of base-files to create a base perllocal.pod for that version, or else some -pmXXX can fail during PostInst or PostRm.

dmacks:
I think fink's perl5182, which is now being revived for 10.14.5, contains M_D_P=10.3, which gets hardcoded into various of its .deb files that are then used when some -pm5182 are built, which then might supply tools to build other -pm5182. That means it propagates a binary difference through several layers of dependencies. And the result is that some of those packages might not even build because xcode on 10.14ish no longer has sufficient support for building to a 10.3 target (see Bug #339).

nieder:
> I think fink's perl5182, which is now being revived for 10.14.5, contains M_D_P=10.3, 

perl5182 for 10.14.5+ is patched to not hardcode M_D_P=10.3. See commit 67d0f5a :: PatchFile2. perl5182 for 10.9 is untouched. Perl5184 is patched for all relevant dists.

nieder:
> Adding a new perl-version requires updating the install.sh of base-files to create a base perllocal.pod for that version, or else some -pmXXX can fail during PostInst or PostRm.

fink/base-files#2

nieder:
> Are Apple's 5.18.2 and 5.18.4 binary compatible? A few packages have .bundle that dyld link to libperl.dylib. Apple's pathname is only 5.18 not 5.18.{2 vs 4} so compiling against 5.18.2 and then upgrading to 5.18.4 will still have the .dylib present. But it will mean that a -pm5182 package would now be loaded by fink's perl5182 but runtime-linking against apple's perl5184. That might work, but only has any possible chance if the .dylib is compatible.

I believe they are binary compatible. Looking at the output of `nm -o` between 10.13.6/5.18.2 and 10.15b3/5.18.4, the only difference between 5.18.2 and 5.18.4 is in the address location (bolded) `libperl.dylib: *00000000000e4389* T _XS_version_vcmp`. The symbols list is identical except for gaining ____chkstk_darwin and ___cxa_atexit on 5.18.4, both of which are marked U=undefined.

remkos:
I leave it to @nieder to merge.

nieder:
@dmacks base-files-1.9.18 has been released that knows about perl-5.18.4

PR #430, opened 2019-07-12
New upstream versions; no significant dependency changes

babayoshihiko:
The CRAN packages are updated.  For the packages of which R version requirement has changed, new info files are created for older R.  In some packages, --no-manual has been added.  Tested with r-base31 through r-base35 on two Sierra machine.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-boot-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-boot-r.info
@@ -12,14 +12,14 @@ Distribution: <<
 	(%type_pkg[rversion] = 32 ) 10.12
 <<
 Type: rversion (3.5 3.4 3.3 3.2 3.1)
-Version: 1.3-20
+Version: 1.3-23
 Revision: 1
 Description: GNU R Tools for bootstrap methods
 Homepage: https://cran.r-project.org/package=boot
 License: GPL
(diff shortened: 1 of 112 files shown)


nieder:
Looks fine (but untested). Do the --no-manual changes mean that those rmods no longer need the texlive BuildDepends?

babayoshihiko:
--no-manual may be added to all the packages, as PDF manuals are not installed anyway.

nieder:
Ugh. I missed something.
@babayoshihiko 
cran-proj-rXX has a downgrade in the version from 1.0-9 to 1.0-8. What is that about?

babayoshihiko:
> Ugh. I missed something.
> @babayoshihiko
> cran-proj-rXX has a downgrade in the version from 1.0-9 to 1.0-8. What is that about?

I checked the version at CRAN site, which holds the older version than forge:
https://cran.r-project.org/web/packages/proj4/index.html

I should have been more careful.  I will update to 1.0-9 after I test it.

nieder:
Got it. When your bump to 1.0-9 after testing, bump the revision to 100 just to make sure we clear the previous version we had.

babayoshihiko:
> Looks fine (but untested). Do the --no-manual changes mean that those rmods no longer need the texlive BuildDepends?

I've just found that adding "mkdir tmp" solves the issue.
However, because the manual PDF is not installed by cran-*-r* packages, I just add --no-manual.

PR #431, opened 2019-07-12
nco-4.8.1-1: upstream update

remkos:
This is the result of an upstream update from version 4.7.9 to 4.8.1.

The addition of the `export PERL5LIB` is necessary as the package looks for `Locale/Messages.pm`, which is only available in the directory `%p/lib/perl5/5.18.2`

If there is a way to avoid hardcoding that path, please let me know.

Successfully built with `fink -m build` on Mac OS 10.14.5 with Command Line Tools 10.2

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/nco.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/nco.info
@@ -1,6 +1,6 @@
 Info2:<<
 Package: nco
-Version: 4.7.9
+Version: 4.8.1
 Revision: 1
 Description: The NetCDF Operators
 License: GPL3
@@ -96,8 +96,8 @@ Replaces: <<
 # Unpack Phase:
 Source: https://github.com/nco/nco/archive/%v.tar.gz
 SourceRename: %n-%v.tar.gz
-Source-MD5: 6ad3febf223f676ca9281c39c48fb9f3
-Source-Checksum: SHA1(a45fb9097078038842b64e3b85f5dd2f1fa2f275)
+Source-MD5: 547209b355cc20b87f6e093fbf810557
+Source-Checksum: SHA1(4ceab5676690a04058f71794541cc14800278eae)
 
 # Patch Phase:
 PatchScript:  <<
@@ -127,6 +127,7 @@ CompileScript: <<
 	export	NETCDF_INC=%p \
 		NETCDF_LIB=%p \
 		NETCDF4_ROOT=%p
+	export PERL5LIB=$PERL5LIB:%p/lib/perl5/5.18.2
 	export PATH=%p/opt/texinfo-legacy/bin:$PATH
 	%{default_script}
 	fink-package-precedence --prohibit-bdep=%N-dap,%N-opendap,%N-netcdf,%N .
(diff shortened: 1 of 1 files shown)


nieder:
If it's looking for Locale/Messages.pm, shouldn't there be a (Build)Depends on locale-textdomain-pm*? If you hardcode a specific perl (5.18.2), then it'll be using system-perl5182 on 10.10-10.14.4 and Fink-perl5182 on 10.9 and 10.14.5+.

nieder:
OK, I have some more clarity on the locale-textdomain-pm situation (this is all on 10.13 or 10.14.5, but should be universal).
nco BuildDepends on texinfo, which has an explict Depends on locale-textdomain-pm%type_pkg[systemperl]. texinfo has different .info files to match the different systemperl versions, so hardcoding %p/lib/perl5/5.18.2 in nco will make it not find locale-textdomain-pm5162 on 10.9 (or 5.18.4 in the upcoming dists). Standard Fink practice is to make separate .info files for different systemperl versions when hardcoding a systemperl version (because we don't have a %Distribution expansion token).

Regardless, the nco build is currently finding texinfo-legacy if it's installed because PATH is being set to `PATH=%p/opt/texinfo-legacy/bin:$PATH` (line 131).
```
checking for makeinfo... yes
MAKEINFO_VERSION=4.13
```
If I comment out setting the PATH in the .info or remove texinfo-legacy, then texinfo-6.5 is found AND the nco build finishes, even if I comment out the PERL5LIB export.

What exactly is the error message you see if you don't modify PERL5LIB?
Since you're on 10.14.5 are you using the released 10.14 or the dist10.14.5 tree that needs to be merged into fink-distributions.git?

remkos:
Thanks @nieder, for your inputs so far.

1. I'm working on 10.14.5 using the released 10.14 (which is a link to 10.9-libcxx) without any further modifications. I am not aware of anything that needs to merged in addition.
2. I do not have texinfo-legacy installed, and removing the PATH in line 131 makes no change.
3. The error message I get during the configuration, and it is repeated at the end and then fails the build is this:
<pre>
checking for makeinfo... yes
Can't locate Locale/Messages.pm in @INC (you may need to install the Locale::Messages module) (@INC contains: /sw/lib/texinfo /sw/share/texinfo/lib/Unicode-EastAsianWidth/lib /sw/share/texinfo /sw/share/texinfo /sw/lib/perl5/darwin-thread-multi-2level /sw/lib/perl5 /sw/lib/perl5/darwin /Library/Perl/5.18/darwin-thread-multi-2level /Library/Perl/5.18 /Network/Library/Perl/5.18/darwin-thread-multi-2level /Network/Library/Perl/5.18 /Library/Perl/Updates/5.18.4 /System/Library/Perl/5.18/darwin-thread-multi-2level /System/Library/Perl/5.18 /System/Library/Perl/Extras/5.18/darwin-thread-multi-2level /System/Library/Perl/Extras/5.18 .) at /sw/bin/makeinfo line 109.
BEGIN failed--compilation aborted at /sw/bin/makeinfo line 109.
MAKEINFO_VERSION=
configure: Documentation support enabled
</pre>

So I think there needs to be something done with to have `makeinfo` work like it does on your system. No matter what, on my system `%p/bin/makeinfo` looks on line 109 for `Locale/Messages`. Alternatively, I can just switch off documentation support.

remkos:
I looks like I'll need to wait until `dist10.14.5` from PR #423 has been merged.
The issue is clearly that `makeinfo` now fails on 10.14.5 because of the changes in perl. So this is best resolved after that.

Alternatively, @nieder, can I commit the update to `nco.info` (without the `export PATH` and `export PEL5LIB`) to `dist10.14.5`?

remkos:
After merging `dist10.4.5` it indeed appeared that PATH and PER5LIB are not needed to be set.
Removed in the latest commit.
Please re-review.

PR #441, opened 2019-08-13
ghostscript-9.27-1: upstream update

remkos:
This upstream update required newer versions of `libjbig2dec` and `freetype219` as well.

Built and tested with `fink -m build` on Mac OS 10.14.6 with Command Line Tools 10.3


nieder:
If I recall correctly, @dmacks had some concerns with macOS regressions on newer versions of Freetype, though I might be confusing that with another part of the freetype/fontconfig/xft/etc stack.

remkos:
@nieder: I invite @dmacks to please respond to your concern. If this updating freetype is truly a problem (and I can see the concern), than I suggest I make a separate freetype package (freetype224 in this case).
If @dmacks has no time to respond, or does not respond in a week or so, I suggest I go that way by default.

remkos:
As you can see, I have reverted the introduction of freetype219-2.10.1, and instead opted to just update to the same version as delivered with ghostscript, i.e. 2.9.1.

@nieder: there have been more changes to freetype219 in the past. In 2016 @dmacks made some updates. From his commit logs I found that including the Debian patches and patching one of the public structs to keep it the same is essential. I have also made those patches. So, I think we are now good with this.

Maybe @dmacks has something to add.

This PR is now ready for rereview.

nieder:
@htodd asked this on fink-devel:
```
Doesn't Ghostscript come with its own copy of freetype? Shouldn't you
just use that and not worry about the freetype package? 
```

remkos:
> @htodd asked this on fink-devel:
> 
> ```
> Doesn't Ghostscript come with its own copy of freetype? Shouldn't you
> just use that and not worry about the freetype package? 
> ```

That is a legitimate question. This comes down to what is the general consensus among the Fink developers: is it OK to not always enforce the use of shared libraries but use build-in ones?

In the case of `mupdf` from the same makers as `ghostscript` @dmacks actually wrote:
```
dmacks: Build system mixes included and externally supplied
        flavors of several libs, which gives an unreliable build and
        unpredictable result.
```
`Ghostscript` and `mupdf` both supply a lot of libraries with their code.

So I have a problem satisfying two requirements attributed to @dmacks:
1. "Do not mess with `freetype219` because of MacOS regressions"
2. "Do not use included versions of libs"

The included library `jbig2dec` is actually unique to `ghostscript` and `mupdf`. I do not see the need to have this as an external shared library. The whole library is under 100 kB, so the gains in doing so are minute.

If the consensus is to avoid troubles with `freetype219` by not updating, I suggest that I do the following:
1. Use the internal version of `freetype` in `ghostscript`
2. Use also the internal version of `jbig2dec` in `ghostscript`
3. Do the same for `mupdf` (and update to the latest version)
4. Kill the `jbig2dec` package (as it is only used by `ghostscript` and `mupdf`
5. Close this PR and start a new one for `ghostscript` and `mupdf` jointly.

Inviting responses from @nieder, @danielj7 and @htodd

htodd:
I'd do what you want. I used to maintain ghostscript until I finally was
pushed over the edge by the constant emails from jhowarth about how I
was doing it wrong. I don't want to be that guy.

I also used the fink libraries and not the internal, but jhowarth
pointed out that the ghostscript upstream did not want people to use
anything but the packaged libraries. That's why I suggested you use the
internal library for freetype (and update freetype for everything else.)

Also, aren't we suppossed to be discussing these things on fink-devel? I
gave up on updating things I didn't own because nieder wasn't receiving
(or reading) my github comments and I'm not sure how the email system
there works.

On Mon, 26 Aug 2019, Remko Scharroo wrote:

>
>       @htodd asked this on fink-devel:
> 
> Doesn't Ghostscript come with its own copy of freetype? Shouldn't you
> just use that and not worry about the freetype package? 
> 
> That is a legitimate question. This comes down to what is the general consensus among the Fink developers: is it OK to not
> always enforce the use of shared libraries but use build-in ones?
> 
> In the case of mupdf from the same makers as ghostscript @dmacks actually wrote:
> 
> dmacks: Build system mixes included and externally supplied
>         flavors of several libs, which gives an unreliable build and
>         unpredictable result.
> 
> Ghostscript and mupdf both supply a lot of libraries with their code.
> 
> So I have a problem satisfying two requirements attributed to @dmacks:
>
>  1. "Do not mess with freetype219 because of MacOS regressions"
>  2. "Do not use included versions of libs"
> 
> The included library jbig2dec is actually unique to ghostscript and mupdf. I do not see the need to have this as an external
> shared library. The whole library is under 100 kB, so the gains in doing so are minute.
> 
> If the consensus is to avoid troubles with freetype219 by not updating, I suggest that I do the following:
>
>  1. Use the internal version of freetype in ghostscript
>  2. Use also the internal version of jbig2dec in ghostscript
>  3. Do the same for mupdf (and update to the latest version)
>  4. Kill the jbig2dec package (as it is only used by ghostscript and mupdf
>  5. Close this PR and start a new one for ghostscript and mupdf jointly.
> 
> Inviting responses from @nieder, @danielj7 and @htodd
> 
> ?
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the
> thread.[AANLE45PD3EJHXBPG7ELPB3QGO56HA5CNFSM4ILK5KA2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOD5EDN4I.gif]
> 
> 
>

-- 
Hisashi T Fujinaka - htodd@twofifty.com
BSEE + BSChem + BAEnglish + MSCS + $2.50 = coffee

remkos:
@htodd: Thank you for shepherding the `ghostscript` package in the past. I also "do not want to be that guy", but I use `ghostscript` a lot in conjunction with `gmt5` and see bugs in `ghostscript` coming and going, so need to update regularly.

As for discussing on fink-devel: as you see this escalated from a simple update to something more complicated. I did not like to spread the discussion over two platforms. Github will normally send you an e-mail when you are mentioned in a comment, like this one. @nieder has been good in responding here.

I'll then go on and do what I suggested above and will include you as reviewer, and in future releases of the package as well. Thanks for being an active Fink participant!

remkos:
@dmacks: Thanks for your inputs. I too prefer to use shared libraries for all the good reasons you mentioned. Given your concern about breakage, in the latest commit for FT (c2b5710) I actually went back to 2.9.1 (in contrast to the earlier 2.10.1 that @nieder referred to above). I did this for two reasons:
- `ghostscript` only asks for version 2.9.1
- We have the patches for the debian package for this version, which I have included in ours.
(Of course, also the backward compatibility patch that you introduced a while ago is still included.)
Can you reflect on your concern to upgrade from 2.6.4, other than the patches mentioned above?

I have now run with this version for some time now, and have not encountered any issues with other packages that use `freetype219`, so I think I got it correct. 

By the way, `ghostscript` does not "want to use its own FT". It just provides all the dependencies inside its package for _easy_ compiling, even though we agree that is not _wise_.

If you agree that `freetype219-2.9.1-1` from commit c2b5710 works, then it would be sufficient if you approve this PR.

Thanks @nieder, @htadd, @dmacks for all your well-voiced inputs and concerns. I think we are conversing now.

remkos:
Ready for approval.

remkos:
I am closing the PR. I have just noticed that ghostscript version 9.27 has a significant bug in it that I am not happy about: https://bugs.ghostscript.com/show_bug.cgi?id=701473 

Waiting until 9.28 and I will incorporate the outcomes of this discussion then.

PR #449, opened 2019-08-29
New upstream versions based on bioconductor 3.6

babayoshihiko:
Simply updated the bioconductor modules to the latest versions of the upstream.


babayoshihiko:
> LGTM. There are ~10 other bioconductor mods that are still at 3.4. Will those be getting an upgrade as well?

At the moment, they have some building issues, mostly about dependency.  I will figure out the packages one by one.

PR #453, opened 2019-09-02
CRAN packages update 20190902

babayoshihiko:
New upstream versions.  

car and httr require R >=3.5 and >=3.2 respectively.  New files are added for older Rs.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-aer-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-aer-r.info
@@ -12,14 +12,14 @@ Distribution: <<
 	(%type_pkg[rversion] = 32 ) 10.12
 <<
 Type: rversion (3.5 3.4 3.3 3.2 3.1)
-Version: 1.2-6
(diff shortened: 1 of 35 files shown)


nieder:
For cran-coin-r, the Depends on cran-libcoin-r should be restored to `1.0-0-1`. Otherwise, fink thinks that the trailing `-0` refers to the package revision %r (and it'll look for %v=1.0, %r=0). I've had problems with rmods trying older packages when the Depended package includes a dash and the revision is not explicitly listed. While at it, might as well fix the lines for `cran-modeltools` and `cran-mvtnorm` in cran-libcoin-r.info that also list hyphenated %v (just add `-1` to the end).

We don't need to go through the entire rmod directory to fix every situation like this, but I've been trying to fix them as they appear.

nieder:
Didn't test all the packages, but they seem fine otherwise as straightforward updates.

babayoshihiko:
> For cran-coin-r, the Depends on cran-libcoin-r should be restored to `1.0-0-1`.

Fixed.

PR #454, opened 2019-09-04
gsl updated

babayoshihiko:
I created a new package gsl25 (and its variant gsl25-shlibs).  The new package is installed in /sw/opt because it has libgslcblas.0.dylib, which is also installed by gsl-shlibs.

Unfortunately, CRAN packages that depend on the new version of gsl are not buildable at the moment.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-gsl-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-gsl-r.info
@@ -12,14 +12,14 @@ Distribution: <<
 	(%type_pkg[rversion] = 32 ) 10.12
 <<
 Type: rversion (3.5 3.4 3.3 3.2 3.1)
-Version: 1.9-10.3
+Version: 2.1-6
 Revision: 1
 Description: Gnu Scientific Library R wrapper
 Homepage: http://cran.r-project.org/web/packages/gsl/index.html
 License: GPL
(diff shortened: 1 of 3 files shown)


nieder:
Bringing in @smaret since gsl is his.

What do you mean the CRAN packages are not buildable? The ones in this PR after updating to gsl25, or in general with the older gsl?

babayoshihiko:
> What do you mean the CRAN packages are not buildable? The ones in this PR after updating to gsl25, or in general with the older gsl?

The former.  The build log is as below:

checking for gsl-config... /sw/opt/gsl25/bin/gsl-config
checking if GSL version >= 2.1... checking for gcc... gcc
checking whether the C compiler works... no
configure: error: in `/private/tmp/RtmpqJYplW/Rbuildc37f217e13c4/gsl':
configure: error: C compiler cannot create executables
See `config.log' for more details
ERROR: configuration failed for package 'gsl'
* removing '/private/tmp/RtmpqJYplW/Rinstc37f5c43ba09/gsl'
      -----------------------------------
ERROR: package installation failed

nieder:
And of course R removes the build directory, so it's impossible to know why it failed :(

dmacks:
@neider pushed a new gsl suite via 13cfd42e797640667488299b814e9ac5570c1104. Let's close this, and can file bugs against the new package elsewhere.

nieder:
@babayoshihiko My apologies. I completely forgot about this PR when I created the updated libgsl25 package.

PR #455, opened 2019-09-04
ghostscript-9.27-2: Upstream update

remkos:
Second try at building a ghostscript-9.27 package. Closed earlier attempt #441. 

This upstream update required newer versions of `jbig2dec` and `freetype219`.
Now includes a patch (provided by the ghostscript developers) to fix a glaring bug in ghostscript version 9.27.

`jbig2dec` is only used by `mupdf` and will be updated forthwith.
`freetype219-2.9.1` is compatible with the previous version used in Fink (2.6.3) after necessary patches were applied.

Built and tested with `fink -m build` on Mac OS 10.14.6 with Command Line Tools 10.3


remkos:
Anybody interested to review this PR? It's related to #441

dmacks:
Regarding jbig2dec, looks like was 0.17 is tagged for release 2 days ago. Its BDep on fink>0.24 is no longer needed (inertia from now-many-years ago relevant to dists/OSX's/fink's long-EOLed; we should push a whole-dist cleanup of that some day). Ping @danielj7 for jbig2dec as its maintainer.

Regarding ghostscript, I think it would be nicer to have its new PatchFile2's filename begin with "ghostscript" (instead of, or prior to "ghostpdl") so that in the future others will recognize what this random .patch is for in our dist directory. Thanks for including a Desc note about it! Starting a new note on a separate line to minimize the diff would make easier to see what is/isn't changing in successive fink updates.

Regarding freetype219, please keep the literal "freetype" rather than switching to "%n" in a few places until we drop support for all OS X prior to Catalina. It minimizes diffs vs the freetype219-32bit package (no need to support/update that one unless someone wants it, but no need to make it more confusing to do so).

Also for freetype219, what happened to the backwards-compat patch? That patch was originally added because upstream made a change that apparently broke binary compatibility, which is what the patch restored. Did they roll back the change upstream, or somehow else determine that it actually isn't an ABI break?

remkos:
@dmacks: Thanks for your elaborate suggestions to make these packages better.

Regarding jbig2dec: with the above commit I removed the superfluous BDep on fink >= 0.24. I noticed also the tagging of 0.17. This is to be part of ghostscript 9.28 which is not in RC3. Note that jbig2dec always comes alongside a ghostscript release, so I'll hold off with 0.17 until ghostscript 9.28 is released. @danielj7 is aware of the changes to the package. He approved a previous (cancelled) PR already.

Regarding ghostscript: I originally kept the original name of the file. But, indeed, it is more practical to start with ghostscript rather than ghostpdl. So that's changed in the above commit as well. 

Regarding freetype219: I reverted to using freetype219 instead of %n as it was in the previous release. You are right that that is easier. The backward-compatibility patch is no longer in Debian either, not since Jan 2018 in their release of freetype 2.8.1-1. Note that this patch dates from 2004, and has since become obsolete (I presume). I defer here to the wisdom of Debian to do so. I added a Desc note to this affect. See above commit.

Ready for re-review.

remkos:
OK, I caved in and created jbig2dec-0.17.1 after all and committed once I was convinced it did not break ghostscript-9.27.

Ready for re-review.

remkos:
I'm abandoning this request as ghostscript v9.27 has been superseded by v9.50.

PR #457, opened 2019-09-09
Update some base packages

nieder:
Update libiconv and gzip to the latest upstream.
tar also has a newer 1.32 version, but 3 tests failed, so I didn't touch it.

Also updated some other base packages to use a CustomMirror so that it'll search both upstream and our SF space.


nieder:
Unzip has a bunch of CVE patches through Debian/Ubuntu. Should we apply those?

remkos:
Regarding `unzip`. I would suggest we are proactive, so if you have time, yes please implement those CVE's.

PR #458, opened 2019-09-10
netcdf-fortran8-4.5.1-1: new package needed due to upstream update

remkos:
This upstream version of netcdf-fortran is said to require netcdf-c >= 4.6.0, whereas for netcdf-fortran-4.4.5 it was >= 4.6.2. The latter was actually not reflected in the package info file, so that is updated here as well.
In addition, Conflicts/Replaces pairs are added between netcdf-fortran7 and netcdf-fortran8.

Note that this package should have been netcdf-fortran_7_ (it contains libnetcdff._7_.dylib) but the previous package was already _incorrectly_ called netcdf-fortran_7_.

Build and tested with `fink -m build` on Mac OS 10.14.6 with Command Line Tools 10.3.

Note: I purposely did not update netcdf-c to the current version 4.7.1 at this time as there is a significant bug in it that fails testing; thus awaiting 4.7.2. Also an upstream update of netcdf-cxx4 is expected soon.


remkos:
Thanks, @nieder, for the suggestion. I'll keep that in mind for the next bump.

PR #462, opened 2019-09-16
Removed excessive "<<" in libicu55-dev.info

remkos:
Parsing of libicu55-dev.info failed, because of one excessive "<<"


nieder:
It's actually missing the opening `<<`. Fixed by f809a16

remkos:
OK. I misinterpreted that, sorry. Best is that it is fixed now.
Now I have you "on the line", could you have a look at my PRs #455 and #458?
Thanks.

PR #463, opened 2019-09-18
Put up my packages for adoption.

bcully:
I am afraid the time has come. I'm no longer much of a mac user and don't have
the time or energy needed to properly maintain these packages anyway. Thanks to
the fink team for all you've done to make the mac a comfortable home for unix heads
like me!


remkos:
Thanks you @bcully for your enduring support of Fink. While I'm afraid I cannot adopt any of your packages, as I'm not an active user of them, I am confident that many users have used them and that someone will pick them up.
Of course, all updates are trivial, so I will accept them, but leave the PR open for a while longer to make others aware of your offer for adoption.
And: thank *you* for all the fish.

PR #471, opened 2019-10-01
Add r-base36

babayoshihiko:
R 3.6.1 has been added.  CRAN and Bioconductor packages will follow.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/r-base31.info
@@ -1,7 +1,7 @@
 Info2: <<
 Package: r-base31
 Version: 3.1.3
-Revision: 7
+Revision: 8
 Distribution: 10.9, 10.10, 10.11, 10.12
 Description: R Framework
 Type: rversion (3.1)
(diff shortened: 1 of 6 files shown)


nieder:
Should probably bump readline6 to readline7 unless it's known to fail.

Can we also bump the gcc5 BuildDepends to gcc8? That would probably need to be propagated across all the rmods that use gfortran to build, however.

babayoshihiko:
> Should probably bump readline6 to readline7 unless it's known to fail.
> 
> Can we also bump the gcc5 BuildDepends to gcc8? That would probably need to be propagated across all the rmods that use gfortran to build, however.

Can we stay with readline6 and gcc5 for R <=3.5?

For CRAN, we may use:
(%type_raw[rversion] < 3.6) gcc5-shlibs | (%type_raw[rversion] >= 3.6) gcc8-shlibs,
and so on.

nieder:
@babayoshihiko yes to keeping R <= 3.5 at the older readline/gcc.

nieder:
I recommend switching r-base36 to gcc9 (not gcc8 as I said earlier).  I just patched gcc9 to build on 10.15, and currently, it's the only gccX that builds on 10.15.

babayoshihiko:
> I recommend switching r-base36 to gcc9 (not gcc8 as I said earlier). I just patched gcc9 to build on 10.15, and currently, it's the only gccX that builds on 10.15.

Now r-base depends on gcc9 (and readline7).

I am building CRAN and bioconductor packages.  So far, there seems no problem.

PR #473, opened 2019-10-15
CRAN upstream update 20191016

babayoshihiko:
Regular updates.  Build test on Sierra for R 3.5 and R 3.1.  Dependency checked.  The packages that require a new CRAN package are excluded.


nieder:
The new file `cran-rgdal-1.4.-4.info`  has an extra period in the file name. Should be `cran-rgdal-1.4-4.info`

`cran-recipes-r.-0.1.5info` also needs to have the filename fixed.

babayoshihiko:
> The new file `cran-rgdal-1.4.-4.info` has an extra period in the file name. Should be `cran-rgdal-1.4-4.info`
> 
> `cran-recipes-r.-0.1.5info` also needs to have the filename fixed.

Fixed as:
cran-rgdal-r-1.4-4.info
cran-recipes-r-0.1.5.info

PR #474, opened 2019-10-18
CRAN new packages; update with new dependency

babayoshihiko:
Add new packages as some existing packages require them.
aws for fmri
linprog for gemetry
ranger for VIM
lifecycle for tidyr
Rtsne for labdsv

Please let me know if the license for Rtsne is appripriate (BSD) or not.
https://cran.r-project.org/web/packages/Rtsne/LICENSE

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-aws-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-aws-r.info
@@ -1,0 +1,58 @@
+Info2: <<
+
+Package: cran-aws-r%type_pkg[rversion]
+Type: rversion (3.5 3.4)
+Version: 2.2-1
+Revision: 1
+Description: Adaptive weights smoothing
+Homepage: https://cran.r-project.org/package=aws
+License: GPL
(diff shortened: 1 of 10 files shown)


nieder:
cran-fmri-r.info v1.9 now needs cran-aws-r, which only exists for 3.4/3.5, so cran-fmri-r needs to be revived as a separate info file for the older %v1.7-2 that will support 3.1-3.3.

The versioning at the end of the Shlibs fields in cran-aws-r.info, cran-ranger-r.info, and cran-rtsne-r.info doesn't look correct: `(>=02.2-1-1)`

babayoshihiko:
> cran-fmri-r.info v1.9 now needs cran-aws-r, which only exists for 3.4/3.5, so cran-fmri-r needs to be revived as a separate info file for the older %v1.7-2 that will support 3.1-3.3.
> 
> The versioning at the end of the Shlibs fields in cran-aws-r.info, cran-ranger-r.info, and cran-rtsne-r.info doesn't look correct: `(>=02.2-1-1)`

I think I've fixed them.

babayoshihiko:
Reverted:
(Build)Depends: r-baseXX(-dev)

Issue #472, opened 2019-10-10
Issue while reinstalling gpgme11 on 10.13 in /opt/sw/


x-ii-ii-x:
Good Day the Fink team and @dmalloc,

Even if not on 10.15 I'm reinstalling from scratch my packages in /opt/sw. Clean install.
While trying to reinstall claws-mail, gpgme11 failed compiling because of what looks like a permission issue.

Any hint on that please?

```
creating t-thread1
gcc -DHAVE_CONFIG_H -I. -I../.. -I../../src -I/opt/sw/include -I/opt/sw/include  -g -O2 -Wall -Wcast-align -Wshadow -Wstrict-prototypes -MT t-genkey.o -MD -MP -MF .deps/t-genkey.Tpo -c -o t-genkey.o t-genkey.c
mv -f .deps/t-genkey.Tpo .deps/t-genkey.Po
/bin/sh ../../libtool --tag=CC   --mode=link gcc  -g -O2 -Wall -Wcast-align -Wshadow -Wstrict-prototypes  -L/opt/sw/lib -o t-genkey t-genkey.o ../../src/libgpgme.la 
gcc -g -O2 -Wall -Wcast-align -Wshadow -Wstrict-prototypes -o .libs/t-genkey t-genkey.o  -L/opt/sw/lib ../../src/.libs/libgpgme.dylib /opt/sw/lib/libgpg-error.dylib
creating t-genkey
srcdir=. ./mkdemodirs 
-n Creating:
-n  Alpha
gpg: Attention : les droits du répertoire personnel « /opt/sw/src/fink.build/gpgme11-1.1.8-4/gpgme-1.1.8/tests/gpg »
            ne sont pas sûrs
gpg: Attention : rien n'a été exporté
gpg: Attention : les droits du répertoire personnel « /opt/sw/src/fink.build/gpgme11-1.1.8-4/gpgme-1.1.8/tests/gpg/Alpha »
            ne sont pas sûrs
gpg: aucune donnée OpenPGP valable n'a été trouvée.
make[3]: *** [Alpha/Secret.gpg] Error 2
make[2]: *** [all-recursive] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
### execution of /usr/bin/make failed, exit code 2
### execution of /tmp/fink.x7nH9 failed, exit code 2
```

Thank you very much in advance

nieder:
[...]

PR #476, opened 2019-10-19
resolve gnupg2 incompatibility

jptrainor:
Add --export option to gpg2 command in mkdemodirs test script to resolve install failure. This resolves issues #472.


nieder:
Nice. Can you add a note to the DescPackaging field as to what the patch does? Is the `--export` flag a newish requirement for gnupg2.2?

jptrainor:
My change is incorrect. The --export command I added is exporting the public key. The mkdemodirs script intends to export the secret key. It did not fail with my change so I incorrectly assumed all was well. I'm closing this and will comment further on the issue thread.

Issue #482, opened 2019-10-28
gcc6 migration tracker


dmacks:
As companion to #481, here are the reverse-(build)depends on gcc6 (according to dotty on my 10.13 box):
- [x] arpack-ng (maintainer: none; this package is the only rdep of lapack361)
- [x] ccpnmr-py27 (maintainer: none; can we add newer python variants?)
- [x] gracegtk (maintainer: @akhansen, also has dep on lapack350, which is tracked in #481 as a gcc5 user)
- [x] lapack361 (maintainer: @remkos)

This can all probably be killed in short order.

dmacks:
I'm going to work on our pile of lapackXXX libs

dmacks:
ccpnmr is not currently policy compliant: PostInst ```UpdateAuto.py``` attempts to download and install new versions of some of its runtime files from the upstream server. PostInst ```make links``` also creates additional symlinks for some of its runtime files at that point (triggers some debian system-integrity flags) rather than having them actually in the .deb

dmacks:
ccpnmr does not have an open license and does not even fully permit redistribution. Upstream has a new major version available that is directly downloadable ready-to-run (no compiling). Should we kill this package in fink?

nieder:
Probably a good idea. It's unmaintained, nothing else uses it, and bad if it's breaking policy like that.

dmacks:
Looks like we're all done here. Could someone with a clean checkout of master double-check?

nieder:
Nothing using it, already dist-restricted.

Issue #481, opened 2019-10-28
gcc5 migration tracker


dmacks:
We're up to gcc9 now, and apple keeps making changes that break things (see #359 and #470). Let's think about migrating packages that still use older gccX to later ones and reduce the amount of old stuff we have to support (or other packages that wind up unavailable as reverse-(build)depends of them).

1. It looks like r-baseXX propagates the gcc that was used to build it into a BDep for all rmods built to use it, so we'd need a flag-day where we rev-up everything all at once if we migrate any of r-base{31,32,32,34,35} to a newer gccX. However, with r-base36 about to land (#471) with gcc9, we could have *-r36 use that. And maybe mask older r-baseXX (and their *-rXX mods) out of the later OS X distros (there already are Distribution: tags but they are all obsolete and functionally no-ops).
2. Octave also has some gccX dependencies that might need to remain synchronized. But a bunch of *-octXXX have Type:gcc that aren't used at all so maybe that would be a sepate cleanup task.
3. Here are the other gcc5 reverse-(build)depends I see by a quick grep:
* Maintainer: Pierre-Henri Lavigne
- [x] adonthell
- [x] wastesedge
* Maintainer: Jack Fink
- [x] freedroidrpg
- [x] mathgl6
* Maintainer: Hanspeter Niederstrasser
- [x] gmic
- [x] xcrysden
* Maintainer: Daniel Macks
- [x] pdftk - currently needs gcj, which is gccX<=5 only. Masked to dist<<10.15
* Maintainer: W. G. Scott (#561)
- [x] clipper
- [x] coot
- [x] libccp4
- [x] phaser-mpi
- [x] refmac
- [x] tinker
- [x] usf
* Maintainer: Martin Costabel (#546 and #569)
- [x] ddscat
- [x] harminv
- [x] libctl
- [x] meep
- [x] melina
- [x] modulef
* Maintainer: Remi Mommsen (#496)
- [x] geant4.9
- [x] oot-pythia
- [x] root5
* Maintainer: Kurt Schwehr
- [x] hdf
- [x] sioseis
* Maintainer: Aurelien Chanudet 
- [x] mpb
* Maintainer: Sylvain Barbot
- [x] relax (#658)
* Maintainer: Alexander Hansen
- [x] libnc-dap3
- [x] qrupdate
- [x] xraylib3-gfortran
* Remkoo Scharroo
- [x] lapack350 (#483 and akh's qrupdate are the only rdep of this old lapack libversion)
[...]

PR #491, opened 2019-11-03
Gracegtk and qrupdate updates: gcc and lapack

dmacks:
gracegtk is one of the last remaining packages to use gcc6 (#482) and is the only user of lapack360, which is a gcc5 user (#481). Here's an update to gcc9 and to lapack380 (which will need to migrate to gcc9, but may as well only have to bring the latest lapackXXX moving forward). There's a newer version upstream, but it requires a newer gtk stack than fink has at the moment.


dmacks:
Also qrupdate (it and gracegtk are forks of same original)

PR #493, opened 2019-11-04
gmt-6.0.0-1: Introducing GMT 6.0.0

remkos:
This is a new branch of the GMT package, thus introducing a new package gmt6.
Built and tested with 'fink -m build' on MacOS 10.15.1 with Command Line Tools 11.0.

Also updated the Conflicts/Replaces lines of packages gmt and gmt5.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/gmt.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/gmt.info
@@ -1,6 +1,6 @@
 Package: gmt
 Version: 4.5.18
-Revision: 2
+Revision: 3
 Source: ftp://ftp.soest.hawaii.edu/gmt/gmt-%v-src.tar.bz2
 Source2: ftp://ftp.soest.hawaii.edu/gmt/gmt-%v-non-gpl-src.tar.bz2
 Source-MD5: b35cf18fddcf4823dc75b1ea32808d71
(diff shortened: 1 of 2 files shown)


nieder:
Finds and uses /usr/lib/libcurl.4.dylib and system-curl headers even with libcurl4-dev installed. Will need to set CURL_INCLUDE_DIR and CURL_LIBRARY.

There are checks for ffmpeg, graphicsmagick, and a couple other utilities. Do these affect the buildtime results, or does their presence only do anything at runtime?

remkos:
Thanks, @nieder, for catching that. The above patch now sets
CURL_INCLUDE_DIR
CURL_LIBRARY
so that there's no more mix between curl from the system and Fink.

Also, explicitly added GS_ROOT to find Fink's ghostscript.

The presence of ffmpeg and graphicsmagick do not affect the buildtime results. When they are available at runtime they can be used for some capabilities of GMT, if they are not available at runtime, these capabilities will not be there. I do not want to explicitly require these tools, as they are "heavy" and used only in one case in GMT (to create animations or movies). The program gracefully mentions them missing if so. Hence, that should be it for the gmt6.info file.

nieder:
Looks good. Just wanted to confirm that Ffmpeg and GM were only runtime deps. Could add them to Suggests: field in the future if you want, but not needed.

Issue #489, opened 2019-11-01
gcc8 migration tracker


nieder:
Also need to migrate gcc8 packages to gcc9 if we want them to work on 10.15.

none
- [x] raster3d
- [x] scipy-py
- [x] tcoffee (uses systemperl; no current 10.14.5/10.15 package)
- [x] xfoil

@dhomeier (#529)
- [x] openblas

@mommsen (#496)
- [x] cernlib2006
- [x] patchy4

@nieder 
- [x] cd-hit
- [x] fasttree

@remkos 
- [x] cdo
- [x] eccodes
- [x] fftw3
- [x] grib-api-fortran
- [x] lapack380
- [x] openmpi
- [x] pgplot
- [x] netcdf-fortran7 (dist restricted to not carry forward to 10.15)
- [x] netcdf-fortran8
- [x] wip

@schwehr 
- [x] sioseis

remkos:
Feel free to update netcdf-fortran7 to use gcc9.
I've just handed in my packages (PR #507)

dmacks:
netcdf-fortran7 was easy to switch over, so I did it to help clean the tree.

dmacks:
sioseis throws errors with gcc9 (I'm looking into it, but others with fortran experience are welcome to help too!)

dhomeier:
> sioseis throws errors with gcc9 (I'm looking into it, but others with fortran experience are welcome to help too!)

Uh, `EQUIVALENCE`...!

@dmacks I managed to hack around the `IAND` exceptions with this patch (what kind of developer portal is this that does not allow `.patch` extensions?!?)
[sioseis.patch](https://github.com/fink/fink-distributions/files/3976018/sioseis.patch.txt)

Not 100 % confident it does the right thing™, but at least the tests on https://sioseis.ucsd.edu/installation.html look OK, and `mp32ieee.f`, which I am most concerned about, only seems to be relevant for the `dutil` aux program - that _appears_ to basically function as well...

nieder:
So gcc8.4.0 was just released. It builds under 10.14 and 10.15. `sioseis` is the last gcc8 dependee.

nieder:
Is it worth making gcc8.4.0 available to 10.15 to save a single package?

nieder:
18ca51d1507f64836250ffa50d64036150e72dcf updates sioseis to latest that now builds cleanly with gcc9. That's the end of the gcc8 users.

Issue #484, opened 2019-10-29
gcc7 migration tracker


dmacks:
As companion to #481 and #482, here are the reverse-(build)depends on gcc7 (according to dotty on my 10.13 box):
* @kamischi
- [x] floppy
- [x] flow
- [x] freehelix
- [x] ftidy
- [x] ifeffit
- [x] snoopy
- [x] to-f90
* @remkos
- [x] lapack371 (no reverse-depends)
* Jack Fink
- [x] mopac7
- [x] ortep3

kamischi:
I'll have a go at my packages. Should be pretty forward. Same for ortep3.

dmacks:
I'm going to work on our pile of lapackXXX libs

dmacks:
Looks like we're all done here. Could someone with a clean checkout of master double-check, and then Distribution-restrict gcc7 itself?

nieder:
gcc{5-8} are already dist restricted by needing special .info files for 10.14 and 10.14.5 (and gccX.info is 10.9-10.13). gcc9 is the only one with a single .info for all dists.

I don't see gcc7 being used anywhere anymore

Issue #483, opened 2019-10-28
Upgrade cpplapack to use newer lapackXXX libs


dmacks:
cpplapack-2015.05.11-1 has: ``Recommends: atlas | lapack350 | lapack342 | lapack341`` but our current lapackXXX versions are: lapack350, lapack360, lapack361, lapack371, lapack380, many of which have build problems and/or no other users in fink. It would be great to be able to kill off those old versions, or at least not require them for this package here. Can cpplapack use newer lapackXXX?

nieder:
@okayamat This is yours I think.

dmacks:
Looks like it might not be doable because it has template code around symbols that were deprecated and then removed in lapack>=360

dmacks:
Upstream has been dead for years, no other Mac package-manager or major linux distro seems to have it. Should we just kill it (or at lest dist-restrict to <= 10.14.5 so we don't have to keep ancient lapack350 forever?

okayamat:
It is ok for me to remove this package. Thank you for the notice.

dmacks:
Thanks.

PR #497, opened 2019-11-07
Lapack cleanup

dmacks:
We have a lot of lapackXXX, only two of which are still in use and many of which use old gccX. Here I switch all to gcc9 and kill off the headers packages for the unused versions. For the latter, the -shlibs are kept per policy. But I masked them out of the 10.15 where they were uncompilable due to (formerly) using a gccX that wasn't available there. And I inverted the .info layout so that -shlibs was the main package in all cases to make it easier to turn off the headers pkgs and minimize the diff among those files.


dmacks:
Oops, didn't realize another PR touched one of the files in a way that blocked the merge:)

dmacks:
lapack350 is still in use by a ton of octave packages...working on it...

dmacks:
Other tickets that are tracking lapack* packages: #481, #482, #483, #484, #489

remkos:
Feel free to do with lapack packages as you please. I have just handed them in (PR #507).

dmacks:
rebased on master (PR #507 changed Maintainer fields)

PR #502, opened 2019-11-10
libarchive (and bsdtar) upstream version 3.4.0

PovlAbrahamsen:
An update to the latest upstream version 3.4.0. Tested on 10.14.5 and 10.15.

The previous update was made by @nieder in 2016, noting a "maintainer timeout".

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libarchive31.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libarchive31.info
@@ -1,5 +1,5 @@
 Package: libarchive31
-Version: 3.2.1
+Version: 3.4.0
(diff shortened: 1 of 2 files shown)


dmacks:
I assume the changed line-numbers in the .patch are to match the new version, so it would be good to update the +++/--- lines there as well. Maybe apply the .patch to a clean source directory and regenerate it properly using 'diff'?

PovlAbrahamsen:
Yes, the line numbers were changed to match the updated source (the previous version dates back to 3.1.2, so older than the previous info file). I've regenerated from the latest source - even if this is just cosmetic. And updated the checksum in the info file.

PR #504, opened 2019-11-11
Rmods r36

babayoshihiko:
Added r36 variant to cran and bioconductor packages.  All of them were successfully built on Sierra.


dmacks:
I think a lot of these don't need to have the Revision increased if all that you are doing is adding/removing variants. It is necessary if there are changes to the dependencies (such as gcc* version), but otherwise it just forces users to rebuild for no obvious reason.

babayoshihiko:
Can I revert the revision numbers as follows?

Revision change is needed:
(%type_raw[rversion] << 3.6) gcc5-compiler | (%type_raw[rversion] >= 3.6) gcc9-compiler
Homepage (bioconductor-*) or Description (bioconductor-xcms) is changed.

Revision change is not needed:
Only distribution field is changed.
Conditional dependency is deleted because the variant does not meet the condition.
(e.g. removed ( %type_raw[rversion] = 3.2 ) for -r31 only packages)
new .info for older Rs (cran-amap-r-0.8-16.info, cran-analyzefmri-r-1.1-17.info, cran-bayesm-r-3.0-2.info, cran-tree)

babayoshihiko:
I think I have fixed all the revision issues.

babayoshihiko:
Can anyone review and possibly approve the request?  I am not updating other CRAN packages.

dmacks:
I found a few inconsistencies in the availability of certain variants. Quick-fix, pushed as 193fda8

PR #508, opened 2019-11-12
Fixed git-lfs build, bumped to v2.9.0

leighsmith:
Advanced git-lfs to v2.9.0, revamped installation method to use `git clone`, since the git command is used within Makefile and needs to find a .git directory to work, unpacking a tarball doesn't work.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/git-lfs.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/git-lfs.info
@@ -1,6 +1,6 @@
-# -*- coding: ascii; tab-width: 4 -*-
+# -*- coding: utf-8; tab-width: 4 -*-
 Package: git-lfs
-Version: 2.8.0
+Version: 2.9.0
 Revision: 1
 Description: Git Large File Storage extension
 DescDetail: <<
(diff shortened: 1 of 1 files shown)


nieder:
We can't allow network activity during *Script. If git-lfs does need an actual git directory, the best way is to locally clone the repository at a specific tag or commit, zip it up, and use that as `Source:` (we can host it at SF.net). Does Makefile do any git connectivity once it has a full git archive?

leighsmith:
I don't think it does. I'll do as you suggest. I'm very surprised they need to use git in the Makefile anyway.

danielj7:
I will take a look at this. I'm in the middle of making python38 at the moment but I'll do this next.

danielj7:
Ok, while the makefile does use the git command, it isn't necessary for the build to succeed. You'll get a bunch of `fatal: not a git repository (or any of the parent directories): .git` but it doesn't hurt anything. I've updated to 2.9.0 and it builds fine. Thanks for the note.

leighsmith:
It turns out there is another problem with `make` failing to build because the `lint` rule fails, due to the `@! ` prefix before `$(GO)` is causing gnumake v4.21 (installed by fink) to fail. I found removing this with:

```
@@ -435,7 +435,7 @@ endif
 # are vendored in via vendor (see: above).
 .PHONY : lint
 lint : $(SOURCES)
-	@! $(GO) list -f '{{ join .Deps "\n" }}' . \
+	$(GO) list -f '{{ join .Deps "\n" }}' . \
 	| $(XARGS) $(GO) list -f \
 		'{{ if and (not .Standard) (not .Module) }} \
 			{{ .ImportPath }} \
```
Fixed the problem. I can submit an updated pull request.

PR #511, opened 2019-11-14
Building git-lfs using fink gnumake 4.2.1

leighsmith:
Added patch to remove the use of git which was failing to retrieve the version number correctly, and `@!` which was causing gnumake v4.2.1 (as installed by fink) to fail to build.


dmacks (approved):
That fixed it for me. Push?

PR #516, opened 2019-11-21
Fix build.  No longer have to worry about build using /tmp/telnet.  A…

akhansen:
No longer have to worry about build using /tmp/telnet.  Add 10.14.5 to distributions.


akhansen:
Tested on 10.13, 10.14.5, and 10.15, so let's do this thing.

PR #518, opened 2019-11-24
json-c version 0.13.1 with new library version

wodev:
New package json-c4-shlibs for json-c .

json-c version 0.13 bumps a new library version. I created a new package json-c4-shlibs to keep alive the old library version .

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/json-c4-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/json-c4-shlibs.info
@@ -1,0 +1,62 @@
+### json-c 0.13 bumps to libN=5
+Package: json-c4-shlibs
+Version: 0.13.1
+Revision: 1
+License: BSD
+Maintainer: Wolfgang Fischer <wodev@users.sourceforge.net>
+Homepage: https://github.com/json-c/json-c/wiki
+###
+BuildDepends: <<
+        fink-package-precedence (>= 0.7-1)
+<<
+###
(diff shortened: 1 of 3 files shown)


10.9-libcxx/stable/main/finkinfo/libs/json-c4-shlibs.info
```diff
@@ -0,0 +1,62 @@
+### json-c 0.13 bumps to libN=5
```
nieder: It's just a comment, but Fink standard is that libN matches the library install name, not the compatibility version. So json-c4 is correct because the library name is libjson-c.4.dylib, and libN is also '4' when we comment about having to update (or not) the package name.

nieder:
Do we need to keep the old libjson.info around?
1) Nothing uses it and is just a static library,
2) hasn't been updated upstream since 2012,
3) the `Conflicts` in the json-c packages against libjson are confusing because they are not meant to be against our current libjson-7.6.1 package. The json-c package [used to be called libjson 7+ years ago](https://github.com/fink/fink-distributions/commit/d96c74992a372c9d75f1bc660dcd02b1ed846692#diff-a92d56a7a8ed5fd35e66302cdc6b13ec) but was renamed to json-c in order to match the library name. The `Conflicts` field was a leftover from that renaming. Current json-c* and libjson don't have any reason to Conflict (no file collision).

nieder:
Closing this because #699 added the newest libN=5 (v0.15). If there's a need for the intermediate libN=4 (v0.13), we can revisit.

PR #520, opened 2019-11-25
New package postgresql12 12.1

wodev:
new package postgresql12 (12.1) for latest postgresql major version 12.


nieder:
Thanks for keeping the postgresql packages going. Do we need to keep all the old versions around? Nothing is currently using postgresql95/10/11.

PR #524, opened 2019-11-29
Drop maintainership of packages

smaret:
I am no longer a Fink user, so I am marking the package I used to maintain as unmaintained.


smaret:
Thanks @dmacks !

PR #525, opened 2019-12-01
Sundials update

dmacks:
Switch to gcc9 (fixes part of  #481), and new version that is new libversion. It now uses cmake, so if @nieder or other wizard knows if there's some better way to have the library files go into a lib/ subdir, I'd love to hear it...


nieder:
Add
```
		-DCMAKE_INSTALL_NAME_DIR=%p/lib/sundials50 \
		-DCMAKE_INSTALL_LIBDIR=%p/lib/sundials50 \
```

dmacks:
Ahhh, yes. That works. Thanks!

dmacks:
Yuppers.

PR #528, opened 2019-12-05
ESS update to 18.10.2 + emacs26 compatibility

dhomeier:
The updated `emacs26` package (thanks!) is not in the dependency options of `ess`. This version adds the new emacsen and is updated to the latest upstream, with the emacsen-common install voodoo now pulled separately from Debian. Tested on 10.12.6.


dhomeier:
Agreed, I had to look up some of them myself.

dhomeier:
Thanks for the review!

PR #529, opened 2019-12-05
OpenBLAS migrated to gcc9 and update to upstream 0.3.7

dhomeier:
Part of the gcc8 migration #489. Added some fine tuning of the optimisation options and a few benchmark runs as part of the `TestScript`. Tested on 10.12.6.


dmacks:
After this gets into master, check it off on #489

dhomeier:
> In CompileScript, you always set -mtune=intel and then set different -march= depending on platform, and patches it into a Makefile. TestScript only sets -mtune (not -march) and does so via command-line variable passing. Can the TestScript also be done by patching into a Makefile? I'm mainly wondering how much of this could be moved into PatchScript.

Ideally it should not do any further compilation after the `Compile` phase; but this was apparently really left over from previous versions - everything just controlled from `Makefile.system` now.

And right now there are only branches for `darwin_vers` >= or < 18 anyway, but I have not tested anything but 16...

PR #531, opened 2019-12-08
libnc-dap3: gcc upgrade, resolve some warnings, etc

dmacks:
* -I ordering (put local before external/global
* resolve implicit-declaration warnings
* fix printf string security


dmacks:
<akh> Fine with me.
<akh> ok

PR #532, opened 2019-12-08
Xraylib misc fixes

dmacks:


dmacks:
On #fink, akh says "Go for it"

PR #539, opened 2019-12-19
fortune-mod: new upstream version

PovlAbrahamsen:
Changes:
- new version with new upstream repository
- uses recode3 instead of recode
- uses cmake
- uses "rinutils" headers (which have been added as a separate package in libs as part of this PR)

Tested on 10.15.2, including with "-m"

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/games/fortune-mod.info
+++ b/10.9-libcxx/stable/main/finkinfo/games/fortune-mod.info
@@ -39,32 +39,37 @@ PatchFile: %n.patch
 PatchFile-MD5: f5b82d985e48bb3a308af3615f6be6fb
 PatchScript: sed 's|@PREFIX@|%p|g' < %{PatchFile} | patch -p1
 CompileScript: <<
+	#!/bin/sh -ev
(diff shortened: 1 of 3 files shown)


nieder:
```
	cd build && cmake -D CMAKE_C_FLAGS=-I%p/include \
 		-D CMAKE_INSTALL_PREFIX=%i \
 		-D COOKIEDIR=%i/share/fortunes \
 		-D LOCALDIR=%i/share/fortunes ..
 	perl -pi -e 's|%i|%p|g' build/config.h
```
You call fink-buildenv-cmake.sh, but don't use $FINK_CMAKE_ARGS, which sets CMAKE_INSTALL_PREFIX=%p among other options. And if you set COOKEDIR and LOCALDIR to `%p/...`, then you can also get rid of the perl pie that changes %i back to %p.

nieder:
If you add `#!/bin/sh -ev` to the start of CompileScript, then $FINK_CMAKE_ARGS is useable to the cmake command (same shell instead of different processes) in CompileScript. Then also use `make install DESTDIR=%d`

PovlAbrahamsen:
I've made a few tweaks based on @nieder 's suggestions. Thank you!

PR #540, opened 2019-12-20
ENH: Python3 updates and test suite fixes

### LINKED ISSUE #530: python38 build fails due to wrong ncurses include path
Compilation of `python38_3.8.0-1` fails on
```
building '_curses_panel' extension
gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -fwrapv -O3 -Wall -std=c99 -Wextra -Wno-unused-result -Wno-unused-parameter -Wno-missing-field-initializers -Wstrict-prototypes -Werror=implicit-function-declaration -I./Include/internal -DHAVE_NCURSESW=1 -D_XOPEN_SOURCE_EXTENDED=1 -I/sw/ncursesw -I./Include -I. -I/sw/include -I/scratch.noindex/fink.build/python38-3.8.0-1/Python-3.8.0/Include -I/scratch.noindex/fink.build/python38-3.8.0-1/Python-3.8.0 -c /scratch.noindex/fink.build/python38-3.8.0-1/Python-3.8.0/Modules/_curses_panel.c -o build/temp.macosx-10.12-x86_64-3.8/scratch.noindex/fink.build/python38-3.8.0-1/Python-3.8.0/Modules/_curses_panel.o
/scratch.noindex/fink.build/python38-3.8.0-1/Python-3.8.0/Modules/_cursesmodule.c:621:9: error: implicit declaration of function 'setcchar' is
      invalid in C99 [-Werror,-Wimplicit-function-declaration]
        setcchar(&wcval, wstr, attr, attr_to_color_pair(attr), NULL);
```
with the include path in `setup.py` changed to `@PREFIX@/ncursesw`. Changing the patch to
```
+                curses_includes.append('@PREFIX@/include/ncursesw')
```
lets the `python38` build succeed with
```
 i   libncurses5                   5.9-20110507-1         Full-screen ascii drawing library
 p   libncurses5-dev                                      [virtual package]
 i   libncurses5-shlibs            5.9-20110507-1         Shared libraries for libncurses5 package
 i   libncursesw5                  5.9-20110507-1         Full-screen ascii drawing (wchar version)
 i   libncursesw5-shlibs           5.9-20110507-1         Shared libraries for libncursesw5 package
 i   ncurses                       5.9-20110507-1         Executable files for libncurses5
```

dhomeier:
When checking the newer Python 3.x packages, I noticed that the tests are all failing immediately because `install_name_tool` is called on a non-existent `python.exe`.
Changed that to the default executable name and updated to the latest upstream.
3.7.6 and 3.8.1 however are failing with one persistent failure in `test_startup_imports`.

Also fixed #530; tested on 10.12.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/python27.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python27.info
@@ -2,7 +2,7 @@
 Info2: <<
 Package: python%type_pkg[python]
 Version: 2.7.18
-Revision: 1
+Revision: 2
 Epoch: 1
 Type: python 2.7
 Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
--- a/10.9-libcxx/stable/main/finkinfo/languages/python27.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python27.info
@@ -103,17 +103,31 @@ CompileScript: <<
 InfoTest: <<
 	TestConflicts: pyxml-py%type_pkg[python], gnome-applets, apple-gdb
 	TestScript: <<
-		# Change install_name of lib during tests since DYLD_LIBRARY_PATH doesn't work on 10.11+.
-		install_name_tool -change %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib %b/libpython%type_raw[python].dylib python.exe
+		#!/bin/sh -ex
+		# Change install_name of lib during tests since DYLD_LIBRARY_PATH is not exported to test_subprocess on 10.11+.
+		if [ -x python.exe ]; then
+			python_exe=python.exe
+		else
+			python_exe=python
+		fi
+		install_name_tool -change %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib %b/libpython%type_raw[python].dylib $python_exe
 		/usr/bin/find `ls -d %b/build/lib*` -name *.so -exec install_name_tool -change %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib %b/libpython%type_raw[python].dylib {} \;
 		install_name_tool -id %b/libpython%type_raw[python].dylib libpython%type_raw[python].dylib
-		
-		LANG=en_US.UTF-8 make -k test EXTRATESTOPTS='-w -x test_distutils test_argparse test_httpservers test_gdb test_ssl test_httplib test_urllib2_localnet' || exit 2
-		
+
+		# test_socket can fail with some DNS settings.
+		# multiprocessing_spawn fails on "KeyboardInterrupt not raised by wait",
+		# the other 4 hang indefinitely
+		EXTRATESTOPTS='-w -x test_socket test_distutils test_ssl test_multiprocessing_spawn test_asyncio test_httpservers test_logging test_xmlrpc'
+		LANG=en_US.UTF-8 make -k test EXTRATESTOPTS="$EXTRATESTOPTS" || TESTRETURN=$?
+
 		# Put install_name back.
-		install_name_tool -change %b/libpython%type_raw[python].dylib %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib python.exe
+		install_name_tool -change %b/libpython%type_raw[python].dylib %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib $python_exe
 		/usr/bin/find `ls -d %b/build/lib*` -name *.so -exec install_name_tool -change %b/libpython%type_raw[python].dylib %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib {} \;
 		install_name_tool -id %p/lib/python%type_raw[python]/config/libpython%type_raw[python].dylib libpython%type_raw[python].dylib
+
+		# Remove stray .pyc that get made during tests to keep validator happy.
+		find ./Tools -name "*.pyc" -delete
+		exit $TESTRETURN
 	<<
 <<
 
--- a/10.9-libcxx/stable/main/finkinfo/languages/python27.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python27.info
@@ -269,6 +269,9 @@ DescPort: <<
 	
 	Upstream patch from https://bugs.python.org/issue27369 to fix tests
 	with expat 2.2.0.
+
+	Call build executable by correct filename in test suite
+	(python on case-sensitive, python.exe on -insensitive file systems).
 <<
 License: OSI-Approved
 Homepage: http://www.python.org
--- a/10.9-libcxx/stable/main/finkinfo/languages/python27.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python27.info
@@ -18,7 +18,7 @@ Depends: <<
 	libgettext8-shlibs,
 	libncursesw5-shlibs,
 	openssl110-shlibs (>= 1.1.1d-1),
-	readline6-shlibs,
+	readline7-shlibs,
 	sqlite3-shlibs  (>= 3.23.1-1),
 	tcltk (>= 8.4.1-1),
 	x11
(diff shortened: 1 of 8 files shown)


nieder:
```
install_name_tool -change /sw/lib/libpython3.8.dylib /sw/build.build/python38-3.8.1-1/Python-3.8.1/libpython3.8.dylib python
error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't map file: python (Invalid argument)
### execution of install_name_tool failed, exit code 1
### execution of /tmp/fink._GvSP failed, exit code 1
```
It needs the `.exe` extension in the install_name_tool command the way it was originally when doing things at this stage. This was tested with python38, but should also affect the 36 and 37 flavors.

dhomeier:
> install_name_tool -change /sw/lib/libpython3.8.dylib /sw/build.build/python38-3.8.1-1/Python-3.8.1/libpython3.8.dylib python
> error: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/install_name_tool: can't map file: python (Invalid argument)

Which Xcode version is that? So will simply renaming the executable to `python.exe` work? (on 10.12, Xcode command-line tools: 9.2.0.0.1.1510905681, it works with either name as long as the file exists...)

nieder:
10.12.6/Xcode9.2.0.0.1.1510905681

This is my %b directory listing. There's no plain `python`, just `python.exe` and manually running otool on 'python' here picks up the "Python" directory because of case insensitivity, so I think that's probably the source of the install_name_failure above (i_n_t on non-existent file FOO fails differently):
```
$ ls
CODE_OF_CONDUCT.md  Misc/               aclocal.m4          libpython3.8.dylib*
Doc/                Modules/            build/              m4/
Grammar/            Objects/            config.guess*       pybuilddir.txt
Include/            PC/                 config.log          pyconfig.h
LICENSE             PCbuild/            config.status*      pyconfig.h.in
Lib/                Parser/             config.sub*         python-config
Mac/                Programs/           configure*          python-config.py
Makefile            Python/             configure.ac        python.exe*
Makefile.pre        README.rst          install-sh*         python.exe-gdb.py
Makefile.pre.in     Tools/              libpython3.8.a      setup.py
$ otool -L python
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/objdump: 'python': Is a directory
```

dhomeier:
Ah, I'm building on a case-sensitive FS - my listing is
```
> ls -F                                                                                                                                
CODE_OF_CONDUCT.md	Makefile.pre		Programs/		config.status*		m4/			python-gdb.py
Doc/			Makefile.pre.in		Python/			config.sub*		platform		setup.py
Grammar/		Misc/			README.rst		configure*		pybuilddir.txt
Include/		Modules/		Tools/			configure.ac		pyconfig.h
LICENSE			Objects/		aclocal.m4		configure.orig*		pyconfig.h.in
Lib/			PC/			build/			install-sh*		python*
Mac/			PCbuild/		config.guess*		libpython3.8.a		python-config
Makefile		Parser/			config.log		libpython3.8.dylib*	python-config.py
```
which is probably the reason it picked the plain executable name.
Looks like I can force it to use the suffix anyway by adding  `--with-suffix=.exe` to the ConfigureParams, so that should be the best solution to ensure consistent builds.

dhomeier:
Adding `--with-suffix=.exe` to `ConfigureParams` appears to be the best solution to ensure consistent naming on both case-sensitive and insensitive FS, but it requires now a bit of tweaking in `InstallScript` to restore all the installed files without the suffix. It's not clear to me how this worked itself out previously on case-ignorant FS, just hoping it still works with these changes.
If not, the simpler remedy might be to just modify the `TestScript` conditional on the FS type (i.e. depending on the existence of a `python.exe`).
(Back)Forwarded the modifications to python35 as well for testing; 2 failures but no hangs. This should not be a top priority though, as the 3.5 branch is certainly on its way out as well.

dhomeier:
Additional note: I added `--enable-optimizations` to python38.info as suggested by the build script; trying the same for the 3.6/3.7 versions resulted in llvm segfaults on my system.

dhomeier:
> it requires now a bit of tweaking in `InstallScript` to restore all the installed files without the suffix. It's not clear to me how this worked itself out previously on case-ignorant FS, just hoping it still works with these changes.

Needed additional manual processing to make the installed scripts in `%i/bin` call the interpreter without `.exe` suffix. Newer upstream minor versions for 3.7 and 3.8; 3.8 also includes the `configure` fix from #604.
Both `python37` and `python38` are hanging in some tests with subprocesses running indefinitely and having to be killed externally, but otherwise appear to pass.

dhomeier:
> Both `python37` and `python38` are hanging in some tests with subprocesses running indefinitely and having to be killed externally, but otherwise appear to pass.

I've reverted to the default naming scheme for all versions now, checking within `TestScript` for the name of the executable. Also excluded the tests that were consistently hanging on my machine (also tried with fewer test processes down to `-j 1`); some other tests still fail, so they should be built with `--tests=warn`, or I have to change the return value to 1.

dhomeier:
Added fix to correct the order of library and include paths for python27 as well; might resolve #624, but needs testing on 10.14 or higher.
I have also removed the bulk substitution of `/usr/include` strings in `setup.py`, as

1. the required `$SDKPATH` paths should already be pulled from the Makefile

2. indiscriminately replacing all instances seems risky, since there are other checks like
   `if a in path and '/usr/include' not in path`...

nieder:
python27 needs a revision bump (readline change). It built for me on 10.14.6, but so did the current 2.7.18-1. For compiling readline.c, the -I flags changed from `-I/sw/include/ncursesw -I<SDK>/usr/include -I/sw/include` to `-I/sw/include -I/sw/include/ncursesw -I<SDK>/usr/include`

dhomeier:
> python27 needs a revision bump (readline change). It built for me on 10.14.6, but so did the current 2.7.18-1. For compiling readline.c, the -I flags changed from `-I/sw/include/ncursesw -I<SDK>/usr/include -I/sw/include` to `-I/sw/include -I/sw/include/ncursesw -I<SDK>/usr/include`

Thanks for the heads-up! Otherwise moving to readline7 is OK? It was the only Python version still using 6...
The flags order is how it's supposed to come out now. Still no idea why the include is working so inconsistently otherwise, but having `%p/include` at the front should at least give us the best setup to include the right file.

nieder:
On 10.13.6 with the latest listed here:
```
Validating .deb dir /sw/build.build/root-python38-3.8.3-1...
Error: Compiled python module points to fink install dir.
	Offending file: /sw/lib/python3.8/__pycache__/_bootlocale.cpython-38.pyc
```

@kamischi FYI, the tests `test.test_site.HelperFunctionsTests` & `test.test_site.StartupImportTests` failed on py34, py35, and py37. I've tracked it down to having sphinxcontrib-websupport-pyXX installed. The tests all pass otherwise when that's removed. I've tried updating to v1.1.2 (last that supports py34) and the failure still happens. Just noting it here to have someplace to put it before I forget. Don't know what can be done to deal with it.

dhomeier:
FWIW if all else fails we'd probably need to manually repeat the calls
```
PYTHONPATH=%i/lib/python%type_raw[python] DYLD_LIBRARY_PATH=%b \
	./python -E -Wi -O %i/lib/python%type_raw[python]/compileall.py \
	-j0 -d %p/lib/python%type_raw[python] -f \
	-x 'bad_coding|badsyntax|site-packages|lib2to3/tests/data' \
	%i/lib/python%type_raw[python]
PYTHONPATH=%i/lib/python%type_raw[python] DYLD_LIBRARY_PATH=%b \
	./python -E -Wi -OO %i/lib/python%type_raw[python]/compileall.py \
	-j0 -d /sw/lib/python%type_raw[python] -f \
	-x 'bad_coding|badsyntax|site-packages|lib2to3/tests/data' \
	%i/lib/python%type_raw[python]
```
following all the `make install` jingo.

nieder:
159d626 fixes the leftover .pyc issue for me on 10.13.

My vote is for commiting this. There are multiple examples on the mailing lists that this fixes the failed readline module compilation. It fixes the tests that I've had fail in the past. And the _scproxy failure is now pretty clear due to problems with badly installed/upgraded CLT, and not a problem with python itself.

nieder:
Merged. Any outstanding issues can go into new PR.

PR #542, opened 2019-12-21
Updated sdl2 with patch for 10.11 systems.

sth0:
Tested on 10.11 system, no response to maintainer contact.


nieder:
@fingolfin 
This is needed for SDL2 to build on 10.11. Here's the upstream source for this patch:
http://hg.libsdl.org/SDL/rev/449d5819a84a

I'll look into checking this in on Jan 1. It's a minimal change that fixes FTBFS. Please reply or approve. Thanks.

Issue #499, opened 2019-11-09
openssl100 migration tracker


nieder:
Openssl 1.0.2 is the current LTS and goes EOL on 2019-12-31.
https://www.openssl.org/policies/releasestrat.html

This lists our current packages using openssl100:

none
- [ ] cargo
- [ ] qt4-base-mac
- [ ] ettercap
- [ ] re-alpine
- [ ] tor
- [ ] xar
- [ ] lynx-cur

Chris Turkel
- [ ] links

@danielj7 
- [ ] libevent2
- [ ] libserf0
- [ ] pyqt4-mac-py
- [ ] pyqt5-mac-py
- [ ] python34
- [ ] ruby20
- [ ] ruby21
- [ ] ruby22
- [ ] ruby23
- [ ] cyrus-sasl2
- [ ] postfix
- [ ] rtmpdump

@dmacks 
- [ ] elinks

@dmalloc 
- [ ] xmlsec-shlibs

@elcepi
- [ ] lynx

@fangism 
- [ ] libtorrent15
- [ ] libtorrent19
- [ ] libtorrent20

@htodd 
- [ ] alpine-alpha

Jack Fink
- [ ] libetpan-shlibs
- [ ] stone

Jesse Alama
- [ ] swi-prolog

@kamischi 
- [ ] libimobiledevice

Keith Ward
- [ ] crypt-openssl-bignum-pm
- [ ] crypt-openssl-rsa-pm

@nieder 
- [ ] dclib5
- [ ] kdelibs4
- [ ] linuxdcpp
- [ ] virtuoso

@RangerRick 
- [ ] gst-plugins-bad-0.10
- [ ] qt4-base-x11
- [ ] qt5-qtbase
- [ ] libnetsnmp30-shlibs

@TheSin- 
- [ ] php5
- [ ] libc-client1-shlibs
- [ ] apache2
- [ ] nodejs

Stefan Bruda
- [ ] mupdf

elcepi:
I do not own a Mac anymore. Can someone take care of my packages?

TheSin-:
I have updated for all my packages but they can't be added till my PR is since they all require tiggers, except libc-client1 and nodejs which I haven't updated yet, I'll look into them. but I feel like nodejs update will require triggers as well.

dmacks:
[...]

PR #543, opened 2019-12-24
Nut update

dmacks:
openssl upgrade (for issue #499), and general dependency and build-system audit/fixes


clepple:
@dmacks thanks for putting this together. The build logs look good, and I plan to do some runtime testing this afternoon.

The `CFLAGS` -> `CPPFLAGS` changes generally made sense for the `-I` flags, but if that's the case in general, we might want to rename all of them upstream. The only catch would be if some of the pkgconfig files are introducing other flags. Haven't had a chance to check on that myself, but if you already researched that, I'll take your word for it.

nieder:
@dmacks ping

dmacks:
@nieder does this also resolve #636?

nieder:
@dmacks probably. Untested, but this explicitly BDepends on libnetsnmp35, so that should deal with the failure I saw on 10.14.5.

nieder:
ping?

clepple:
Sorry for the massive delay - I was hoping we'd have another release out by this point, but that's just not in the cards. Still don't have a great setup for testing here.

clepple (approved):
LGTM

PR #545, opened 2019-12-24
sth misc

sth0:
(I think I am doing this correctly....)
Updated some packages to new upstream code, adopted a couple of packages, and consolidated my different email addresses on all my packages.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.10.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.10.info
@@ -12,6 +12,6 @@ RuntimeDepends: <<
 <<
 Description: OBSOLETE: use %n%type_pkg[systemperl] instead
 License: Artistic
-Maintainer: Scott Hannahs <sth@magnet.fsu.edu>
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
 Homepage: http://search.cpan.org/dist/Config-IniFiles/
 <<
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.11.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.11.info
@@ -12,6 +12,6 @@ RuntimeDepends: <<
 <<
 Description: OBSOLETE: use %n%type_pkg[systemperl] instead
 License: Artistic
-Maintainer: Scott Hannahs <sth@magnet.fsu.edu>
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
 Homepage: http://search.cpan.org/dist/Config-IniFiles/
 <<
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.12.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/config-inifiles-pm-10.12.info
@@ -12,6 +12,6 @@ RuntimeDepends: <<
 <<
 Description: OBSOLETE: use %n%type_pkg[systemperl] instead
 License: Artistic
-Maintainer: Scott Hannahs <sth@magnet.fsu.edu>
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
 Homepage: http://search.cpan.org/dist/Config-IniFiles/
 <<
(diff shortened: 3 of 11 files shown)


nieder:
The current `Source:` URLs are fine, but to avoid having the hash in the URL and having to change the field for every version update, the following format currently works as a permalink:
```
https://pypi.python.org/packages/source/p/pyCLI/pyCLI-%v.tar.gz
```

sth0:
Ok will revise if I can get it to work.   I had trouble getting a rational permalink to work.  Maybe I am not getting the template correct.

curl https://pypi.python.org/packages/source/f/fasteners/fasteners-0.15.tar.gz

Should download the package but in this case I get a "301 Moved Permanently” response.  In fact I tried your example below (with the current version)

curl https://pypi.python.org/packages/source/p/pyCLI/pyCLI-2.0.3.tar.gz > ~/Downloads/pyCLI.tar.gz

and same “moved” response.  Am I missing something obvious?

-Scott

nieder:
This is Fink's curl command:
```
curl --connect-timeout 30 -f -L -A 'fink/0.45.0' -O https://pypi.python.org/packages/source/p/pyCLI/pyCLI-1.1.0.tar.gz
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0   122    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100   270  100   270    0     0    923      0 --:--:-- --:--:-- --:--:--   923
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
100  204k  100  204k    0     0   260k      0 --:--:-- --:--:-- --:--:--  260k
```
The `-L` flag causes curl to follow redirects and moves.

dhomeier:
> Ok will revise if I can get it to work. I had trouble getting a rational permalink to work. Maybe I am not getting the template correct. curl/source/f/fasteners/fasteners-0.15.tar.gz

@sth0, @nieder, if I am not mistaken the paths of type
`https://files.pythonhosted.org/packages/source/f/fasteners/fasteners-%v.tar.gz`
**are** the current stable download URL schemes, as they are still used in `isort-py.info` and `lazy-object-proxy-py.info` (and they did work for me for all the packages). The `https://pypi.python.org/packages` paths probably date back to before the site was moved to `pypi.org`.

As a side note, except for `requests-oauthlib` I succeeded in building `py38` variants of all the above modules without further changes; the latter requires the corresponding variant of `requests-mock` with an affiliated rat-tail of subdependencies...

sth0:
> On Dec 26, 2019, at 4:14 PM, Derek Homeier <notifications@github.com> wrote:
> 
> Ok will revise if I can get it to work. I had trouble getting a rational permalink to work. Maybe I am not getting the template correct. curl/source/f/fasteners/fasteners-0.15.tar.gz
> 
> @sth0 <https://github.com/sth0>, @nieder <https://github.com/nieder>, if I am not mistaken the paths of type
> https://files.pythonhosted.org/packages/source/f/fasteners/fasteners-%v.tar.gz
> are the current stable download URL schemes, as they are still used in isort-py.info and lazy-object-proxy-py.info (and they did work for me for all the packages). The https://pypi.python.org/packages paths probably date back to before the site was moved to pypi.org.
> 
> As a side note, except for requests-oauthlib I succeeded in building py38 variants of all the above modules without further changes; the latter requires the corresponding variant of requests-mock with an affiliated rat-tail of subdependencies…
> 
Well I think it is the -L which tells curl to follow the referenced new link.  I am not sure how long pypi will keep that forwarding link operational.

Both destinations will work:
curl --connect-timeout 30 -f -L -A 'fink/0.45.0' -O https://pypi.python.org/packages/source/f/fasteners/fasteners-0.15.tar.gz <https://pypi.python.org/packages/source/f/fasteners/fasteners-0.15.tar.gz>

curl --connect-timeout 30 -f -L -A 'fink/0.45.0' -O https://files.pythonhosted.org/packages/source/f/fasteners/fasteners-0.15.tar.gz <https://files.pythonhosted.org/packages/source/f/fasteners/fasteners-0.15.tar.gz>

both hosts map onto fastly.net and may even be a shared file structure.  Any thoughts on which host of the two is *more* stable?  It seems neither hosting scheme is documented.

I will look at requests-oauthlib and requests-mock.  I haven’t started trying to implement python 3.8 yet.

-Scott

dhomeier:
I don't know if the download links have been documented anywhere; seems that the `pypi.python.org` ones had been phased out sometime last year when first the `pythonhosted` links w/ hash popped up, then for a while `pypi.io` aliases where introduced before the direct `files.pythonhosted.org/packages/source` apparently became standard (see e.g. #171 and #238). I really thought the original `pypi.python.org` downloads had been deactivated at some point, but the redirects may have been restored later...

PR #548, opened 2019-12-29
update audiofile dependees

nieder:
update packages that use audiofile (and esound) so that they're all on the same libN=libaudiofile1. Otherwise, can end up with packages using old audiofile-shlibs and new audiofile-bin from libaudiofile1-shlibs.

libaudiofile-shlibs is now unused and turned into an shlibs-only stub.


dmacks:
Will leave it to you to nuke your branch if you like.

dhomeier:
Is there a way to do a clean update without being thrown back to `apt-get`?
```
> fink update audiofile-shlibs                                                                                                         
Information about 14785 packages read in 1 seconds.
 audiofile-shlibs
Reading buildlock packages...
	All buildlocks accounted for.

While trying to install:
  audiofile-shlibs-0.2.6-3

The following inconsistencies found:
  Unsatisfied dependency in audiofile: audiofile-shlibs (= 1:0.2.6-2)

Trying to resolve dependencies...
Could not resolve inconsistent dependencies!

Fink isn't sure how to install the above packages safely. You may be able to fix things by running:

  fink scanpackages
  sudo apt-get update
  sudo apt-get install audiofile-shlibs=1:0.2.6-3

Failed: Fink::SysState: Could not resolve inconsistent dependencies
```

nieder:
Unfortunately, I think that's a 'feature' of how audiofile was depended on audiofile-shlibs ( with = to %v-%r ). So fink can't install audiofile-shlibs-%r=3 because installed audiofile-%r=2 wants shlibs-%r=2. Just hand remove audiofile.

PR #551, opened 2020-01-02
mp3diags: dep updates

dmacks:
Switch to openssl110 (though doesn't seem to be used anyway. maybe obsolete IBD?) and misc other dep cleanups.


dmacks:
Not substantial changes, builds for me; no resp from maintainer in >2 weeks. I'm pushing

PR #552, opened 2020-01-02
stone: switch to openssl110

dmacks:
Thanks for the patch, debbugs!


dmacks:
I can't find a git account for maintainer Jack Fink

PR #557, opened 2020-01-10
libtool2: upstream bug fix

### LINKED ISSUE #541: libtool2 bug, fixed upstream
There is a typo in the ltmain.sh file that is installed in the .deb, which propagates into toolchains via 'libtoolize', and then can cause a crash during 'make'. Interestingly, it's a crash during an error state, so the effect is that this error masks the actual error the toolchain is trying to report. Upstream has an unreleased fix:

http://git.savannah.gnu.org/cgit/libtool.git/commit/build-aux/ltmain.in?id=350082b6aa89f9ef603fcebbb4cf33f15a743f2f

dmacks:
Closes #541

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/libtool2.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/libtool2.info
@@ -16,7 +16,7 @@
 
 Package: libtool2
 Version: 2.4.6
-Revision: 3
+Revision: 4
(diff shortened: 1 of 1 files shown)


dmacks:
>3 weeks no resp, is just a function-name typo leading to a crash, already accepted upstream.

fingolfin:
Ouch, sorry, I missed this :-(. Thank you for taking care of it.

PR #561, opened 2020-01-18
migrate wgscott packages to gcc9

nieder:
This is for #481 to close down gcc5.

There are upstream bumps for some packages (coot, mmdb2, libccp4), but I didn't follow them.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/clipper.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/clipper.info
@@ -1,21 +1,21 @@
 Package: clipper
 Epoch: 2
 Version: 2.1.20160708
-Revision: 1
+Revision: 3
 GCC:4.0
 BuildDepends: <<
 fftw (>= 2.1.5-1010) | fftw-mpi (>= 2.1.5-1010), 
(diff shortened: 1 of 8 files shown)


dmacks:
By eye, this looks sane (and also switched to newer boost, which is a Good Thing). I suspect some of the dependencies on "gcc9" could be reduced to "gcc9-compiler" but that shouldn't hold up getting this stuff committed...set a week timer?

Can PR #124 be merged into here?

dmacks:
Can these upgrade from fftw to fftw3? That would push out an old-version dep that is itself gcc5 (unmaintained).

nieder:
re: fftw
Not simply. At least one of them (coot?) was looking for -lsfftw, which does not have a direct replacement in fftw3. Others might be OK, but I wanted to keep the stack consistent.

dmacks:
Good idea. Might want to add a link to https://www.mail-archive.com/coot@jiscmail.ac.uk/msg00336.html to explain this intentional unusual linking.

dmacks:
These are just build/dep updates, not new versions...let's unblock these on new OSXen.

dmacks:
coot still has Depends:gcc5.

PR #562, opened 2020-01-19
xmlsec: new version (new -shlibs name)

dmacks:
Seems to be interface loss-of-features break from previous even though
same install_name version string, so repackaging in new %n and buried
.dylib

This is a security package and old had no reverse-depends and was tied
to EOL'ed old openssl100, so killing it altogether (instead of keeping
as -shlibs stub)


dmacks:
@dmalloc is maintainer of older package but doesn't seem to be listed as an available reviewer to approve (and seems inactive on fink and github for many years)

nieder:
The test logs are full of errors like this:
```
Test: /nist-aesgcm/aes256/aes256-gcm-96-408-0-128-13 (success)
/sw/build.build/xmlsec1-shlibs-1.2.28-1/xmlsec1-1.2.28/apps/xmlsec1 check-transforms  --crypto gcrypt --X509-skip-strict-checks --crypto-config /sw/build.build/xmlsec1-shlibs-1.2.28-1/xmlsec1-1.2.28/tests/tmp/xmlsec-crypto-config aes256-gcm
Error: parameter "--X509-skip-strict-checks" is not supported or the requested
feature might have been disabled during compilation.
Error: invalid parameters
```
manual says that's a valid flag for xmlsec1. Does it matter?

dmacks:
Try this new new-version.

PR #565, opened 2020-01-26
First attempt to install dateutils.

htodd:
Can someone check this? I just make sure -m doesn't error.

@nieder @dmacks


nieder:
The package also has a fairly large test suite. Is it worth enabling? It passed when I ran it by hand.

htodd:
I think I fixed it @nieder

htodd:
Yeah, I give up on GitHub.

PR #571, opened 2020-02-09
Add missing dependency on help2man

nieder:
help2man is needed to rebuild doc/libtool.1


dmacks:
This was reported as "Fail in updating libtool2 libtool2-shlibs" on the fink-users mailing list. LGTM trivial fix, but leaving it for $maintainer

nieder:
Closes #581

PR #572, opened 2020-02-11
Disable apexplugin (abandoned and not openssl110)

dmacks:


dmacks:
Running of the openssl autoconf test is conditional on the plugin not being disabled.

PR #575, opened 2020-02-13
Re-enable cups on 10.15 and unify .info files

nieder:
10.15 (or Xcode11) ditched the older cups API that gtk2-2.18 was using.
Patches to upgrade cups functions by Tom Liehe from fink-beginners


nieder:
Added DescPort bit about the new CUPS patch

dmacks:
Okay to delete this branch?

PR #577, opened 2020-02-21
OpenBLAS updated to 0.3.8 and fixed smallscaling build on 10.11 and lower

dhomeier:
Updated the `openblas` package to latest upstream and worked around a failure on compilation of the `smallscaling` benchmark program on OS X 10.11 (and presumably earlier) due to missing `CLOCK_MONOTONIC` reported [here](https://sourceforge.net/p/fink/mailman/message/36927850).
Also modified the `-mtune` option - see discussion in `DescPackaging`, as the `intel` option did not seem to enable all optimisations for the latest CPUs supported with gcc9 - might see some performance improvement on the latest iMac Pro or Mac Pro.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
@@ -14,7 +14,7 @@ Source: https://github.com/xianyi/OpenBLAS/archive/v%v.tar.gz
 Source-MD5: a3cb780c2d79e3fe13af58a261308fdf
 SourceRename: OpenBLAS-%v.tar.gz
 PatchFile: %n.patch
-PatchFile-MD5: f4ab9508a7d4bde8e28ba4e84707056e
+PatchFile-MD5: 3aefa8256082dd2c2edf04e4d25c6a7d
 
 PatchScript: sed 's|@PREFIX@|%p|g' < %{PatchFile} | patch -p1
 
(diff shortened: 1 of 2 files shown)


dhomeier:
Thanks to David N. for pointing out a possible substitute for CLOCK_MONOTONIC, which is now included in the patch. It turned out to require yet a bit more to build a functional `smallscaling`, because it still needs a `clock_gettime` without undue overhead in the timing calls. I could only test this on a 10.6 system (after hacking the package to build with gcc48), but the results look sensible now.

dhomeier:
Addendum: I have also tested the option for [building multiple targets](https://github.com/xianyi/OpenBLAS#support-for-multiple-targets-in-a-single-library) `DYNAMIC_ARCH = 1`, which seems maybe more adequate for building binaries for all supported OS X versions - that would be currently `DYNAMIC_LIST="CORE2 PENRYN NEHALEM SANDYBRIDGE HASWELL SKYLAKEX"`. However I have found no way to compile the `kernel/x86_64` variants for AVX instructions not present on the build architecture, e.g.

- on Core i7 (registers up to `hw.optional.avx2_0: 1`) the AVX512 (*zmmN*) instructions fail

- on Core2 (registers up to `hw.optional.sse3: 1`) AVX (*ymm[0-15]*) and AVX2 (*ymm[16-31]*) instructions fail

although from the docs and stackoverflow discussion it sounds as it should be possible to compile them the `-march=skylake-avx512` (so these object files would not work on pre-Skylake CPUs, but the library should automatically switch to the highest supported version linked in). But in both cases above compiling the C or assembler sources fails on inlined *ymm** or *hmm** instructions, and I haven't found a gcc option to make it ignore these errors.

The `DYNAMIC_ARCH` versions of the libraries are 60-80 % larger, but don't seem to incur noticeable performance penalties. I don't know if we have an AVX512-capable build host that might create binary package that hopefully would run on all our supported CPU versions, otherwise it would probably be easier to stick with the host-specific packages built from source.

dhomeier:
Additional successful test under 10.11 (4-Core Mac Pro) reported by David N.

PR #585, opened 2020-03-17
Fix circular dependencies

babayoshihiko:
CRAN testthat and prettyunits depend on each other, causing a circular dependency conflict.

testthat:
Added 
DEP: cran-prettyunits-r%type_pkg[rversion] (>= 1.1.1-2),
CONFLICTS: cran-prettyunits-r%type_pkg[rversion] (= 1.1.1-1)

prettyunits:
Removed the dependencies (which were only "suggested").

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-prettyunits-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-prettyunits-r.info
@@ -34,7 +34,7 @@ Distribution: <<
 <<
 Type: rversion (3.6 3.5 3.4 3.3 3.2 3.1)
 Version: 1.1.1
-Revision: 1
+Revision: 2
 Description: Human Readable Formatting of Quantities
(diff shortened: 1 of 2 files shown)


nieder:
testthat needing prettyunits seems to be an indirect dependency.
testthat -> pkgload -> pkgbuild -> prettyunits

Does it need to be explicitly listed in `cran-testthat-r:Depends:` ?

babayoshihiko:
> testthat needing prettyunits seems to be an indirect dependency.
> testthat -> pkgload -> pkgbuild -> prettyunits
> 
> Does it need to be explicitly listed in `cran-testthat-r:Depends:` ?

That was also my question.

Is Conflicts: cran-prettyunits-r (= 1.1.1-1) sufficient to avoid this bug?

nieder:
To be honest, I'm not sure if either the explicit indirect Depends: or the Conflicts: are needed. I think as long as the final set of Dependencies for both packages are correct, it should work itself out. I don't think having them will cause problems however.

PR #587, opened 2020-03-31
Cran 20200331 new upstream versions

babayoshihiko:
These packages are simple updates.  Some packages (catools, chemospec, glmnet, hmisc, hyperspec, mcmcpack, modeest, mstate, partitions, pbkrtest, rattle, rcmdrplugin.ezr, rms, rngtools, robcompositions, tidyselect, vim) have higher R dependency and thus may produce additional files for older Rs.  License is corrected for some files (catools, haven, openxlsx).


nieder:
Does the new `cran-tidyselect-r.info` still need libgettext8? That used to be only needed when a cranmod built a shared module .so/.dylib and it seems that v1.0.0 doesn't have that any more.

babayoshihiko:
> Does the new `cran-tidyselect-r.info` still need libgettext8? That used to be only needed when a cranmod built a shared module .so/.dylib and it seems that v1.0.0 doesn't have that any more.

Removed libgettext8 from Dep and BuildDep.

PR #589, opened 2020-04-05
Ruby updates

nieder:
Updates for the ruby packages that are still maintained upstream. Simple %v bumps. ruby24 has a little extra change added one test that consistently failed w/out that change. With the change, that test consistently passes on my system.

Tested on 10.13/Xcode10.1


nieder:
Pushing.
Month with no contact and there are some security updates going here.
ruby24 has been end of lifed at v2.4.10, so it might be a good idea to dist restrict it and the earlier rubys as well (ruby20-ruby23 are OPENSSL110 FTBS anyway).

PR #592, opened 2020-04-13
tcltk update to 8.6.10 with MacOSX headers and Aqua support.

dhomeier:
When trying to build a package against Tcl/Tk, compilation failed because 8.6.7's private headers `default.h` and `tkPort.h` are including `tkMacOSXDefault.h` and `tkMacOSXPort.h`, respectively, which are not packaged with `tcltk-dev`. The project also needs `_TkMacOSXMakeRealWindowExist` and some other functions that were not compiled in the current library.

This update to 8.6.10 has the required `TkMacOSX*` included by enabling Aqua support. All needed headers are added from the `generic`, `unix` and `macosx` directories; since they all have unique names, I have changed the structure of `%p/include/tcltk-private` to use a single subdirectory for `tcl8.6` and `tk8.6` each. This shortens the paths a bit and avoids additional patches in the likes of
`#	include "../macosx/tkMacOSXPort.h"`

The only  package I found that is building against the private headers is `expect`; I've checked that the existing package still works and passes all tests with the new `tcltk-shlibs` installed, but to build against `tcltk-dev-8.6.10-1` of course the include paths need to be adapted. This is done with the update to `expect-5.45.4`.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/tcltk-8.6.10.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/tcltk-8.6.10.info
@@ -1,0 +1,261 @@
+Package: tcltk
+Epoch: 1
+
+# update itcl path-flags in "itk" and versions in "iwidgets" when updating
+Version: 8.6.10
+
+Revision: 1
+BuildDepends: <<
+	fink (>= 0.28),
+	fink-package-precedence,
+	fontconfig2-dev (>= 2.10.0-1),
+	freetype219 (>= 2.6-1),
+	pkgconfig,
+	x11-dev,
+	xft2-dev (>= 2.2.0-1)
+<<
+Depends: <<
+	%N-shlibs (= %e:%v-%r),
+	fontconfig2-shlibs (>= 2.10.0-1),
+	freetype219-shlibs (>= 2.6-1),
+	x11-shlibs,
+	xft2-shlibs (>= 2.2.0-1)
+<<
+Source: mirror:sourceforge:tcl/tcl%v-src.tar.gz
+Source-MD5: 97c55573f8520bcab74e21bfd8d0aadc
+SourceDirectory: tcl%v
+Source2: mirror:sourceforge:tcl/tk%v-src.tar.gz
+Source2-MD5: 602a47ad9ecac7bf655ada729d140a94
+PatchFile: %n-%v.patch
+PatchFile-MD5: 458c0e7ea6996a7e0f1e762791f8269a
+PatchScript: <<
+	%{default_script}
+	# patch *ancient* darwin-ignorant autoconf
+	perl -pi -e 's/(a so sl)/dylib \1/' tk%v/unix/configure
+	# autoconf2.6ish patch for modern XQuartz paths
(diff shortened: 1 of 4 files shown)


dmacks:
MAC_OSX_TK is defined for the Aqua GUI model, but fink's tk is configured for x11 and therefore has the tkUnix*.h instead of tkMacOSX*.h component headers. If you have something that is using fink's tk but looks for the Aqua headers at build-time, there's something broken in its autodetection/configuration. Perhaps it's using "on OS X" as a proxy for "uses OS X graphics" or it's finding Apple's tk but then mixing in fink's headers at compile-time?

dhomeier:
> If you have something that is using fink's tk but looks for the Aqua headers at build-time, there's something broken in its autodetection/configuration. Perhaps it's using "on OS X" as a proxy for "uses OS X graphics" or it's finding Apple's tk but then mixing in fink's headers at compile-time?
> 
Background of this whole update is, I am trying to build a reasonably recent version of vtk as a dependency of a (Python 3-supporting) update to Mayavi2. Both vtk 8.1.2 and 8.2.0 build with support for (one of) Pythons 3.6 - 3.8, but built against tk 8.6.7 running just about anything fails on detecting an even remotely up-to-date OpenGL. Thus my attempt with this Aqua-enabled tk, which lets all vtk tests and almost all of the examples pass.
An alternative could of course be to provide a sufficiently updated OpenGL that is available under x11. But Fink's Mesa package is now about 13 major versions behind, and while I was able to get it bumped to 7.11, I don't feel like tackling the jump to 20.x... Or am I missing some other path to get OpenGL support in an x11 package?

> Switching the underlying GUI model of the "tcltk" package is a non-starter because it breaks the behavior of the existing package and is the root of a bunch of other packages that need a consistent GUI model in their dep-tree. "tcltk-aqua" as a separate %n in a separate location (that is, not having filename collisions with the existing), like we have gui-varianted qt and kde.

Yes, I was getting worried this might get out of control, given the number of packages depending on it. Although `expect` had been the only one explicitly configuring for the `tcltk-private` includes, it turned out most of the dependent packages used them in the default setup.
So restoring the original structure with separate `generic`, `unix` and `macosx` turned out to cause the least breakage; however one or two packages so far still needed patching their setup, since even the `--enable-aqua=yes` `tcltk` build includes X11 headers e.g. in `generic/tk.h`, `unix/tkUnixPort.h` and `macosx/tkMacOSXPort.h`, yet the default configures in many packages forcibly remove any `X11_INCLUDES` once they detect any `MAC_OSX` defines.

Seeing if I can get vtk 8.1 built against the system tcltk right away now, otherwise maybe I will have to think about the completely separate "tcltk-aqua" alternative. But as you noted, having it install directly install into `%p/lib` won't make this easier.

dmacks:
If this is WONTFIX by the maintainer and we now have tcltk-8.6.10 via a different PR (a variant of which has now been pushed to master), should this PR be closed?

dhomeier:
Apart from the itcl conflict #681 seems to have taken care of that, and I checked that our existing expect also builds against it. I'll see if I can eventually get a newer VTK to build, but should I end up needing the Aqua version after all, I'll just reopen this or a new PR.

PR #596, opened 2020-04-16
jbig2dec 0.17

nieder:
update jbig2dec to latest upstream 0.17


nieder:
@danielj7 ping

PovlAbrahamsen:
Note that version 0.19 is now available...

dmacks:
The patches to the configure file no longer apply. Looks like upstream is now using a more modern libtool toolchain and libpng API test. Not sure these were even needed back in 0.15, but while we have third-party eyes looking at this, may as well scrap obsolete hackery. The SetLDFLAGS is also not needed (autoconf now tests for libintl).

I un-disabled dependency-tracking and added fink-package-precedence, didn't uncover anything missing. But libgettext actually isn't used at all despite being autodetected. I think it would only be invoked at compiletime if OS X didn't have getopt_long (which it does). And without needing that, we also don't need libiconv. Can just patch out "libintl.h" from configure

According to otool, the splitoff's lib/libjbig2dec.0.dylib does not use libpng16 (only the main package's bin/jbig2dec does).

PovlAbrahamsen:
I think you can also delete jbig2dec.patch, which appears to be a remnant from version 0.11, from the repository!

PovlAbrahamsen:
It's now been over six months since the original PR. I understand the need to review updates, but long delays like this really do hold back Fink; it's frustrating to find that libraries in Fink are often badly out of date...

nieder:
Pushing. No communication from maintainer in 6 months and has been looked at by several others.
It's also required for other package upgrades.

PR #600, opened 2020-04-19
New upstream versions

babayoshihiko:
cran-rcpp-r has an include directory in a subdirectory, which is moved to the -dev variant.  I am not sure if this is the best solution.
The blow packages use R-3.2 functions and thus -r31 is left unupdated: glue, locfit, lubridate, prospectr, subselect and tibble.
cran-forecast-r31 does not build since 8.10, so reverted to 8.9.  Similarly, cran-igraph-r31 is reverted to 1.2.4.1.


nieder:
`DescPackaging` in cran-lubridate-r.info is screwy with the "License: GPL" line at the very end of the `DescPackaging` field. The file already has a `License:` field in line 36.

PR #602, opened 2020-04-21
mtr: new upstream

htodd:
This needs some checking.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/net/mtr.info
+++ b/10.9-libcxx/stable/main/finkinfo/net/mtr.info
@@ -1,9 +1,9 @@
 Info2:<<
 Package: mtr%type_pkg[-nox]
 Type: -nox (boolean)
-Version: 0.86
+Version: 0.93
 Revision: 1
-Maintainer: Hisashi T Fujinaka <htodd@users.sourceforge.net>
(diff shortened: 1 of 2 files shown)


dmacks:
During InstallScript, I see:
```
`setcap cap_net_raw+ep /sw/build.build/root-mtr-nox-0.93-1/sw/bin/mtr-packet` \
|| chmod u+s /sw/build.build/root-mtr-nox-0.93-1/sw/bin/mtr-packet
/bin/sh: line 1: setcap: command not found
```
I think that's a linux-only thing. The makefile recipe is:
```
install-exec-hook: 
	`setcap cap_net_raw+ep $(DESTDIR)$(sbindir)/mtr-packet` \
	|| chmod u+s $(DESTDIR)$(sbindir)/mtr-packet
```
but the .deb winds up losing the u+s flag, so the resulting /sw/bin/mtr-packet does not have the intended permissions. How about removing this block altogether (patch Makefile.am) so users don't get anxious at the irrelevant error message, and put the chmod in PostInstScript.

You can save yourself two whole lines by scrapping ```SourceDirectory:``` and ```--mandir=```, as those are the default values for those variables.

htodd:
@dmacks Added your changes. Please make sure I did it correctly.

Thanks.

PR #603, opened 2020-04-21
fping

htodd:
This built with a very minimal info file but I'm not sure I have all the bits right.


htodd:
@dmacks, I think I took all your suggestions, but I'm not sure how to separate out mtr.

htodd:
@dmacks I think I fixed it and I think I figured out how to get the mtr files out.

htodd:
@dmacks I think I fixed it and I think I figured out how to get the mtr files out.

PR #605, opened 2020-04-22
Update pysvn to 1.9.11 for svn19 1.13 compatibility and add py38 variant

dhomeier:
pysvn 1.9.10 fails to build against `svn19` 1.13. I've bumped the version to latest upstream, tested to build both against `svn19-dev` 1.12.1 and 1.13.0, and also added a `py38` variant.
https://pysvn.sourceforge.io/downloads.html only states support for Python 3.5 upwards, but the `py34` variant built without any obvious problems (again with both `svn19` versions).

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pysvn-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pysvn-py.info
@@ -1,9 +1,9 @@
 # -*- coding: ascii; tab-width: 4 -*-
 Info2: <<
 Package: pysvn-py%type_pkg[python]
-Version: 1.9.10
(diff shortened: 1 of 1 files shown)


nieder:
WFM. Simple version bump to deal with known failure with latest version of a dependency.
What does the `--embed` flag do in py38?

dhomeier:
> What does the `--embed` flag do in py38?

Change in the `python-config` interface: `--ldflags` by itself only provides the framework libraries `-ldl -Wl,-framework,CoreFoundation`; to link to `libpython3.8.dylib` as well, as of 3.8 this flag has to be added.

nieder:
Pushing. No response and is simple fix to deal with an dependency lib change.

PR #608, opened 2020-04-25
Add Java fix to R; CRAN rJava

babayoshihiko:
R has Java settings stored in Makeconf and ldpaths, but they do not match.  The fix is to modify Makeconf to read ldpaths.

With Java fix to r-base, rJava builds without any extra fix.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-rjava-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-rjava-r.info
@@ -33,14 +33,14 @@ Distribution: <<
 	(%type_pkg[rversion] = 35 ) 10.14.5
 <<
 Type: rversion (3.6 3.5 3.4 3.3 3.2 3.1)
-Version: 0.9-8
-Revision: 2
+Version: 0.9-12
+Revision: 1
 Description: Low-level R to Java interface
 Homepage: https://cran.r-project.org/package=rJava
 License: GPL
 Maintainer: None <fink-devel@lists.sourceforge.net>
 Source: mirror:custom:rJava_%v.tar.gz
-Source-MD5: 51ae0d690ceed056ebe7c4be71fc6c7a
+Source-MD5: 8d17e50cda6297a85254f7b9cd57db60
 SourceDirectory: rJava
 CustomMirror: <<
 	Primary: https://cran.r-project.org/src/contrib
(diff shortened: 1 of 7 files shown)


nieder:
I haven't tested anything yet, but why is r-base36 the only one that needs to use `/usr/libexec/java_home` ?

babayoshihiko:
> I haven't tested anything yet, but why is r-base36 the only one that needs to use `/usr/libexec/java_home` ?

Oops.  I fixed it.

babayoshihiko:
on my environment, r-base31 to r-base33 require additional Java for OS X 2017-001 installation.

nieder:
This is the resulting java related block in `Resources/etc/Makeconf`:
```
JAR = $(/usr/libexec/java_home)/bin/jar #/usr/bin/jar
JAVA = $(/usr/libexec/java_home)/bin/java #/usr/bin/java
JAVAC = $(/usr/libexec/java_home)/bin/javac #/usr/bin/javac
JAVAH = /usr/bin/javah
## JAVA_HOME might be used in the next three.  
## They are for packages 'JavaGD' and 'rJava'
JAVA_HOME = /Library/Java/JavaVirtualMachines/jdk1.8.0_66.jdk/Contents/Home/jre
JAVA_CPPFLAGS = -I$(JAVA_HOME)/../include -I$(JAVA_HOME)/../include/darwin
JAVA_LIBS = -L$(JAVA_HOME)/lib/server -ljvm
JAVA_LD_LIBRARY_PATH = $(JAVA_HOME)/lib/server
```
* Should JAVAH also be changed like JAR|JAVA|JAVAC? I don't know the correct answer.

babayoshihiko:
>     * Should JAVAH also be changed like JAR|JAVA|JAVAC? I don't know the correct answer.

It may be safe to have JAVAH change.  By the way, what is your ldpaths?

I have jdk1.8.0_231.jdk in /Library/Java/JavaVirtualMachines, but JAVA_HOME and ldpaths ignore it and use jdk-14.0.1.jdk.

babayoshihiko:
I was a bit confused.  If JAVA_HOME from ldpaths is to be used, 

  sed -i -e 's|JAR = |JAR = $(/usr/libexec/java_home)/bin/jar #|g' %i/Library/Frameworks/R.framework/Versions/%type_raw[rversion]/Resources/etc/Makeconf

should be changed to:

  sed -i -e "s|JAR = |JAR = ${JAVA_HOME}/bin/jar #|g" %i/Library/Frameworks/R.framework/Versions/%type_raw[rversion]/Resources/etc/Makeconf

But this will result in hardcoded JAVA_HOME.  What do you think?

babayoshihiko:
I have updated the r-base3?.info.

Now both ldpaths and Makeconf has /usr/libexec/java_home.  But ldpaths is hardcoded, while Makeconf reads /usr/libexec/java_home.

This means binary package may not work if JAVA_HOME is different.  Should ldpaths be set to read /usr/libexec/java_home?  If it is not possible, maybe set at post-install?

nieder:
What's the logic behind moving the JAVA_HOME fix into r-baseXX-shlibs::PostInstScript? The drawback from doing it this way is that it changes the MD5 of the installed ldpaths file vs what's in the .deb and sanity checkers will complain about a changed file.

Separately, in `Resources/etc/Makeconf`, JAVA_HOME is set to `/Library/Java/JavaVirtualMachines/jdk1.8.0_66.jdk/Contents/Home/jre`, but PostInstScript edits the JAVA_HOME path in `Resources/etc/ldpaths` to `/Library/Java/JavaVirtualMachines/jdk1.8.0_66.jdk/Contents/Home` (trailing "jre" removed). And so `R_JAVA_LD_LIBRARY_PATH` in the `ldpaths` file no longer resolves because "lib/server" is under the full path ending in `/Library/Java/JavaVirtualMachines/jdk1.8.0_66.jdk/Contents/Home/jre` path.

I have not tried to build cran-rjava yet, so I don't know if the above situation is needed or blocks building. I just want to make sure the output is as expected.

babayoshihiko:
> What's the logic behind moving the JAVA_HOME fix into r-baseXX-shlibs::PostInstScript? The drawback from doing it this way is that it changes the MD5 of the installed ldpaths file vs what's in the .deb and sanity checkers will complain about a changed file.

ldpaths has hardcoded JAVA_HOME.  This means the .deb file has hardcoded JAVA setting.  When someone tries to install from binary which is different from his/her JAVA_HOME, the binary-installed ldpaths will not work.

PostInstScript will modify the content of the file based on the user's /usr/libexec/java_home.

But you are right.  It is not a good solution.  Also, this solution will not work if the user changes java_home.

babayoshihiko:
Maybe we should stick to system-java14?

babayoshihiko:
When adding 

export JAVA_HOME=$(/usr/libexec/java_home -v 1.6) # or whatever version

in CompileScript of r-base* packages, all the settings seem to be correctly set

The question is, is it OK for r-base* packages to depend on system-java16 (the latest Java JDK?)?

It seems that cran-rJava-r* cannot be built with older JDK (e.g. 14).

nieder:
(never saw this)
It's fine to force a specific system-java version. But note that Apple's old java1.6 (available through 10.14) shows up as system-java16 in Fink. It wasn't until java9 that the java versioning scheme dropped the leading '1'.

PR #609, opened 2020-04-26
picocom: new upstream

htodd:
Only build tested so far. Please let me know if anyone tests this.


htodd:
@nieder I think I got your changes in.

PR #617, opened 2020-05-02
Fix extutils-makemaker-pm to deal with hardened system-perl on 10.15

### LINKED ISSUE #461: term-readkey-pm build failure with Xcode11
On 10.15b7 with Xcode GM, term-readkey-pm fails with this error:
```
cc  -I/Applications/Xcode-beta.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/System/Library/Perl/5.18/darwin-thread-multi-2level/CORE -bundle -undefined dynamic_lookup -fstack-protector  ReadKey.o  -o blib/arch/auto/Term/ReadKey/ReadKey.bundle  \
	      \
	  
chmod 755 blib/arch/auto/Term/ReadKey/ReadKey.bundle
"/usr/bin/arch" -x86_64 perl5.18 "-Iblib/arch" "-Iblib/lib" ReadKey_pm.PL ReadKey.pm 
Creating ReadKey.pm
Can't load 'blib/arch/auto/Term/ReadKey/ReadKey.bundle' for module Term::ReadKey: dlopen(blib/arch/auto/Term/ReadKey/ReadKey.bundle, 1): no suitable image found.  Did find:
	file system relative paths not allowed in hardened programs at /System/Library/Perl/5.18/darwin-thread-multi-2level/DynaLoader.pm line 194.
 at /System/Library/Perl/5.18/darwin-thread-multi-2level/DynaLoader.pm line 98.
	DynaLoader::croak('Can\'t load \'blib/arch/auto/Term/ReadKey/ReadKey.bundle\' fo...') called at /System/Library/Perl/5.18/darwin-thread-multi-2level/DynaLoader.pm line 194
	DynaLoader::bootstrap('Term::ReadKey') called at ReadKey_pm.PL line 515
Bootstrapping the XS for blockoptions: make: *** [ReadKey.pm] Error 255
```
Looks like more 'security' from Apple.
@cschaffner
@dmacks

nieder:
Fix #461 by patching EU::MM to use absolute paths for INST_ARCHLIB so that system-perl on 10.15 is happy loading bundles.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.11.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.11.info
@@ -1,6 +1,6 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
 Package: extutils-makemaker-pm
-Version: 7.36
+Version: 7.44
 Revision: 1
 Distribution: 10.11
 Depends: system-perl5182, %n5182 (>= %v)
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.12.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.12.info
@@ -1,6 +1,6 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
 Package: extutils-makemaker-pm
-Version: 7.36
+Version: 7.44
 Revision: 1
 Distribution: 10.12
 Depends: system-perl5182, %n5182 (>= %v)
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.13.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.13.info
@@ -1,6 +1,6 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
 Package: extutils-makemaker-pm
-Version: 7.36
+Version: 7.44
 Revision: 1
 Distribution: 10.13
 Depends: system-perl5182, %n5182 (>= %v)
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.5.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.5.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
 Package: extutils-makemaker-pm
-Version: 7.36
-Revision: 2
+Version: 7.44
+Revision: 1
 Distribution: 10.14.5
 Depends: system-perl5184, %n5184 (>= %v)
 Type: bundle
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/extutils-makemaker-pm-10.14.info
@@ -1,6 +1,6 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: extutils-makemaker-pm.info -*-
 Package: extutils-makemaker-pm
-Version: 7.36
+Version: 7.44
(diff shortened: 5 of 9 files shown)


dmacks:
Thanks for figuring this all out! As I read the extutils-makemaker-pm.info:
```
if [ "$darwin_vers" -ge 18 ]; then
	# only apply SDK fix for 10.14+ (that hides system-headers) and systemperl
	if [[ "$darwin_vers" -ge 18 && $type_num[perl] -eq 5184 ]]; then
	  patch -p1 < %{PatchFile2}
fi
```
would mean building on 10.14+ gets a different .deb than on 10.9-10.13 for the -pm5184 variant (same %n-%v-%r). I think that would mean we need dist-separated %r. Does the patched version not work with fink's perl5184 on those lower systems?

nieder:
I haven't tried applying the SDK patch on <= 10.13/pm51*. I know 10.14&10.15/pm5182 perlmods fail with the EU::MM SDK patch applied (hence the two-test conditional). It's easy to test on earlier OS by just removing the `$darwin_vers -ge 18` conditional, building EU::MM with it, and then building a package like `package-stash-pm`.

htodd:
I made a comment on the web page. I don't understand how github and
email work together.

On Sat, 2 May 2020, nieder wrote:

> 
> I haven't tried applying the SDK patch on <= 10.13/pm51*. I know 10.14&10.15/pm5182 perlmods fail with the EU::MM SDK patch
> applied (hence the two-test conditional). It's easy to test on earlier OS by just removing the $darwin_vers -ge 18 conditional,
> building EU::MM with it, and then building a package like package-stash-pm.
> 
> ?
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, orunsubscribe.[AANLE4YPEHZIKQGMNOBYFHLRPRWDNA5CNFSM4MXYYRS2YY3PNVWWK3TUL52HS4DFVREXG43VMVBW63LNMVXHJKTDN5WW2ZLOORPWSZGOEURDXLA.gi
> f]
> 
> 
>

-- 
Hisashi T Fujinaka - htodd@twofifty.com
BSEE + BSChem + BAEnglish + MSCS + $2.50 = coffee

nieder:
On 10.13, extutils-makemaker-pm buils w/ and without the patch.  And package-anon-pm5184 builds w/ and without the EU::MM SDK patch. The package-anon-pm build log is slightly different:
Without the SDK patch:
>gcc -c   -I/sw/include -fno-common -DPERL_DARWIN -no-cpp-precomp -mmacosx-version-min=10.13 -fno-strict-aliasing -pipe -fstack-protector -I/sw/include -O3   -DVERSION=\"0.05\" -DXS_VERSION=\"0.05\"  "-I/sw/lib/perl5-core/5.18.4/darwin-thread-multi-2level/CORE"   Anon.c

With the SDK patch originally meant for 10.14+:
> gcc -c   -I/sw/include -fno-common -DPERL_DARWIN -no-cpp-precomp -mmacosx-version-min=10.13 -fno-strict-aliasing -pipe -fstack-protector -I/sw/include -O3   -DVERSION=\"0.05\" -DXS_VERSION=\"0.05\"  -iwithsysroot "/sw/lib/perl5-core/5.18.4/darwin-thread-multi-2level/CORE"   Anon.c

The difference is the -I flag to `darwin-thread-multi-2level/CORE` gets changed to a `-withsysroot` parameter.

So the failure of applying the patch on non-system perl only matters for 10.14 and up.

nieder:
> I made a comment on the web page. I don't understand how github and email work together.

On a github ticket, you can add a comment via the web, which then sends an email to everyone that's subscribed to that ticket. You can also reply to one of those emails, and the receiving github server then posts that email as a new entry in the ticket. And so everyone that's subscribed to the ticket will then get an email about your email/post.

nieder:
@dmacks There might be a way to to test the perl type (`/System/Library/Perl` vs not-that) and then have rel2abs kick-in to fix paths, but that's way beyond what I can do (there's already some code in EU::MM to identify the perl type, so perhaps something could be copied).

This is an accurate summary of what happens with the different perls I think:
10.13 and earlier:
system-perl: works with and without rel2abs patch
fink-perl: works with and without rel2abs patch

10.14 and above:
system-perl: needs rel2abs patch
fink-perl: fails with rel2abs patch

To do everything at runtime and avoid revisioning by distribution, runtime EU::MM::MM_Any.pm would need to check for system-perl being greater than or equal to 5.18.4. That's beyond my perl abilities.

nieder:
@dmacks since no one is working on how to apply the SDK fix to do the include flag magic at runtime, I'm going to commit this in a couple days. It's better to have a working system that can be improved vs a totally broken perl system.

dmacks:
concurred

PR #620, opened 2020-05-06
XEphem fixes for Guide Star Catalogue and other downloads

dhomeier:
Upstream package has not seen any real updates in 5 years but is still working fine, however some of the online sources for background star catalogues, minor planets and comets ephemerides have moved and/or changed formats.

The initial motivation for the fix was the failure to load background stars from the HST Guide Star Catalogue 2.4 (enabled in the `Field stars` menu, which is still successfully accessed, but could not be parsed due to a change in no. and order of columns. The download test was also trying to access R band magnitudes, which are missing for almost all stars in GSC, making the setup fail. I have replaced them with the IR J magnitude; this is only used as a last fallback option for a stellar magnitude.

Also updated default URLs for some other sites, but reading them in via the `Download` menu still fails on parsing them for reasons I could not track down – especially as the files are already provided in XEphem format and can simply be manually placed into the `catalogs` or `~/Library/XEphem` directory.
Added additional patches from http://www.clearskyinstitute.com/xephem/contrib/contrib.html that should enable handling of `https` downloads.

Since the package only comes with a very minimal set of data files, I have also tried to make it easier to use separately downloaded catalogue files (from the complete tarballs on the XEphem site or directly from online sources like minorplanetcenter.net). The executable is now patched to use the Fink install location `%p/lib/xephem` as hardcoded default, and will therefore find the included data without the wrapper script setting `TELHOME`. If additional data are installed outside of %p – e.g. the CD versions of the GSC 2.2 come with a `pkg` installer putting them in `/usr/local/xephem` – launching xephem with the `-env TELHOME=/usr/local` option will let it find them (it would also find them automatically in the user's `~/Library/XEphem`, but this would provide an option to make them available system-wide).

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/xephem.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/xephem.info
@@ -1,14 +1,22 @@
+# -*- coding: ascii; tab-width: 4; x-counterpart: xephem.patch -*-
+Info2: <<
 Package: xephem
 Version: 3.7.7
-Revision: 1
+Revision: 2
 BuildDepends: <<
 	fink-buildenv-modules,
 	fink-package-precedence,
 	libxt-flat,
 	openmotif4,
+	openssl110-dev,
 	x11-dev
 <<
-Depends: libxt-flat-shlibs, openmotif4-shlibs, x11-shlibs
+Depends: <<
+	libxt-flat-shlibs,
+	openmotif4-shlibs,
+	openssl110-shlibs,
+	x11-shlibs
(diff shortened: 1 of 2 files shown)


nieder:
Pinging @khorton since it's his.

dmacks (approved):
LGTM (built and dependencies ok on 10.13). Not sure why I'm in this loop, but if $Maintainer doesn't respond soon let's push it.

nieder:
maintainer is 'none ' (used to be @khorton)

PR #621, opened 2020-05-10
Check aquaterm build environment has run XCode.app.

leighsmith:
Added check that XCode.app has been run by checking for presence of /Library/Developer/PrivateFrameworks


nieder:
Closes #618

PR #622, opened 2020-05-11
Added libaubio5 package

leighsmith:
Added libaubio5 package to support v0.4.9 version, which has updated the version of the dylib to 5.4.8.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sound/libaubio2.info
+++ b/10.9-libcxx/stable/main/finkinfo/sound/libaubio2.info
@@ -1,6 +1,6 @@
 Package: libaubio2
 Version: 0.3.2
-Revision: 4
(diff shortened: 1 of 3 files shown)


dmacks:
What is in the -bin splitoff? If it's userland programs, the "new package" break might be a good time to switch to a unified package-name for it. For example, "audio-bin" or "aubio" could be constant rather than users having to switch up to each successive libaubioX-bin as version availability changes.

nieder:
-bin looks like end user programs based on the manfiles.

nieder:
Submitter @leighsmith has said offlist that he's no longer using Fink. I'll work on this so that we're not using a 6+ year old version of this library (although I can't find any package that's using any of our libabioN packages)

PR #599, opened 2020-04-19
GnuTLS 3.6.13

nieder:
Fix build failure with Xcode11.4
Fix test failure from code that expects (but forgets to check for) not-yet-released version of nettle library.

PR #627, opened 2020-05-20
Fix GnuTLS 3.6.13 build when xcode isn't installed

PovlAbrahamsen:
This builds on #599. If xcode isn't installed, but the command-line tools are, the patch still won't be applied correctly. This small change should fix that. Since nothing is changed with the compilation, the revision has not been bumped.


dmacks:
While we're fixing this pkg, there's a bug in #599: the compatibility_version of a .dylib in the Shlibs field was increased but the pkgversion dependency was not updated. The line:
```
%p/lib/gnutls30/libgnutls.30.dylib         58.0.0 %n (>= 3.6.10-1)
```
should have 3.6.13 instead of 3.6.10. That entails a rev-up, so it should be "3.6.10-2" I guess. I'm not going to just push that to master while you're working here, to avoid creating git conflict. If you want to include a fix here, great. If not, I'll push it separately after your work gets merged.

PR #631, opened 2020-05-29
jp2a

htodd:
New package.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/jp2a.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/jp2a.info
@@ -1,27 +1,23 @@
 Package: jp2a
 Version: 1.0.7
-Revision: 1
+Revision: 2
 Maintainer: Hisashi T Fujinaka <htodd@twofifty.com>
 Source: https://github.com/cslarsen/jp2a/archive/v%v.tar.gz
 Source-MD5: 7da403ce64bb6630bbb0002655831ab4
 Source-Checksum: SHA256(e509d8bbf9434afde5c342568b21d11831a61d9942ca8cb1633d4295b7bc5059)
 SourceRename: jp2a-%v.tar.gz
 DocFiles: COPYING README LICENSES BUGS
-BuildDepends: libcurl4, libjpeg9
+BuildDepends: libcurl4, libjpeg9, autoconf2.6,automake1.15
 Depends: libcurl4-shlibs, libjpeg9-shlibs
-Description: jp2a is a simple JPEG to ASCII converter
+Description: Simple JPEG to ASCII converter
+ConfigureParams:  <<
+        --with-jpeg-prefix=%p \
+        --with-curl-config=%p/bin
+<<
 CompileScript: <<
-	autoreconf -vi
-	./configure \
-		--with-jpeg-prefix=%p \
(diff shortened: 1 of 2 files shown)


htodd:
OK, I've really screwed this up. I somehow pushed an intermediate info file to master. Can you guys check this so I can push this over the mistake?

dmacks:
Could you explain what's the "bad" vs the "good" content here? Is 861252a17eab96ed9c0dd632e07daf0d523e736e ("jp2a: new package") what you want and 28162d09cdce78264f25d5d51afccc0800d605d2 ("jpegoptim: new upstream") something that should be ignored?

htodd:
jpegoptim was just a new version. jp2a should be a new package and I don't know how they both got pushed togehter. jp2a is the one I would like checked.

dmacks:
```
Validating package file jp2a.info...
Warning: Description of "jp2a" starts with lower case. (jp2a.info)
Warning: Description of "jp2a" contains package name. (jp2a.info)
```
Along with that, by eye, a few parts of the jp2a.info can be simplified:
```
Description: jp2a is a simple JPEG to ASCII converter
DescDetail: <<
jp2a is a simple JPEG to ASCII converter
<<
```
I can't think of a context where DescDetail is used alone, rather than being accompanied by Description, whereas Description is often used alone. So you can scrap repeating Description in DescDetail (only bother with DescDetail if you have something substantive to add). How about just "Description: Simple JPEG to ASCII converter"?

In CompileScript, you have an explicit "make -j", which seems like it saturates a user's system, overriding the env-var control used by fink. Better to let them decide how much multiprocessing to allocate to fink themselves. That follows a ./configure that includes a --prefix that is fink's defaults anyway and a --mandir that is ./configure's default. Cleaner to just tell fink what is special for this package and let it respect users' system configs. Instead of:
```
CompileScript: <<
        autoreconf -vi
        ./configure \
                --with-jpeg-prefix=%p \
                --with-curl-config=%p/bin \
                --prefix=%p \
                --mandir=%p/share/man
    make -j
<<
```
how about:
```
ConfigureParams:  <<
        --with-jpeg-prefix=%p \
        --with-curl-config=%p/bin
<<
CompileScript: <<
        autoreconf -vi
        %{default_script}
<<
````
If  you want to see the results while you are debugging, you can ```fink dumpinfo -fcompilescript jp2a```

Using autoreconf requires ```BuildDepends: autoconf2.6,automake1.15```

**Comment by htodd** [comment]
I'm using fink -Kk -y -m rebuild to check. I guess I missed any errors. How's this?

**Comment by htodd** [comment]
@dmacks Can you check?

**Comment by dmacks** [comment]
```
checking ncurses/term.h usability... no
checking ncurses/term.h presence... no
checking for ncurses/term.h... no
checking for tgetent in -lncurses... yes
[...]
gcc  -g -O2  -L/sw/lib -L/sw/lib -o jp2a aspect_ratio.o html.o term.o curl.o jp2a.o options.o image.o  -ljpeg -L/sw/lib -lcurl  -lncurses
```

so the package is missing ```BuildDepends:libncurses5``` and ```Depends:libncurses5-shlibs```

htodd:
@dmacks OK, I figured out I needed to check with otool -L and added libncurses5. I think this should be OK.

htodd:
I'm merging this.

Issue #567, opened 2020-01-30
PyQt5 build problems


nieder:
PyQt5 has a couple build problems:

1) has deps on several qt5 -shlibs libraries (such as qt5-mac-qtbluetooth-shlibs), but not on the dev package (qt5-mac-qtconnectivity).
2) misses declaring dependency several qt5 packages that have buildable modules (such as qt5-mac-qtwebengine)
3) Uses old openssl100

(1) f-p-p can be used if `QMAKE_CXXFLAGS="-MD"` is passed to configure.py.
(2) the `--disable=MODULE` flag can be used to specifically disable some modules.
(3) Local tests show that openssl110 will build, but I see no actual linkage or header useage.

TheSin-:
I just built it recently without issues on 10.15, except #3 it cause a build failure trying o switch back to openssl100, but a restart the build and it built and ran perfectly.

PR #635, opened 2020-06-07
build cleanup for PyQt5

nieder:
Fixes the issues with #567
* Remove old inherited OpenSSL dep
* Add f-p-p to help track deps
* clean up deps to fit what's actually being used by the package


nieder:
4 weeks with no response. This is a build cleanup to enforce consistency. No functional changes.

PR #607, opened 2020-04-25
alpine: new upstream

htodd:
Alpine is a bit tricky. I got it to build but could you look it over?

PR #639, opened 2020-06-28
fix build-paths in binary

dmacks:
Resolves bug noted in #607 

Not sure why we're patching both configure and its configure.ac template, but whatever.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/net/alpine.info
+++ b/10.9-libcxx/stable/main/finkinfo/net/alpine.info
@@ -1,7 +1,7 @@
 Info2: <<
 Package: alpine
 Version: 2.22
-Revision: 3
+Revision: 4
 Source: http://alpine.x10host.com/alpine/release/src/alpine-%v.tar.xz
 Source-checksum: SHA256(849567c1b6f71fde3aaa1c97cf0577b12a525d9e22c0ea47797c4bf1cd2bbfdb)
 PatchFile: alpine.patch
@@ -16,6 +16,9 @@ BuildDepends: cyrus-sasl2-dev, libgettext8-dev, libiconv-dev, libncurses5, openl
 GCC: 4.0
 Conflicts: pine, pine-ssl, alpine, re-alpine
 Replaces: pine, pine-ssl, alpine, re-alpine
+# local-support-info and pinerc paths are set explicitly because the
+# autoconf defaults are based on --prefix, which is set to %d (rather
+# than %p) for other reasons
 ConfigureParams: <<
 	--prefix=%d \
 	--bindir=%i/bin \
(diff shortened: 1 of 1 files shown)


nieder:
```
$ strings bin/alpine
...
/sw/build.build/root-alpine-2.22-4/sw/share/locale
...
```
(where `%p/build.build` is my root build dir). Current %r=3 had 4 lines there, so this is an improvement.
The remaining one is because we have `--datarootdir=%i/share` in `%c`. But setting that to `%p/share` makes it try to install directly to %p during InstallScript.


```
	--disable-debug \
	--with-ssl-dir=%p \
	--with-ssl-certs-dir=%p/etc/ssl/certs \
	--with-ldap-dir=%p \
	--with-local-password-cache-method \
	--without-pubcookie \
	--without-tcl
```
This %c seems more robust to me (alongside `make install DESTDIR=%d`) because it minimizes the number of hardcoded paths. And it doesn't need the non-standard `--prefix=%d`. I don't know what the original reasons are for setting `prefix=%d`.

dmacks:
--prefix=%d is *really* weird...no way it sanely does the Right Thing except possibly to work around some other unstated weird autoconf detection. The ssl path detection in autoconf is a huge mess and still might not be deterministic (detection of things outside of [apple+fink] paths).

cooljeanius (approved):

nieder (changes requested):
binary still has build path strings:
```
$ strings bin/alpine | grep build.build
/sw/build.build/root-alpine-2.25-2/sw/share/locale
'--prefix=/sw' '--prefix=/sw/build.build/root-alpine-2.25-2' '--bindir=/sw/build.build/root-alpine-2.25-2/sw/bin' '--sbindir=/sw/build.build/root-alpine-2.25-2/sw/sbin' '--datarootdir=/sw/build.build/root-alpine-2.25-2/sw/share' '--disable-debug' '--with-ldap-dir=/sw' '--with-local-password-cache-method' '--with-ssl-certs-dir=/sw/etc/ssl/certs' '--with-ssl-dir=/sw' '--without-pubcookie' '--without-tcl' '--with-local-support-info=/sw/lib/pine.info' '--with-system-pinerc=/sw/lib/pine.conf' '--with-system-fixed-pinerc=/sw/lib/pine.conf.fixed' 'LDFLAGS=-lintl -L/sw/lib' 'CPPFLAGS=-DMAC_OSX_KLUDGE -DUSE_INTERP_RESULT -I/sw/include'
```
```
$ grep -r build.build sw-4210fc1
grep: sw-4210fc1/bin/alpine: binary file matches
grep: sw-4210fc1/bin/rpdump: binary file matches
grep: sw-4210fc1/bin/pico: binary file matches
grep: sw-4210fc1/bin/rpload: binary file matches
grep: sw-4210fc1/bin/pilot: binary file matches
```

nieder:
@htodd alpine has moved to https://alpineapp.email/ and there's a new release there 2.26

cooljeanius:
> @htodd alpine has moved to [alpineapp.email](https://alpineapp.email/) and there's a new release there 2.26

Yeah, MacPorts has version 2.26:
```
$ port installed alpine
The following ports are currently installed:
  alpine @2.26_0+kerberos+ldap+ssl+tcl (active)
```

PR #640, opened 2020-06-30
Setuptools-py update to 47.3.1

nieder:
Update setuptools to the latest upstream (including separate last versions for py27 and py34).
Still can't enable tests because of circular dependencies on pytest (and a couple missing modules we don't have).
Tested using the new setuptools-tng-py37 on several pymods


nieder:
Ping. Will commit this soon.

PR #641, opened 2020-07-04
OpenBLAS update to 0.3.10 and new fix for missing CLOCK_REALTIME

dhomeier:
New upstream version with two additional changes:

1. I got another report on a build failure of the entire benchmark suite on 10.11 due to missing CLOCK_REALTIME etc. support in `time.h`; judging from [this comment](https://github.com/Tarsnap/tarsnap/issues/286#issuecomment-331666331) this is systemic to Xcode 8 and earlier. Trying to work around here with `mach/clock_types.h` definitions, but yet to be tested on 10.11.
EDIT: The failure occurs for 6 of the benchmarks that have been newly added or changed to use `clock_gettime`; see comment for resolution.

2. Switched the benchmark build to dynamically linked binaries, since the `*.goto` executables were growing well beyond 1 GB in total. No obvious performance differences to the statically linked versions.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/openblas.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4 -*-
 Info2: <<
 Package: openblas
-Version: 0.3.9
+Version: 0.3.10
 Revision: 1
 Type: gcc (10)
 Maintainer: Derek Homeier <dhomeie@gwdg.de>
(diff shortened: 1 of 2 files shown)


dhomeier:
Since the benchmarks in question (predictably) still failed on linking without `clock_gettime` available, `CompileScript` now just tries to build one of them, and on failure removes them from the target list. I tried to use the substitute implementation from `smallscaling.c`, but found it to have apparently too low resolution for those routines, so decided it's not worth the effort.
Now successfully built on 10.11, thanks to David Neuhaus for testing and helping to debug!

nieder:
Is it just failure on 10.11, or 10.11 and lower?
Wouldn't it be simpler to test for the system version, then remove benchmarks in question if the No-clock_gettime condition is met?

dhomeier:
> Is it just failure on 10.11, or 10.11 and lower?
> Wouldn't it be simpler to test for the system version, then remove benchmarks in question if the No-clock_gettime condition is met?

I haven't had any reports on earlier versions, but according to the comments on the Tarsnap link this is missing on all OS X <= 10.11. It does mention an Xcode update

> OSX 10.11 with current XCode: they ship a `<time.h>` aimed at 10.12, so this does include `clock_gettime()`

but I don't know if that is a 9.x version or how to get it on 10.11. Anyway there is not much harm in stripping those targets unconditionally for all `$XCODE_VERSION` < 9.0 or  `darwin_version -lt 16`.

nieder:
Does the lack of clock_gettime affect normal runtime at all in systems without clock_gettime? Or is it that the benchmarks expect clock_gettime without shielding the function call, while the library shields it correctly?

dhomeier:
> Does the lack of clock_gettime affect normal runtime at all in systems without clock_gettime? Or is it that the benchmarks expect clock_gettime without shielding the function call, while the library shields it correctly?

I think the library itself is not using any of those timing routines; at least building it was never an issue without clock_gettime support.

PR #648, opened 2020-07-21
New upstream version CRAN

babayoshihiko:
New upstream versions

RhpcBLASctl:  The license is AGPL.  Is it one of GPL variations?


nieder:
For AGPL packages, either GPL3 or OSI-Approved would be close enough. That's what other packages are using. Would be easy enough to add it into Fink if requested.

babayoshihiko:
1 and 2 fixed.

PR #653, opened 2020-07-26
flex-devel update to 2.6.4 and PDF doc build

dhomeier:
New upstream version, added the PDF manual to the docs – this does increase the installed size by about 1 MB.

On a related issue, does anyone recall details about the compatibility problems of the newer version vs. the standard 2.5.4a package? I.e. is Fink stuck with defaulting to a quarter-century old (alpha?) version that no longer has an original source location? There is a "final" version 2.5.39 of the 2.5 branch, but as `flex-devel` originally seems to have branched off with 2.5.35, I presume any issues that version had would be present in 2.5.39 as well... There are also some older versions on https://github.com/westes/flex/releases (but not quite as old as 2.5.4), but it's unclear whether these are alpha, beta or gamma versions, and none of them even has a working `configure` (attempts to generate a new build with `autogen.sh` failing early on missing `Makefile.in`...).
OTOH the only packages I could spot that explicitly depend on `flex` (rather than `flex-devel` or listing it as a `BuildConflict` / pointing to the system `/usr/bin/flex`) are
`apple-gdb biopython-py27 cscope detex ginac2 pcb`
and I just successfully rebuilt all 6 against a default `flex` updated to 2.5.39`...

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/flex-devel.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/flex-devel.info
@@ -1,19 +1,20 @@
 Package: flex-devel
-Version: 2.6.0
+Version: 2.6.4
 Revision: 1
 BuildDepends: <<
 	fink (>= 0.28.0-1),
 	fink-package-precedence,
 	libgettext8-dev,
-	help2man
+	help2man,
+	texinfo
 <<
 Depends: libgettext8-shlibs
 Description: Fast lexical analyser generator
 License: BSD
 Maintainer: None <fink-devel@lists.sourceforge.net>
 
-Source: mirror:sourceforge:flex/flex-%v.tar.bz2
-Source-MD5: 266270f13c48ed043d95648075084d59
+Source: https://github.com/westes/flex/releases/download/v%v/flex-%v.tar.gz
+Source-MD5: 2882e3179748cc9f9c23ec593d6adc8d
 
 UseMaxBuildJobs: false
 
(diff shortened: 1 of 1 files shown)


nieder:
Is it worth sending the PDF docs to a new SplitOff flex-devel-docs?
Also, I'm curious as to why flex-devel deletes the man and info files. This traces back to the original flex-devel commit 17 (!) years ago: https://github.com/fink/fink-distributions/commit/9ebcee5b484809fe7dd48fff38c82532cae0fb71

Do those 6 packages build with flex-devel-2.6.x? Do we want to make a hidden flex-legacy (to match bison-legacy and autoconf2.13-legacy) just in case, update 'flex' to the latest, and ditch flex-devel?

dhomeier:
> Is it worth sending the PDF docs to a new SplitOff flex-devel-docs?

For saving install space if it's not wanted? Not sure, now it has 1.3 MB installed docs, ½ MB of that is the PDF, ⅓ the Changelog and another ⅓ the `examples` directory I also just added. Info and localisation make for another ⅓ MB each; I think the best way to save some space would be to xz the Changelog for a start.

> Also, I'm curious as to why flex-devel deletes the man and info files. This traces back to the original flex-devel commit 17 (!) years ago: [9ebcee5](https://github.com/fink/fink-distributions/commit/9ebcee5b484809fe7dd48fff38c82532cae0fb71)
> 
I guess they were considered of little use when they are not found in their non-standard locations. I left them now with the usage note how to find them so they would at least be available, but Fink was not even asking for an `InfoDocs` entry. A better solution would certainly be to rename them an install to the standard paths, so `man flex-devel` (or `man flex-legacy` to come) would simply show the right version.

> Do those 6 packages build with flex-devel-2.6.x? Do we want to make a hidden flex-legacy (to match bison-legacy and autoconf2.13-legacy) just in case, update 'flex' to the latest, and ditch flex-devel?

Yes, I have also successfully built all 6 with 2.6.4 installed as the standard flex.
Retiring flex-devel would require pointing the 9, if I counted correctly, packages depending on it to the new flex, but since it is a `BuildDependsOnly` with only static libs, that should be fairly painless. +1 from me for that.

nieder:
I think this is the most well-behaved route to take:
1) new package flex-legacy v2.5.4a
2) flex becomes 2.6.4
3) flex-devel becomes a dummy obsolete package that deps on flex (>= 2.6.4)
4) Approx 9 packages BDepping on flex-devel get transitioned to flex
5) Can then kill flex-devel. Or we could kill flex-devel directly and just transition its users on the same commit.

nieder:
@dhomeier ping. Do you want me to take this and push through the migration plan from [this comment](https://github.com/fink/fink-distributions/pull/653#issuecomment-687087866)?

nieder (approved):
Let's at least get this in and let it percolate for a bit, then we can move this %v to replace flex

PR #658, opened 2020-08-08
relax: bump all possible dependencies.

nieder:
Not compatible with gmt > v4
Part of #481
@sbarbot


nieder:
Pushing because no response. Package builds, starts, and these are minor changes needed to trim old fat.

nieder:
Sorry, that's a typo in the commit comment. It should have said that it's incompatible with the gmt library install_name > 4. So gmt-4.5.18 (libgmt.**4**.dylib) is OK, but gmt5-5.4.5 (libgmt.**5**.dylib) is not compatible.
So I didn't touch the gmt dep (only gcc, netcdf, and proj).

PR #668, opened 2020-09-30
Perl5282

nieder:
This creates the new package perl5282, to match system perl on macOS 11.0. It's basically a copy of perl5184, with some necessary test fixes grabbed from upstream.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/perl5162.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/perl5162.info
@@ -1,12 +1,12 @@
 Package: perl5162
 Version: 5.16.2
-Revision: 103
+Revision: 104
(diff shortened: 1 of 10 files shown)


nieder:
One thing that might need editing is the list of provided packages. How was that original list generated? Is there a perl command to output the available mods?

dmacks:
'corelist' (via module-corelist-pmXXX) can help. I keep Fink::VirtPackage's system-perlXXX provides: lists up-to-date. We generally haven't bothered adding to these lists, only removing (among other reasons, they are so incomplete and always have been).

nieder:
Yeah, so `corelist -v 5.14.4` says there are 689 modules provided by CORE. So not going to update.

nieder:
Ping on committing?

nieder:
Will push soon to start populating pm5282 mods

PR #675, opened 2020-10-14
Qt5 and PyQt updates to 5.15

dhomeier:
The current PyQt5 which I am building as a backend for Matplotlib update requires an update to much of the Qt5 ecosystem. The latest versions of all packages, where available, that are _needed_ for pyqt5-mac-pyNN-5.15.1-1 are added in the first commit, the second has the update for PyQt5 and for SIP 4.19.24, also required for the build.
SIP 4 is also the last version to support PyQt4. This makes updating to SIP 5 potentially a bit risky, but the third commit adds build support for the latter (introducing `update-alternatives`-managed binaries); but building pyqt4-mac-py afterwards will require downgrading sip-pyNN-bin to the earlier version (Qt4 has reached its end of life anyway).
Also updated the remaining Qt5 Depends of PyQt5 to 5.15.1. qt5-mac-enginio-shlibs and qt5-mac-qtdeclarative-shlibs turned out to be no longer required - the former is among the discontinued submodules in Qt5 > 5.10.


nieder:
I've not updated Fink's Qt5 past 5.7 because newer releases don't work on older macOS that we still support. Qt5.15 only works on macOS 10.13 and higher.

See my qt5.15 branch: https://github.com/nieder/fink-distributions/tree/qt5.15/10.9-libcxx/stable/main/finkinfo/libs . So in my branch, I've moved it into a separate %p/lib/qt515-mac prefix so that they can coexist (since Qt frameworks don't do any versioning themselves. boo!)

Issue #671, opened 2020-10-05
upgrade matplotlib-py


nieder:
Our current matplotlib-py-1.5.3 is very old (sep 9, 2016)

Latest release is 3.3.2 (Sep 15, 2020).
The latest version that supports py27 is 2.2.5 (Feb 2, 2020).

Currently, 1.5.3 fails building the qt5agg backend (unclear if our Qt5-5.7 is too old or too new for v1.5.3).

Newer matplotlib packages will need some sort of update to (at least)
* qhull (at least 2015.2)
* wxgtk
* pycairo

This is a tracking bug mostly to remember what needs to be done
@schwehr

dmacks:
qhull is @akhansen and at least the current packages of it have no external deps, so should be an ~easy one-off update.

pycairo needs a newer cairo, but that's only on top of the freetype/fontconfig stack, not glib/gtk.

nieder:
Just tested, and qhull-2020.2 is a simple cmake build with libN=8.

I think the pycairo updates are only needed for specific frontends.

schwehr:
I haven't put any time into fink or even used it in more than 6 months.  Anyone is welcome to take any of my packages.

nieder:
See #672 for qhull update

dhomeier:
I've got 3.3.2 to build with Qt4[Agg], Qt5[Agg], Tk[Agg], WX[Agg], PDF, PS, SVG and Agg backends, using the PyQt5 update in #675 (bringing a bunch of further Qt5 updates with it), and our wxpython400 for WX. Qt4 support will be discontinued within the next 2 minor versions, so it would not be worth much effort on debugging any potential build issues. 
Will open a PR with the remaining -py dependencies later.

nieder:
@dhomeier while the qt5.15 stuff is figured out to get 3.3.2, I've tried to get 2.2.5 (last one to support py27). Should I open a PR for that to at least not have an ancient matplotlib? I had to turn off qt5agg and wxagg since they were failing (probably requires newer than qt5.7), and I'm getting a failing test that seems to be a python issue.

dhomeier:
@nieder I have also built `matplotlib-py27-2.2.5` with qt5agg, but without wxagg, but that was with the qt5.15 update. Trying to rebuild with the existing qt5 now, but found that even enabling Qt4 needed at least an update of the sip-py*.
[...]

PR #678, opened 2020-10-20
Matplotlib 2 and 3 updates

dhomeier:
Moving this into its own PR to include updates and suggestions from #671 and #677 plus enabling tests and wxpython400/pyqt4-mac 4.12/pyqt5-mac 5.11 backends for matplotlib 2.2.5 (py27), 3.0.3 (py35) and 3.3.2 (py36+).

To upgrade the SIP libraries to the required version without having to update Qt5 to > 5.7, the `sip-pyqt[45]` versions have been bumped beyond those of `sip-py[-bin]`, which need to stay at <~ 4.19.17 to still build pyqt5-mac against qt5-5.7.

Taking up the discussion from https://github.com/fink/fink-distributions/issues/671#issuecomment-711152458:

> Regarding your 2.2.5-py27:
you're building a local freetype.dylib that has the same install_name as Fink's so that the *.so modules use the ft-2.6.1 library during tests (to get matching results), but once installed the *.so link paths find Fink's FT library in %p/lib/ft219/lib ? I think that's what's happening, but just want to confirm.
But FT-2.6.1 should only be a TestSource. We don't want to download it when someone is not running tests.

No, afaics the `ft2font.cpython-NN-darwin.so` that is built and installed does not point to any of Fink's FT installation (which is at 2.6.3), but provides the full required subset of functions itself.
```
/sw/lib/python3.9/site-packages/matplotlib/ft2font.cpython-39-darwin.so:
	/usr/lib/libz.1.dylib (compatibility version 1.0.0, current version 1.2.11)
	/sw/lib/libbz2.1.dylib (compatibility version 1.0.1, current version 1.0.6)
	/sw/lib/libpng16.16.dylib (compatibility version 54.0.0, current version 54.0.0)
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 400.9.4)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 1252.250.1)
```
So I guess we cannot simply remove it and drop Fink's 2.6.3 in as replacement – probably would have to rebuild and link against the system freetype after the test phase, but that would seem to defeat the purpose of the tests to some extent, not testing the actual version that is getting installed in the end.
So the FT-2.6.1 source is used unconditionally (it is small compared to the full matplotlib source), not sure if linking against Fink's slightly newer version and getting rid of a ~1 MB .so would justify the changes in setup.

> Instead of the rsync magic, does %p/bin/python%type_raw[python] setup.py develop --root=... work? Might clean up the way the build process is ordered (testing in InstallScript, etc).

Yes; while it cannot install to a root inside %i during the test phase, choosing another path inside %b just creates some .pth files and directories pointing to the actual tree created in the compile phase, so it is testing pretty much what is eventually being installed. For 3.3.2 I still needed pointers to the `baseline_images` dirs, and `mpl_toolkits` apparently is inevitably first looked up in `%p/lib/python%type_raw[python]/site-packages` once installed, but the rest of the tests are passing with a few identified failures.

> Missing pytest-mpl-py package.

I reviewed the TestDepends and found it is not actually needed nor changes the number of enabled tests in any way.

> Are the BuildConflicts against qhull*-dev needed? I don't see a difference in the build process or the final _qhull.so module.

The qhull packages are irrelevant for the build since matplotlib only looks for the non-reentrant version and thus will now build its own local copy. It seems there were conflicts with Fink's reentrant libraries installed, but I checked that with the new qhull8.0-dev at least this is no longer the case. I have not tried again with the older versions, so I left the BuildConflicts in place for now. But if you did not find any differences with your 2.2.5 builds, it's probably safe to remove them.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/matplotlib-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/matplotlib-py.info
@@ -16,14 +16,14 @@ Depends: <<
  atk1-shlibs (>= 1.32.0-1),
  cairo-shlibs (>= 1.12.14-1),
  cycler-py%type_pkg[python] (>= 0.10.0-1),
- dateutil-py%type_pkg[python],
+ dateutil-py%type_pkg[python] (>= 2.7.3-1),
  fontconfig2-shlibs (>= 2.10.0-1),
  glib2-shlibs (>= 2.22.0-1),
  gtk+2-shlibs (>= 2.18.0-1),
- kiwisolver-py%type_pkg[python],
+ kiwisolver-py%type_pkg[python] (>= 1.2.0-1),
  libpng16-shlibs,
  pango1-xft2-ft219-shlibs (>= 1.24.5-4),
- pil-py%type_pkg[python],
+ pil-py%type_pkg[python] (>= 6.2.0-1),
  poppler-bin,
  python%type_pkg[python],
  pyparsing-py%type_pkg[python],
@@ -62,8 +62,7 @@ BuildDepends: <<
 Recommends: <<
  ffmpeg,
  imagemagick,
- ipython-py%type_pkg[python],
- pil-py%type_pkg[python]
+ ipython-py%type_pkg[python]
 <<
 
 BuildConflicts: libharfbuzz0-dev, qhull5-dev, qhull6-dev, qhull6.3.1-dev
@@ -77,45 +76,46 @@ SetLDFLAGS: -L%p/lib -lz -lbz2 -lpng16
 UseMaxBuildJobs: True
 
 PatchFile: %{ni}.patch
-PatchFile-MD5: 4da6d6a9685122fda5b2f2bf3481820d
+PatchFile-MD5: 98e8db5dd66b9f68f9ad19e7fb5a01ee
 PatchScript: sed 's|@PREFIX@|%p|g;s|@PY_PKG@|%type_pkg[python]|g;s|@PY_RAW@|%type_raw[python]|g' < %{PatchFile} | patch -p1
 
 CompileScript:  <<
  #!/bin/sh -ev
  cp -a setup.cfg{.template,}
- # build against a local copy of freetype-2.6.1 to ensure exact version match
- # (set via SetMFLAGS in TestInfo; should we set this unconditionally?):
+ # Build against a local copy of freetype-2.6.1 to ensure exact version match
  mkdir build
  mv ../freetype-2.6.1 build
  export MPLLOCALFREETYPE="$MFLAGS"
  export PKG_CONFIG_PATH="%p/lib/glib-2.0/pkgconfig-strict:$PKG_CONFIG_PATH"
  OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES %p/bin/python%type_raw[python] setup.py build
  fink-package-precedence --depfile-ext='\.d' build
- rsync -aC lib/matplotlib/tests/{tinypages,*.ipynb,*.pfb,*.ttf} build/lib.macosx-*-%type_raw[python]/matplotlib/tests
 <<
 
 InfoTest: <<
  TestDepends: <<
+  certifi-py%type_pkg[python] (>= 2019.11.28-1),
   ffmpeg,
   inkscape,
-  pytest-py%type_pkg[python]
+  numpy-py%type_pkg[python] (>= 1.18.1-1),
+  pyparsing-py%type_pkg[python] (>= 2.4.5-1),
+  pytest-py%type_pkg[python] (>= 3.6.0)
  <<
  # test_backends_nbagg checks for nbconvert but fails with older versions.
  TestConflicts: nbconvert-py%type_pkg[python] (<= 6.0.0-1)
- #SetMFLAGS: True
- #TestConfigureParams: --local_freetype=True
  TestScript: <<
-  # conftest.py throws import errors when trying to run tests anywhere but in the
-  # directory matplotlib is imported from - remove them again before installation,
-  # as they would also triple install size.
-  rsync -aC lib/matplotlib/tests/baseline_images build/lib.macosx-*-%type_raw[python]/matplotlib/tests
-  rsync -aC lib/mpl_toolkits/tests/baseline_images build/lib.macosx-*-%type_raw[python]/mpl_toolkits/tests
-  # warnings if flaky extension is not installed, but with it some tests fail on
-  # passing a 'reruns' kwarg that is unknown to any version of flaky 3.x.
-  perl -pi -e 's|(mark.flaky.)(reruns=)(.)|${1}max_runs=${3}, min_passes=2|;' build/lib.macosx-*-%type_raw[python]/matplotlib/tests/test*.py
-  # pytest can end up importing mpl_toolkits from %p/lib/python* (without tests)
-  # regardless of sys.path unless running from the eventual install dir in %i,
-  # so defer actual running of the tests to Install phase.
+  #!/bin/sh -ev
+  export LANG=en_US.UTF-8
+  export PYTHONPATH=%b/build/lib/python%type_raw[python]/site-packages
+  mkdir -p $PYTHONPATH
+  OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES %p/bin/python%type_raw[python] setup.py develop --no-deps --prefix %b/build
+  # Comparison images are expected in `Path(__file__).parent / "baseline_images"`
+  # - mpl_toolkits still may not find them if matplotlib is already installed in system path.
+  ln -s ../../../../lib/matplotlib/tests/baseline_images build/lib.macosx*-%type_raw[python]/matplotlib/tests
+  ln -s ../../../../lib/mpl_toolkits/tests/baseline_images build/lib.macosx*-%type_raw[python]/mpl_toolkits/tests
+  PYTHONPATH="${PYTHONPATH}:$(ls -d %b/build/lib.macosx*-%type_raw[python])"
+  %p/bin/python%type_raw[python] -Bm pytest lib/mpl_toolkits || echo 'mpl_toolkits tests not working with installed matplotlib'
+  %p/bin/python%type_raw[python] -Bm pytest -k 'not (_pdflatex or _rcupdate or _marker_clipping or _interactive_backend or (_mixedmode and pdf))' lib/matplotlib || exit 2
+  rm build/lib.macosx*-%type_raw[python]/m*/tests/baseline_images
  <<
  TestSuiteSize: large
 <<
@@ -125,13 +125,6 @@ InstallScript: <<
  export PKG_CONFIG_PATH="%p/lib/glib-2.0/pkgconfig-strict:$PKG_CONFIG_PATH"
  OBJC_DISABLE_INITIALIZE_FORK_SAFETY=YES %p/bin/python%type_raw[python] setup.py install --root %d
  export LANG=en_US.UTF-8 PYTHONPATH=%i/lib/python%type_raw[python]/site-packages
- # Check if tests have been installed and setup (for `--tests`)
- if [ -d ${PYTHONPATH}/matplotlib/tests/baseline_images ]; then
-  cd $PYTHONPATH
-  # test_backend_tk might fail on accessing display
-  %p/bin/python%type_raw[python] -Bm pytest -k 'not (_pdflatex or _rcupdate or _marker_clipping or _interactive_backend or (_mixedmode and pdf))' || exit 2
-  rm -r m*/tests/baseline_images result_images .pytest_cache
- fi
 <<
 
 License: OSI-Approved
@@ -161,7 +154,7 @@ DescPackaging: <<
  falling back to the local copy until mpl is converted to libqhull_r or we
  find a way to install the non-reentrant dylib with libqhull-shlibs as well.
  (cf. http://www.qhull.org/html/qh-code.htm#reentrant .)
- The 'save to file' interface seems broken with qt4 and ipython.
+ Qt4 backend to be phased out and already unsupported with ipython --pylab.
  Enabled build against wxpython400, requires patch to ipython-py 7.5.0 for
  pylab environment; RuntimeError in backends/backend_wx.py on exit.
  [ Added patch from https://github.com/matplotlib/matplotlib/pull/14292 - merged in 3.1.1]
@@ -174,9 +167,10 @@ DescPackaging: <<
  - workaround taken from the same site.
  Test suite requires exact version match of freetype built against (2.6.1),
  set by local_freetype=True in setup.cfg, to ensure correct image comparison.
- Enforced here by MPLLOCALFREETYPE set with a hack from SetMFLAGS in InfoTest
- and always used in build against local source copy.
+ Always used in build against local source copy - this will create and link
+ against a packaged ft2font.cpython-*-darwin.so.
  Patched tests to make them use matching versions of [i]python and jupyter.
+ Changed 'reruns' kwarg to 'max_runs' to keep flaky 3.x happy.
 <<
 Homepage: http://matplotlib.sf.net
 DescUsage: <<
--- a/10.9-libcxx/stable/main/finkinfo/languages/pyqt5-mac-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/pyqt5-mac-py.info
@@ -80,8 +80,8 @@ BuildDepends:<<
   qt5-mac-qtwebkit (>= 5.6.0-1),
   qt5-mac-qtwebsockets (>= 5.6.0-1),
   qt5-mac-qtxmlpatterns (>= 5.6.0-1),
-  sip-py%type_pkg[python] (>= 4.19.13-2),
-  sip-py%type_pkg[python]-bin (>= 4.19.13-2)
+  sip-py%type_pkg[python] (>= 4.19.13-2), sip-py%type_pkg[python]  (<= 4.19.17-1),
+  sip-py%type_pkg[python]-bin (>= 4.19.13-2), sip-py%type_pkg[python]-bin (<= 4.19.17-1)
 <<
 
 UseMaxBuildJobs: true
--- a/10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt4-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt4-py.info
@@ -2,24 +2,31 @@
 Info2: <<
 
 Package: sip-pyqt4-py%type_pkg[python]
-Version: 4.19.13
-Revision: 2
+Version: 4.19.24
+Revision: 1
 Type: python (2.7 3.4 3.5 3.6 3.7)
 
-Source: mirror:sourceforge:pyqt/sip/sip-%v/sip-%v.tar.gz
-Source-Checksum: SHA256(e353a7056599bf5fbd5d3ff9842a6ab2ea3cf4e0304a0f925ec5862907c0d15e)
+# SF source has not been updated since 4.19.13
+# Source: mirror:sourceforge:pyqt/sip/sip-%v/sip-%v.tar.gz
+Source: https://www.riverbankcomputing.com/static/Downloads/sip/%v/sip-%v.tar.gz
+Source-Checksum: SHA256(edcd3790bb01938191eef0f6117de0bf56d1136626c0ddb678f3a558d62e41e5)
 
 GCC: 4.0
 Depends: python%type_pkg[python]
 BuildDepends: fink-buildenv-modules (>= 0.1.3-1), fink (>= 0.24.12-1)
 
 PatchFile: sip-py.patch
-PatchFile-MD5: d53f5e4456cca0104bdd5a5c7d0dfa99
+PatchFile-MD5: d08dcb0733f7f0bc886672b22777dbf9
 
 UseMaxBuildJobs: true
 CompileScript: <<
 	#!/bin/sh -ev
 	. %p/sbin/fink-buildenv-helper.sh
+ 	if [ %type_pkg[python] -lt 38 ]; then
+		LFLAGS_PY="`%p/bin/python%type_raw[python]-config --ldflags`"
+	else
+		LFLAGS_PY="-L%p/lib `%p/bin/python%type_raw[python]-config --embed --ldflags`"
+	fi
  	%p/bin/python%type_raw[python] configure.py \
 		--sip-module=PyQt4.sip \
 		--no-dist-info \
@@ -27,8 +34,9 @@ CompileScript: <<
 		-p macx-g++ \
 		-v %p/share/sip-py%type_pkg[python] \
 		INCDIR_OPENGL=$X11_INCLUDE_DIR \
+		LFLAGS_PLUGIN="-bundle $LFLAGS_PY"
 		LFLAGS_PLUGIN="-bundle `%p/bin/python%type_raw[python]-config --ldflags`"
-    perl -pi -e 's,exe = \"python.*,exe = \"%p/bin/python\",g' sipconfig.py
+ 	perl -pi -e 's,exe = \"python.*,exe = \"%p/bin/python\",g' sipconfig.py
 	make
 <<
 
@@ -36,7 +44,7 @@ InstallScript: <<
 	#!/bin/bash -ev
 	make install DESTDIR=%d
 <<
-DocFiles: LICENSE* NEWS README
+DocFiles: LICENSE* NEWS README ChangeLog
 
 DescPackaging: <<
 	Disable requirement of framework-built python on darwin.
--- a/10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt5-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt5-py.info
@@ -2,12 +2,14 @@
 Info2: <<
 
 Package: sip-pyqt5-py%type_pkg[python]
-Version: 4.19.13
-Revision: 2
+Version: 4.19.24
+Revision: 1
 Type: python (2.7 3.4 3.5 3.6 3.7)
 
-Source: mirror:sourceforge:pyqt/sip/sip-%v/sip-%v.tar.gz
-Source-Checksum: SHA256(e353a7056599bf5fbd5d3ff9842a6ab2ea3cf4e0304a0f925ec5862907c0d15e)
+# SF source has not been updated since 4.19.13
+# Source: mirror:sourceforge:pyqt/sip/sip-%v/sip-%v.tar.gz
+Source: https://www.riverbankcomputing.com/static/Downloads/sip/%v/sip-%v.tar.gz
+Source-Checksum: SHA256(edcd3790bb01938191eef0f6117de0bf56d1136626c0ddb678f3a558d62e41e5)
 
 GCC: 4.0
 Depends: python%type_pkg[python]
@@ -15,12 +17,17 @@ BuildDepends: fink-buildenv-modules (>= 0.1.3-1), fink (>= 0.24.12-1)
 Replaces: sip-py%type_pkg[python] (<= 4.19.13-1)
 
 PatchFile: sip-py.patch
-PatchFile-MD5: d53f5e4456cca0104bdd5a5c7d0dfa99
+PatchFile-MD5: d08dcb0733f7f0bc886672b22777dbf9
 
 UseMaxBuildJobs: true
 CompileScript: <<
 	#!/bin/sh -ev
 	. %p/sbin/fink-buildenv-helper.sh
+ 	if [ %type_pkg[python] -lt 38 ]; then
+		LFLAGS_PY="`%p/bin/python%type_raw[python]-config --ldflags`"
+	else
+		LFLAGS_PY="-L%p/lib `%p/bin/python%type_raw[python]-config --embed --ldflags`"
+	fi
  	%p/bin/python%type_raw[python] configure.py \
 		--sip-module=PyQt5.sip \
 		--no-dist-info \
@@ -28,8 +35,8 @@ CompileScript: <<
 		-p macx-g++ \
 		-v %p/share/sip-py%type_pkg[python] \
 		INCDIR_OPENGL=$X11_INCLUDE_DIR \
-		LFLAGS_PLUGIN="-bundle `%p/bin/python%type_raw[python]-config --ldflags`"
-    perl -pi -e 's,exe = \"python.*,exe = \"%p/bin/python\",g' sipconfig.py
+		LFLAGS_PLUGIN="-bundle $LFLAGS_PY"
+ 	perl -pi -e 's,exe = \"python.*,exe = \"%p/bin/python\",g' sipconfig.py
 	make
 <<
 
@@ -37,7 +44,8 @@ InstallScript: <<
 	#!/bin/bash -ev
 	make install DESTDIR=%d
 <<
-DocFiles: LICENSE* NEWS README
+
+DocFiles: LICENSE* NEWS README ChangeLog
 
 DescPackaging: <<
 	Disable requirement of framework-built python on darwin.
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/asn1crypto-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/asn1crypto-py.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4 -*-
 Info2: <<
 Package: asn1crypto-py%type_pkg[python]
-Version: 0.24.0
+Version: 1.4.0
 Revision: 1
 Type: python (2.7 3.4 3.5 3.6 3.7 3.8)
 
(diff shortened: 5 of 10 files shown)


dhomeier:
wxpython 4.0.0 did not build for Python 3.8, so I have updated it to 4.1.0 along with the py38 version. I have also removed pytest from its runtime Depends, as there is not a single import from it in the Python modules.

nieder:
I can't find any linking to local-ft-2.6.1 anymore either. I could have sworn I did, but the only linking to ft is to Fink's 2.6.3 in _backend_gdk.so and _gtkagg.so and their linking commands make it clear they're linking to the real dylib in %p. &lt;shrug&gt;

dhomeier:
With a different version linked this should show up fairly quickly in the tests, as the created figures are indeed subtly different.
Of course this should not affect everyday use, but I still have a better feeling testing against the exact build that is actually going to be installed.

nieder:
wxpython400 update changes the install_name of the dylibs, so need new package name. wxpython410 perhaps? It's a shame that it's already using the package version rather than the SO number as the package name, but it'll be more trouble and confusing to change the pattern.
The new wxpython410 should be able to coexist with wxpython400 if I'm reading the changes correctly.

nieder:
Missing sip-pyqt4-py.patch and sip-pyqt5-py.patch

nieder:
Also missing the new TestDepends for the updated tornado-py (constantly-py and twisted-py)

dhomeier:
> Missing sip-pyqt4-py.patch and sip-pyqt5-py.patch

They are still using `sip-py.patch`, also still work with the updated source.

> wxpython400 update changes the install_name of the dylibs, so need new package name. wxpython410 perhaps? It's a shame that it's already using the package version rather than the SO number as the package name, but it'll be more trouble and confusing to change the pattern.

Uh, yes, I was hoping to avoid having yet another versioned package name, but for anything linking to the shlibs there is obviously no way around it.
With the weird SO numbering having wxpython500 succeeded by wxpython100 would possibly create even more confusion though.

> The new wxpython410 should be able to coexist with wxpython400 if I'm reading the changes correctly.

Yes, the -dev and -shlibs packages install all version-numbered files. I've added `Conflicts: wxpython410-py%type_pkg[python]` to wxpython400 as well, but `Replaces` only to wxpython410 – does that make sense?

nieder:
For the sip-pyqt#-py.info patch:
```diff
$ git diff sip-pyqt4-py.info (abridged output)
-Version: 4.19.13
+Version: 4.19.24
...
 PatchFile: sip-py.patch
-PatchFile-MD5: d53f5e4456cca0104bdd5a5c7d0dfa99
+PatchFile-MD5: d08dcb0733f7f0bc886672b22777dbf9
```
```
$ fink validate sip-pyqt4-py.info
Validating package file sip-pyqt4-py.info...
Error: PatchFile-Checksum does not match PatchFile checksum. (sip-pyqt4-py.info)
Expected: d08dcb0733f7f0bc886672b22777dbf9
Actual: MD5(d53f5e4456cca0104bdd5a5c7d0dfa99)
```
I don't see an updated sip-py.patch (accidentally called them sip-pyqt4-py.patch and sip-pyqt5-py.patch) in the changed files list but the .info is looking for a changed file.

dhomeier:
> I don't see an updated sip-py.patch

Oh, my mistake. I built against the updated patch first, before finding that `sip-py` itself cannot be updated with Qt 5.7. It only changed by a 4-line offset; first and last characters of the md5sum being identical did not help much either...

Trying to figure out now if I can rearrange the commits more sensibly with `reset` or `revert`...

nieder:
For the twisted/constantly circular TestDepends, can they be 'fixed' so that tests for the one earlier in the stack (constantly-py?) are only run if twisted-py is already installed? Something similar to what sphinxcontrib-websupport-py is doing w/ sphinx-py and xapian-py?

nieder:
And twisted-py.info never got py34 variant.

I've built all the py packages above sip-py (so no sip-py or pyqt) and with the py34 addition, and working around the circular TestDepends, that chunk seems good.

nieder:
Going to start applying the lower level patches to simplify some of this. The order will be:
* 8a7ffd08 (tornado-py Depends; minus zopeinterface which is updated to 5.2.0 in master and some conflict resolutions)
* 513a0d0c (py38 variants to deps; without sip-py and pyqt changes)
* 2315ef6 and d7e8340 to get wxpython400 and wxpython410

That leaves:
pyqt4/5
sip-py
sip-pyqt4/5
matplotlib

nieder:
pypubsub4-py35 is failing tests with this error:
```
______________________________________________________________ test_export[0] ______________________________________________________________

repeat = 0

    @pytest.mark.parametrize('repeat', range(100))
    def test_export(repeat):
        # create a topic tree from a couple of the topic providers:
        clear_topic_tree()
        all_topics = create_all_defined_topics(topicDefns2, pub.TOPIC_TREE_FROM_STRING)
        all_topics.extend(create_all_defined_topics(topicDefns3, pub.TOPIC_TREE_FROM_CLASS))
    
        # export topic tree:
        try_call(100, pub.exportTopicTreeSpec, 'my_exported_topics')
    
        # import exported tree:
        clear_topic_tree()
>       all_topics2 = create_all_defined_topics('my_exported_topics')

suite/test4_provider.py:222: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
suite/test4_provider.py:82: in create_all_defined_topics
    prov = pub.addTopicDefnProvider(*args)
../build/lib/pubsub/core/topicmgr.py:128: in addDefnProvider
    provider = TopicDefnProvider(source, format or TOPIC_TREE_FROM_MODULE)
../build/lib/pubsub/core/topicdefnprovider.py:372: in __init__
    provider = providerClassObj(source, **providerKwargs)
../build/lib/pubsub/core/topicdefnprovider.py:283: in __init__
    module = import_module(moduleName)
/sw/lib/python3.5/importlib/__init__.py:126: in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
<frozen importlib._bootstrap>:985: in _gcd_import
    ???
<frozen importlib._bootstrap>:968: in _find_and_load
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

name = 'my_exported_topics', import_ = <function _gcd_import at 0x107a88048>

>   ???
E   ImportError: No module named 'my_exported_topics'

<frozen importlib._bootstrap>:955: ImportError
```
All other pypubsub4-pyXX packages are OK.

nieder:
OK, I pushed 1c97adb, 0f6318c, and 0e5a12f1 with your changes to everything except

pybpubsub
wxpython4{0,1}0
pyqt4/5
sip-py
sip-pyqt4/5
matplotlib

Once tests on pypubsub-py35 are fixed, then that and wxpython4X0 can be pushed. PyQt will be trickier esp Qt5 if it still needs Qt5 > 5.7 since that won't be available on many dists.

10.9-libcxx/stable/main/finkinfo/languages/sip-py.info
```diff
@@ -4,7 +4,7 @@ Info2: <<
 Package: sip-py%type_pkg[python]
 Version: 4.19.13
 Revision: 2
-Type: python (2.7 3.4 3.5 3.6 3.7)
+Type: python (2.7 3.4 3.5 3.6 3.7 3.8)
```
dmacks, line 7: For sip-pyXX-bin variants are mutually exclusive via Conflicts/Replaces. Adding py38 means that also needs to be added to the C/R entries of the SplitOff (and therefore bump Revision).

dmacks (changes requested):
By eye, these look pretty good. Definitely important to get us up to py38 available across-the-board. I'll build-test on my 10.13 tonight. One detail that needs fixing...

10.9-libcxx/stable/main/finkinfo/languages/sip-py.info
dhomeier: Right, in #675 I had changed this to handle the binaries with `update-alternatives` for the update in 
738fdd9; forgot I had not pulled that over.

dhomeier:
@nieder I could not reproduce the import error in the `pypubsub4-py35` tests on 10.14.6 even after updating the test setup to make sure it runs all 100 `test_export` incantation. Even have the `py34` version built and tested successfully now.
Which distribution/OS version have you been running?

10.9-libcxx/stable/main/finkinfo/languages/sip-py.info
dmacks: Making any change to a Conflicts or Replaces field means the revision must be bumped.

dmacks (changes requested):
```
Validating package file 10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt4-py.info...
Error: PatchFile-Checksum does not match PatchFile checksum. (sip-pyqt4-py.info)
Expected: d08dcb0733f7f0bc886672b22777dbf9
Actual: MD5(d53f5e4456cca0104bdd5a5c7d0dfa99)
        SHA1(619fca0e641bb0fa393e66cee69d2886160362d0)
        SHA256(36a13109adfcddb80fff338d813703056bece50428e8947ec6e1019b111ac372)
Validating package file 10.9-libcxx/stable/main/finkinfo/languages/sip-pyqt5-py.info...
Error: PatchFile-Checksum does not match PatchFile checksum. (sip-pyqt5-py.info)
Expected: d08dcb0733f7f0bc886672b22777dbf9
Actual: MD5(d53f5e4456cca0104bdd5a5c7d0dfa99)
        SHA1(619fca0e641bb0fa393e66cee69d2886160362d0)
        SHA256(36a13109adfcddb80fff338d813703056bece50428e8947ec6e1019b111ac372)
```

nieder:
For the pypubsub4-py, this is on 10.13/Xcode10.1. -py35 now sometimes passes tests for me, and now py36 sometimes doesn't (again on missing module 'my_exported_topics'). Maybe some sort of race condition, but not very consistent. Will push pypubsub4 and wxpython

nieder:
wxpython4x0-py38 keeps launching a gui app during tests :/

dhomeier:
> wxpython4x0-py38 keeps launching a gui app during tests :/

I should read the comments properly - I thought it was the matplotlib tests. :D
Changed the test setup a bit again with the upstream update, but I hope disabling `test__interactive_backend` will keep it from popping any windows up.
Will check if wxpython has some similar switches.

10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pygments-py.info
```diff
@@ -28,12 +35,12 @@ PostInstScript: <<
 <<
 PreRmScript: <<
  if [ $1 != "upgrade" ]; then
-  update-alternatives --remove pygmentize %p/bin/pygmentize%type_raw[python]
+  update-alternatives --remove pygmentize %p/bin/pygmentize-py%type_pkg[python]
```
nieder, line 38: Don't know if this is an issue or not, but does this leave any residues behind if you have the old version that installs as `%p/bin/pygmentize3.5` and now you're installing the new version with the file `%p/bin/pygmentize-py35` ? The `PreRmScript` from the previous version won't be run (only happens when not upgrading). I don't know how u-a deals with new sets of files taking over an existing namespace.

dhomeier: You're right; had been wondering about this briefly, but I was worried that a manpage like `pygmentize3.8.1` might not be correctly recognised.
I am trying to check for old installed versions of the binaries now, but apparently a test like
```bash
 if [ -f %p/bin/pygmentize%type_raw[python] ]; then
  update-alternatives --remove-all pygmentize
 fi
```
does not respond – is this a limitation of the restricted interpreter for `PostInstScript`, or do you know if it has a different syntax for tests?

nieder: Don't think there are restrictions on P{re,ost}InstScript. The contents of the .info field are just added to a shell script header (see %p/var/lib/dpkg/info for many examples). So that's just normal `#!/bin/sh` call. What happens if you just `ls -l %p/bin/pygmentize*` to debug?
Also, fixing the symlinks probably belongs in PreInstScript for the updated `pygmentize-py.info`

dhomeier: > Also, fixing the symlinks probably belongs in PreInstScript for the updated `pygmentize-py.info`

Uh, of course - it was looking for the old file after it had already been replaced from the new package!
In PreInstScript it works as it should (I hope at least, as `update-alternatives --remove` seems to work somewhat unpredictably regarding the changes to `%p/etc/alternatives/pygmentize`...).

dhomeier:
I could not really identify the gui tests in wxpython410, but disabled a couple more that kept crashing the python interpreter before the test suite could complete, at least on my machine (went to 4.1.1 for this). Of course there are plenty of windows popping up - if those are breaking tests when run remotely, one will probably have to build without tests.

I have also checked that matplotlib 2.2.5 also worked with wxpython410; since that mpl branch has not changed in a while, I have set it as the LTS version for all Pythons < 3.6. With 2.7 and 3.5 having reached end-of-life last April and September, respectively, I'd rather not add any more maintenance burden for those...

nieder:
> Yes, the -dev and -shlibs packages install all version-numbered files. I've added Conflicts: wxpython410-py%type_pkg[python] to wxpython400 as well, but Replaces only to wxpython410 – does that make sense?

The two end-user executables should either C/R each other to allow for clean swapping or share a common name. I think given the current SplitOff layout and naming scheme, having C/R is the easiest.

nieder:
I had to locally disable these tests to pass py36/37 (and py38 just hangs at the start of the tests regardless):
```
tipwin_Tests
(treectrl_Tests and treectrlTreeItemData)
(grid_Tests and grid46))
```
Also, using `xvfb-run -a` when loading the tests blocked a huge percentage of the popup windows (I think the tests still ran, they were just not on the main display).

Anyway, python tests are notoriously buggy, so can deal with those later since the packages themselves look fine.

Note: the perl-pie for `/usr/bin/env python` can probably be improved to provide just the full path to `%p/bin/python%type_raw[python]` rather than depending on ENV sanity. Can deal with that after everything else is done since it's just cleaning up and current is good enough.

dhomeier:
> > Yes, the -dev and -shlibs packages install all version-numbered files. I've added Conflicts: as well, but Replaces only to wxpython410 – does that make sense?
> 
> The two end-user executables should either C/R each other to allow for clean swapping or share a common name. I think given the current SplitOff layout and naming scheme, having C/R is the easiest.

By "share a common name" do you mean both packages being called `wxpython-pyNN` (but not the -shlibs)? I thought I recalled from earlier discussions on the naming scheme, what different names `wxpython4NN` were required by the library version scheme. A single package name still seems easier for the dependent packages to keep up with, but since we now have 400 and 410, it is perhaps best to keep it that way.

I had not added the `Replaces: wxpython410-py%type_pkg[python]` in wxpython400 as I expect the former may have functionality not provided by wxpython400 – or does the C/R always have to apply mutually. Would that mean that packages really depending on wxpython410 will then actually need a `Depends: wxpython410-py%type_pkg[python] (>= 4.1.0)`?

dhomeier:
> Note: the perl-pie for `/usr/bin/env python` can probably be improved to provide just the full path to `%p/bin/python%type_raw[python]` rather than depending on ENV sanity. Can deal with that after everything else is done since it's just cleaning up and current is good enough.

Well, hopefully in a Fink environment Fink's `python%type_raw[python]` will be first in $PATH, so I tried to keep the changes to the scripts minimal. But since this is packaged for that particular executable, I guess we might call it directly as well.

nieder:
pygments-py38 massively fails starting tests (all others OK)
```
plugins: benchmark-3.2.3, relaxed-1.1.4
collected 3399 items / 4 errors / 8 deselected / 3387 selected                                                                           

================================================================= ERRORS =================================================================
_______________________________________________________ ERROR collecting setup.py ________________________________________________________
/sw/lib/python3.8/distutils/fancy_getopt.py:233: in getopt
    opts, args = getopt.getopt(args, short_opts, self.long_opts)
/sw/lib/python3.8/getopt.py:95: in getopt
    opts, args = do_shorts(opts, args[0][1:], shortopts, args[1:])
/sw/lib/python3.8/getopt.py:195: in do_shorts
    if short_has_arg(opt, shortopts):
/sw/lib/python3.8/getopt.py:211: in short_has_arg
    raise GetoptError(_('option -%s not recognized') % opt, opt)
E   getopt.GetoptError: option -k not recognized

During handling of the above exception, another exception occurred:
/sw/lib/python3.8/distutils/core.py:134: in setup
    ok = dist.parse_command_line()
/sw/lib/python3.8/site-packages/setuptools/dist.py:500: in parse_command_line
    result = _Distribution.parse_command_line(self)
/sw/lib/python3.8/distutils/dist.py:475: in parse_command_line
    args = parser.getopt(args=self.script_args, object=self)
/sw/lib/python3.8/distutils/fancy_getopt.py:235: in getopt
    raise DistutilsArgError(msg)
E   distutils.errors.DistutilsArgError: option -k not recognized

During handling of the above exception, another exception occurred:
setup.py:28: in <module>
    setup(
/sw/lib/python3.8/site-packages/setuptools/__init__.py:140: in setup
    return distutils.core.setup(**attrs)
/sw/lib/python3.8/distutils/core.py:136: in setup
    raise SystemExit(gen_usage(dist.script_name) + "\nerror: %s" % msg)
E   SystemExit: usage: pytest.py [global_opts] cmd1 [cmd1_opts] [cmd2 [cmd2_opts] ...]
E      or: pytest.py --help [cmd1 cmd2 ...]
E      or: pytest.py --help-commands
E      or: pytest.py cmd --help
E   
E   error: option -k not recognized
____________________________________________ ERROR collecting external/markdown-processor.py _____________________________________________
ImportError while importing test module '/sw/build.build/pygments-py38-2.7.4-1/Pygments-2.7.4/external/markdown-processor.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
external/markdown-processor.py:38: in <module>
    from markdown.preprocessors import Preprocessor
E   ModuleNotFoundError: No module named 'markdown'
________________________________________ ERROR collecting tests/examplefiles/linecontinuation.py _________________________________________
tests/examplefiles/linecontinuation.py:1: in <module>
    apple.filter(x, y)
E   NameError: name 'apple' is not defined
___________________________________________ ERROR collecting tests/examplefiles/unicodedoc.py ____________________________________________
/sw/lib/python3.8/site-packages/_pytest/python.py:501: in _importtestmodule
    mod = self.fspath.pyimport(ensuresyspath=importmode)
/sw/lib/python3.8/site-packages/py/_path/local.py:701: in pyimport
    __import__(modname)
E     File "/sw/build.build/pygments-py38-2.7.4-1/Pygments-2.7.4/tests/examplefiles/unicodedoc.py", line 2
E       ur"""unicode-raw"""
E         ^
E   SyntaxError: invalid syntax
```

dhomeier:
> pygments-py38 massively fails starting tests (all others OK)
> 
> ```
> plugins: benchmark-3.2.3, relaxed-1.1.4

Urgh, I had forgotten that package is still around! IMO pytest-relaxed should better be purged completely, since it breaks far more pytest runs than it enables; forcing either to put `TestConflicts` or hack their `setup.cfg` or `pytest.ini` in every single of those packages.
AFAICS the only packages really depending it are paramiko-py and invoke-py; one could fix the Tests of the former to use a local build of pytest-relaxed. I haven't been able to get any tests for invoke to run at all on 10.14, neither with or without pytest-relaxed...

nieder:
Confirming that pytest-relaxed (even bumping to latest 1.1.5 which actually passes all its own tests) was the root cause my py38 test failures. Pushed pygments to master.

nieder:
pytest-relaxed is yours, so it can be modified as needed.

nieder:
All sip-pyXX and sip-pyqt4-pyXX build and pass.
All sip-pyqt5-pyXX pass except py38 that has this build failure:
```
/Applications/Xcode.app/Contents/Developer/usr/bin/g++ -headerpad_max_install_names -Wl,-syslibroot,/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.14.sdk -mmacosx-version-min=10.13   -bundle -o QtCore qpycore_post_init.o sipQtCoreQAbstractAnimation.o sipQtCoreQAbstractEventDispatcher.o sipQtCoreQAbstractEventDispatcherTimerInfo.o sipQtCoreQAbstractItemModel.o sipQtCoreQAbstractListModel.o sipQtCoreQAbstractNativeEventFilter.o sipQtCoreQAbstractProxyModel.o sipQtCoreQAbstractState.o sipQtCoreQAbstractTableModel.o sipQtCoreQAbstractTransition.o sipQtCoreQAnimationGroup.o sipQtCoreQBasicTimer.o sipQtCoreQBitArray.o sipQtCoreQBuffer.o sipQtCoreQByteArray.o sipQtCoreQByteArrayBase64Options.o sipQtCoreQByteArrayMatcher.o sipQtCoreQChar.o sipQtCoreQChildEvent.o sipQtCoreQCollator.o sipQtCoreQCollatorSortKey.o sipQtCoreQCommandLineOption.o sipQtCoreQCommandLineParser.o sipQtCoreQCoreApplication.o sipQtCoreQCryptographicHash.o sipQtCoreQDataStream.o sipQtCoreQDate.o sipQtCoreQDateTime.o sipQtCoreQDir.o sipQtCoreQDirFilters.o sipQtCoreQDirIterator.o sipQtCoreQDirIteratorIteratorFlags.o sipQtCoreQDirSortFlags.o sipQtCoreQDynamicPropertyChangeEvent.o sipQtCoreQEasingCurve.o sipQtCoreQElapsedTimer.o sipQtCoreQEvent.o sipQtCoreQEventLoop.o sipQtCoreQEventLoopLocker.o sipQtCoreQEventLoopProcessEventsFlags.o sipQtCoreQEventTransition.o sipQtCoreQFile.o sipQtCoreQFileDevice.o sipQtCoreQFileDeviceFileHandleFlags.o sipQtCoreQFileDevicePermissions.o sipQtCoreQFileInfo.o sipQtCoreQFileSelector.o sipQtCoreQFileSystemWatcher.o sipQtCoreQFinalState.o sipQtCoreQGenericArgument.o sipQtCoreQGenericReturnArgument.o sipQtCoreQHash0100QString0100QVariant.o sipQtCoreQHash18000100QByteArray.o sipQtCoreQHistoryState.o sipQtCoreQIODevice.o sipQtCoreQIODeviceOpenMode.o sipQtCoreQIdentityProxyModel.o sipQtCoreQItemSelection.o sipQtCoreQItemSelectionModel.o sipQtCoreQItemSelectionModelSelectionFlags.o sipQtCoreQItemSelectionRange.o sipQtCoreQJsonArray.o sipQtCoreQJsonDocument.o sipQtCoreQJsonObject.o sipQtCoreQJsonParseError.o sipQtCoreQJsonValue.o sipQtCoreQLibrary.o sipQtCoreQLibraryInfo.o sipQtCoreQLibraryLoadHints.o sipQtCoreQLine.o sipQtCoreQLineF.o sipQtCoreQList0100QAbstractEventDispatcherTimerInfo.o sipQtCoreQList0100QByteArray.o sipQtCoreQList0100QCommandLineOption.o sipQtCoreQList0100QFileInfo.o sipQtCoreQList0100QLocale.o sipQtCoreQList0100QMimeType.o sipQtCoreQList0100QModelIndex.o sipQtCoreQList0100QPersistentModelIndex.o sipQtCoreQList0100QStorageInfo.o sipQtCoreQList0100QUrl.o sipQtCoreQList0100QVariant.o sipQtCoreQList0100QtDayOfWeek.o sipQtCoreQList0101QAbstractAnimation.o sipQtCoreQList0101QAbstractState.o sipQtCoreQList0101QAbstractTransition.o sipQtCoreQList0101QObject.o sipQtCoreQList0600QPair0100QString0100QString.o sipQtCoreQList0600QPair18001800.o sipQtCoreQList1800.o sipQtCoreQList2400.o sipQtCoreQLocale.o sipQtCoreQLocaleNumberOptions.o sipQtCoreQLockFile.o sipQtCoreQMap18000100QVariant.o sipQtCoreQMargins.o sipQtCoreQMarginsF.o sipQtCoreQMessageAuthenticationCode.o sipQtCoreQMessageLogContext.o sipQtCoreQMessageLogger.o sipQtCoreQMetaClassInfo.o sipQtCoreQMetaEnum.o sipQtCoreQMetaMethod.o sipQtCoreQMetaObject.o sipQtCoreQMetaObjectConnection.o sipQtCoreQMetaProperty.o sipQtCoreQMetaType.o sipQtCoreQMetaTypeTypeFlags.o sipQtCoreQMimeData.o sipQtCoreQMimeDatabase.o sipQtCoreQMimeType.o sipQtCoreQModelIndex.o sipQtCoreQMutex.o sipQtCoreQMutexLocker.o sipQtCoreQObject.o sipQtCoreQObjectCleanupHandler.o sipQtCoreQPair18001800.o sipQtCoreQParallelAnimationGroup.o sipQtCoreQPauseAnimation.o sipQtCoreQPersistentModelIndex.o sipQtCoreQPluginLoader.o sipQtCoreQPoint.o sipQtCoreQPointF.o sipQtCoreQProcess.o sipQtCoreQProcessEnvironment.o sipQtCoreQPropertyAnimation.o sipQtCoreQReadLocker.o sipQtCoreQReadWriteLock.o sipQtCoreQRect.o sipQtCoreQRectF.o sipQtCoreQRegExp.o sipQtCoreQRegularExpression.o sipQtCoreQRegularExpressionMatch.o sipQtCoreQRegularExpressionMatchIterator.o sipQtCoreQRegularExpressionMatchOptions.o sipQtCoreQRegularExpressionPatternOptions.o sipQtCoreQResource.o sipQtCoreQRunnable.o sipQtCoreQSaveFile.o sipQtCoreQSemaphore.o sipQtCoreQSequentialAnimationGroup.o sipQtCoreQSet0101QAbstractState.o sipQtCoreQSettings.o sipQtCoreQSharedMemory.o sipQtCoreQSignalBlocker.o sipQtCoreQSignalMapper.o sipQtCoreQSignalTransition.o sipQtCoreQSize.o sipQtCoreQSizeF.o sipQtCoreQSocketNotifier.o sipQtCoreQSortFilterProxyModel.o sipQtCoreQStandardPaths.o sipQtCoreQStandardPathsLocateOptions.o sipQtCoreQState.o sipQtCoreQStateMachine.o sipQtCoreQStateMachineSignalEvent.o sipQtCoreQStateMachineWrappedEvent.o sipQtCoreQStorageInfo.o sipQtCoreQString.o sipQtCoreQStringList.o sipQtCoreQStringListModel.o sipQtCoreQStringRef.o sipQtCoreQSysInfo.o sipQtCoreQSystemSemaphore.o sipQtCoreQTemporaryDir.o sipQtCoreQTemporaryFile.o sipQtCoreQTextBoundaryFinder.o sipQtCoreQTextBoundaryFinderBoundaryReasons.o sipQtCoreQTextCodec.o sipQtCoreQTextCodecConversionFlags.o sipQtCoreQTextCodecConverterState.o sipQtCoreQTextDecoder.o sipQtCoreQTextEncoder.o sipQtCoreQTextStream.o sipQtCoreQTextStreamManipulator.o sipQtCoreQTextStreamNumberFlags.o sipQtCoreQThread.o sipQtCoreQThreadPool.o sipQtCoreQTime.o sipQtCoreQTimeLine.o sipQtCoreQTimeZone.o sipQtCoreQTimeZoneOffsetData.o sipQtCoreQTimer.o sipQtCoreQTimerEvent.o sipQtCoreQTranslator.o sipQtCoreQUrl.o sipQtCoreQUrlComponentFormattingOptions.o sipQtCoreQUrlFormattingOptions.o sipQtCoreQUrlQuery.o sipQtCoreQUrlUserInputResolutionOptions.o sipQtCoreQUuid.o sipQtCoreQVariant.o sipQtCoreQVariantAnimation.o sipQtCoreQVariantMap.o sipQtCoreQVector0100QPointF.o sipQtCoreQVector0100QTimeZoneOffsetData.o sipQtCoreQVector0100QXmlStreamEntityDeclaration.o sipQtCoreQVector0100QXmlStreamNamespaceDeclaration.o sipQtCoreQVector0100QXmlStreamNotationDeclaration.o sipQtCoreQVector0600QPair24000100QVariant.o sipQtCoreQVector1800.o sipQtCoreQVersionNumber.o sipQtCoreQWaitCondition.o sipQtCoreQWriteLocker.o sipQtCoreQXmlStreamAttribute.o sipQtCoreQXmlStreamAttributes.o sipQtCoreQXmlStreamEntityDeclaration.o sipQtCoreQXmlStreamEntityResolver.o sipQtCoreQXmlStreamNamespaceDeclaration.o sipQtCoreQXmlStreamNotationDeclaration.o sipQtCoreQXmlStreamReader.o sipQtCoreQXmlStreamWriter.o sipQtCoreQt.o sipQtCoreQtAlignment.o sipQtCoreQtApplicationStates.o sipQtCoreQtDockWidgetAreas.o sipQtCoreQtDropActions.o sipQtCoreQtEdges.o sipQtCoreQtFindChildOptions.o sipQtCoreQtGestureFlags.o sipQtCoreQtImageConversionFlags.o sipQtCoreQtInputMethodHints.o sipQtCoreQtInputMethodQueries.o sipQtCoreQtItemFlags.o sipQtCoreQtKeyboardModifiers.o sipQtCoreQtMatchFlags.o sipQtCoreQtMouseButtons.o sipQtCoreQtMouseEventFlags.o sipQtCoreQtOrientations.o sipQtCoreQtScreenOrientations.o sipQtCoreQtTextInteractionFlags.o sipQtCoreQtToolBarAreas.o sipQtCoreQtTouchPointStates.o sipQtCoreQtWindowFlags.o sipQtCoreQtWindowStates.o sipQtCorecmodule.o sipQtCoreqintptr.o sipQtCorequintptr.o qpycore_chimera.o qpycore_chimera_signature.o qpycore_chimera_storage.o qpycore_classinfo.o qpycore_decorators.o qpycore_enums_flags.o qpycore_event_handlers.o qpycore_init.o qpycore_misc.o qpycore_public_api.o qpycore_pyqtboundsignal.o qpycore_pyqtconfigure.o qpycore_pyqtmethodproxy.o qpycore_pyqtproperty.o qpycore_pyqtpyobject.o qpycore_pyqtsignal.o qpycore_pyqtslot.o qpycore_pyqtslotproxy.o qpycore_qjsonvalue.o qpycore_qmessagelogger.o qpycore_qmetaobject.o qpycore_qmetaobject_helpers.o qpycore_qobject_getattr.o qpycore_qobject_helpers.o qpycore_qstring.o qpycore_qt_conf.o qpycore_qvariant.o qpycore_qvariant_value.o qpycore_qvariantmap.o qpycore_types.o moc_qpycore_event_handlers.o  -F/sw/lib/qt5-mac/lib -ldl -Wl,-framework,CoreFoundation -framework QtCore -Wl,-framework,DiskArbitration -Wl,-framework,IOKit  
  "_PyArg_ParseTuple", referenced from:
      pyqtBoundSignal_disconnect(_object*, _object*) in qpycore_pyqtboundsignal.o
  "_PyArg_ParseTupleAndKeywords", referenced from:
      qpycore_pyqtslot(_object*, _object*) in qpycore_decorators.o
      pyqtBoundSignal_connect(_object*, _object*, _object*) in qpycore_pyqtboundsignal.o
      pyqtProperty_init(_object*, _object*, _object*) in qpycore_pyqtproperty.o
      pyqtProperty_call(_object*, _object*, _object*) in qpycore_pyqtproperty.o
  "_PyBool_FromLong", referenced from:
      meth_QAbstractAnimation_event(_object*, _object*) in sipQtCoreQAbstractAnimation.o
      meth_QAbstractEventDispatcher_hasPendingEvents(_object*, _object*) in sipQtCoreQAbstractEventDispatcher.o
      meth_QAbstractEventDispatcher_processEvents(_object*, _object*) in sipQtCoreQAbstractEventDispatcher.o
      meth_QAbstractEventDispatcher_unregisterTimer(_object*, _object*) in sipQtCoreQAbstractEventDispatcher.o
      meth_QAbstractEventDispatcher_unregisterTimers(_object*, _object*) in sipQtCoreQAbstractEventDispatcher.o
      meth_QAbstractItemModel_beginMoveColumns(_object*, _object*) in sipQtCoreQAbstractItemModel.o
      meth_QAbstractItemModel_beginMoveRows(_object*, _object*) in sipQtCoreQAbstractItemModel.o
...
<bunch more missing symbols>
  "__Py_TrueStruct", referenced from:
      qpycore_convertTo_QJsonValue(_object*, _object*, QJsonValue**, int*) in qpycore_qjsonvalue.o
ld: symbol(s) not found for architecture x86_64
```

dmacks:
@nieder that's #724

nieder:
Nice.  Changing LIBS to `"-L%p/lib `%p/bin/python%type_raw[python]-config --embed --ldflags`"` makes `pyqt5-mac-py38` build successfully. I've got all the sip and pyqt4/5 packages built now from commits here. The `--embed` addition just has to be conditionalized on py38.

10.9-libcxx/stable/main/finkinfo/sci/matplotlib-py.info
```diff
@@ -1,163 +1,184 @@
+# -*- coding: ascii; tab-width: 4 -*-; tab-width: 4; x-counterpart: matplotlib-py.patch -*-
```
cooljeanius, line 1: duplicated `tab-width: 4;` here

cooljeanius:
```suggestion
# -*- coding: ascii; tab-width: 4 -*-; x-counterpart: matplotlib-py.patch -*-
```

PR #679, opened 2020-11-01
cpan-pmXXX new version

dmacks:
Trivial version update. But could someone check whether the self-tests perform network access, and disable as appropriate?


dmacks:
If someone whines about net-access, we can address it at that time.

PR #684, opened 2020-11-13
rsync autoconf: work-around for new xcode

dmacks:
See WayneD/rsync#116


nieder (approved):
WFM on 11.0/Xcode12.2

PR #685, opened 2020-11-18
Fix slang2 testing on 11.0

nieder:
Using i_n_t on the SONAME symlink causes the actual library to be rebuilt with the original install_name during `make check`, which then fails.
Use i_n_t on the physical file to avoid the rebuild.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/slang2.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/slang2.info
@@ -30,7 +30,7 @@ UseMaxBuildJobs: False
 SetCFLAGS: -isystem %p/include
 NoSetCFLAGS: true
 NoSetLDFLAGS: true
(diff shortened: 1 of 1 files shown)


fingolfin:
I have no idea what this does, so i can't really review it. However, I have no objections to it, either.

nieder:
It seems to be a timing issue (don't know if it's APFS or 11.0 specifically) that if you apply the install_name_tool on the libslang.2.dylib symlink, it causes make to rebuild the library when running make check, therefore undoing the change right before the tests are run. But if install_name_tool is run on the actual file, then it's not rebuilt and `make check` works.

nieder:
Pushed since it doesn't change the actual output, just the path, so that it's consistent on newer systems.

dhomeier:
Sounds quite like my libgsm1 problem in #690, which did occur on 10.14 APFS.

PR #694, opened 2020-11-30
udunits2 2.2.28 - new upstream

PovlAbrahamsen:
Updated the UDUNITS2 package to the latest upstream version. Only major change is to use GCC 9 as the compiler, so it plays nice with Fortran (I could not get this to compile with Xcode 12 otherwise).

Tested on 10.15.7 with xcode command line tools 12.2. Compiles and passes tests with "-m".

Maintainer is @akhansen.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/udunits2.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/udunits2.info
@@ -1,21 +1,20 @@
-Info2: <<
 Package: udunits2
-Version: 2.2.26
+Version: 2.2.28
 Revision: 1
 Maintainer: Alexander Hansen <alexkhansen@users.sourceforge.net>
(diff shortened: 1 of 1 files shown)


nieder:
Upstream bug for the Xcode 12 build failures
https://github.com/Unidata/UDUNITS-2/issues/97

PovlAbrahamsen:
Considering that this is being worked on upstream, perhaps it would be better to wait until this is fixed, so we can continue to use this with Xcode rather than GCC? I'm happy to close this, and keep an eye on the upstream bug.

PovlAbrahamsen:
The latest release (on the Unidata website, not Github) does build with Xcode 12.3! Should be ready to review...

akhansen (approved):
Let's update it.

PR #693, opened 2020-11-30
Netcdf updates: netcdf-c18 4.7.4, netcdf-cxx4 4.3.1, netcdf-fortran8 4.5.3, nco 4.9.5, ncview 2.1.7

PovlAbrahamsen:
Here are a bunch of NetCDF updates:

1. netcdf-c18 version 4.7.4 - latest upstream version.
   - updated other netcdf packages to conflict/replace netcdf-c18. Revisions not bumped.
   - only packaging change from before is to remove private library libh5bzip2.so. This had previously been called libbzip2.so, and there was code in the previous infofile to delete it before packaging. In netcdf-c15 this was installed, without listing it as a private library in the shlibs section. If anyone feels strongly that this will be used by anyone in the future, it can be included and declared. This has previously been discussed at Unidata/netcdf-c#1348 
2. netcdf-cxx4 4.3.1 - latest upstream version.
   - update source path to download from Github
   - remove installation of libh5bzip2.so from this one, too! See Unidata/netcdf-cxx4#75
   - generally, the build chain for this library is a bit of a mess (not our fault, upstream; see, e.g. Unidata/netcdf-cxx4#73, Unidata/netcdf-cxx4#87). Note that HDF5 is required during compile/test time, but not linked directly to this library. It is, however, a dependency of netcdf-c18-shlibs, which in turn is a dependency of this library.
3. netcdf-fortran8 4.5.3 - latest upstream version
4. nco 4.9.5 - latest upstream version.
   - changed to build with netcdf-c18
   - remove patchfile, since this was fixed upstream
   - removed redundant line in patchscript
5. ncview 2.1.7
   - changed to build with netcdf-c18
   - bug fix for xcode 12: there were two implicitly declared functions

PR #695, opened 2020-11-30
gmt6 6.1.1 - new upstream

PovlAbrahamsen:
New upstream version of GMT 6.

Depends on #693 for netcdf-c18.

Changes:
- add dependency to libgeos3.6.1
- update NetCDF to netcdf-c18
- add COPY_DCW=FALSE and COPY_GSHHG=FALSE to cmake parameters to avoid clash with dcw-gmt and gshhg packages!
- updated homepage URL

Tested on 10.15.7 with xcode command line tools 12.2. Compiles with "-m".

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/gmt6.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/gmt6.info
@@ -1,10 +1,9 @@
 Package: gmt6
-# 6.1.0 has file collision with dcw-gmt
-Version: 6.0.0
+Version: 6.1.1
 Revision: 1
 Source: https://github.com/GenericMappingTools/gmt/releases/download/%v/gmt-%v-src.tar.xz
-Source-MD5: 2721f3bd7f39eb7c8ea261f644c5c36e
-Source-Checksum: SHA1(3f11da2740402d4d1d98c55b46f4f18e050f76c0)
+Source-MD5: 3024b7f789d5ab38ae0fa92bdab43b5f
+Source-Checksum: SHA1(591c318cfa1a96fb1f39a1b3df4b1ac4da9f1530)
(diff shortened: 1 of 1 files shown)


nieder:
Are the conflicts/replaces in the -shlibs SplitOff due to `plugins/supplements.so`? If so, that file should probably be moved out of the -shlibs SplitOff. Debian has it in the main gmt package along side the executables. Looking at the gmt* packages, they're a bit misorganized. I can deal with this once netcdf-c18 goes in.

PovlAbrahamsen:
I think the only file conflict would be `plugins/supplements.so`. Perhaps this is something to change in gmt5, too. Will you do this, or shall I?

PovlAbrahamsen:
Also, should a `.so` file be declared as a private library in the Shlibs field?

nieder:
`supplements.so` is not a dylib, it's just a plugin/bundle so it doesn't need to be declared:
```
[Bubble:/sw/fink/dists/stable/main/finkinfo] nieder $ otool -hv /sw/lib/gmt/plugins/supplements.so /sw/lib/libgmt.6.0.0.dylib
Mach header
      magic cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64  X86_64        ALL  0x00      BUNDLE    22       2152   NOUNDEFS DYLDLINK TWOLEVEL
Mach header
      magic cputype cpusubtype  caps    filetype ncmds sizeofcmds      flags
MH_MAGIC_64  X86_64        ALL  0x00       DYLIB    22       2152   NOUNDEFS DYLDLINK TWOLEVEL NO_REEXPORTED_DYLIBS
```
I'll update and refactor the gmt set of libraries.

nieder:
That was fun :/

PR #696, opened 2020-11-30
mbsystem 5.7.6

PovlAbrahamsen:
Latest release of mbsystem.
This PR depends on #693 and #695. Maintainer is @schwehr. 
Changes:
- use latest netcdf, proj, and GMT libraries.
- update source location to Github, remove second source, since Levitus82 database is provided within main distribution.
- custom install script no longer required.
- enable tests (require python 3.6 or above).
- add a note about flat namespaces for libxt...

PR #697, opened 2020-12-02
gdal2 2.4.4 - new upstream

PovlAbrahamsen:
This updates gdal2 to the latest upstream version. At some point, someone should perhaps write a gdal3 package...

List of changes:
- changed to use netcdf-c18 instead of netcdf-c15
- removed two lines from patchscript, which disabled compiler warnings. compiles just fine without this tweak using xcode 12.2.
- updated list of DocFiles
- removed DescPackaging, which was pointing to a non-existent page on the Fink wiki
- removed @schwehr as maintainer, based on his comments in #696, indicating that he no longer uses Fink

In gdal-py:
- removed support for python 3.4, added 3.8
- updated DocFiles and maintainer as above
- updated patchfile (just a few line numbers)

Compiles on 10.15.7 with "-m", with python 2.7 and 3.8.


schwehr:
Thanks for removing me from the maintainer field

nieder:
Can the postrgesql dep be bumped to postgresql12 ? That's the latest major version pgsql. Though it might not be worth it if it pulls different bits of the GIS, etc stack to use different pgsql libraries.

dmacks:
gdal2 (and gdal) seem to have no underlying other packages with pgsql deps. Please also check if they are using the latest libgeos.

PovlAbrahamsen:
Bumped to postgresql12, as well as cfitsio8 and pcre2.
Other observations: bash, json-c, libogdi3, and unixodbc/libodbc3 don't build with Xcode 12...

nieder (approved):

PR #700, opened 2020-12-06
unixodbc: version 2.3.9

PovlAbrahamsen:
New upstream version of unixODBC, with a patchfile to fix a whole bunch of implicit declarations.

Builds on 10.15.7 with xcode 12.2. Maintainer is @TheSin-


TheSin-:
Thanks I’ll try to test it this week

nieder:
Oops. I never received an email about this PR and I already pushed a similar set of changes so that I could build more on 10.15. But the changes from here are cleaner by adding the declarations to headers rather than the .c files. Sorry about that.

BTW: upstream bug report https://github.com/lurcher/unixODBC/issues/49

PR #702, opened 2020-12-08
CRAN 20201208

babayoshihiko:
New upstream versions for CRAN as of 20201208.


nieder (approved):
Look visually OK, but did not try building anything.

PR #708, opened 2020-12-13
Cppunit upgrade

nieder:
Bump packages using cppunit1.12.1 to cppunit1.15.1.  cppunit1.12.1 doesn't build on Big Sur.

After this, `libetonyek` is the only package left using the old cppunit1.12.1, and that's failing tests with either version of cppunit. There's newer `libetonyek`, but it requires an upgrade to a new libN of `mdds`.


dmacks:
Looks good by eye, and I also like the boost upgrade. Flagging danielj as some are his packages, but no need to wait more than a few days at most.

PR #709, opened 2020-12-16
cmake 3.19.1 - new upstream

PovlAbrahamsen:
New upstream version of cmake. Same patches as before. Updated source URL to Github, and website URL to https. Compiles with "-m" using xcode 12.3 on 10.15.7 and appears to work correctly. Maintainer is @costabel.


costabel (approved):
OK with me

PR #712, opened 2020-12-19
A few geospatial updates: libspatialite7, libgeotiff2, libogdi4, gdal2

PovlAbrahamsen:
A few geospatial updates:

- libspatialite7: (latest release in version 4 is 4.3.0; version 5 requires libgeos 3.7.0 or newer)
   - bump to use libproj19 instead of libproj9
   - updated URL
- libgeotiff2: bump to use libproj19 instead of libproj9
- libogdi4: new upstream version 4.1.0
   - patches based on libogdi3
   - libraries are now installed in %p/lib/ogdi/ (new upstream behavior), so can co-exist with libogdi3-shlibs even though version numbers aren't included in the dylib filenames
   - created a symlink from %p/lib/libogdi.dylib to ogdi/libogdi.dylib for consistency with previous versions
   - requires buildconflicts with libogdi3 and libogdi4; otherwise the build process links to the installed library rather than the one being built!
- libogdi3: add conflicts/replaces with libogdi4 (except shlibs)
- gdal2: bump to use libogdi4 instead of libogdi3

Maintainer of all except gdal2 is @babayoshihiko. gdal2 doesn't have a maintainer.


dmacks:
libspatialite7 looks good. libgeotiff2 looks good but might be good practice to remove "-I%p/include" from SetCPPFLAGS (fink automatically does that).

Having libogdi4 bury its libraries is weird, and having it not include the libversion anywhere in the install_name is going to create difficulties with filename collisions in future libversions.

dmacks (changes requested):
Having libogdi4 bury its libraries is weird, and having it not include the libversion anywhere in the install_name is going to create difficulties with filename collisions in future libversions. Should be %p/lib/lib*.4.dylib. Debian fought with this package also, and patched to accomplish it:

https://sources.debian.org/src/ogdi-dfsg/4.1.0+ds-3/debian/patches/shared/

noting that the linux convension is libFOO.(EXT).(X) instead of our libFOO.(X).(EXT)

dmacks:
Could you check if gdal2 can be bumped from postgresql96 to postgresql12 ?

PovlAbrahamsen:
> Could you check if gdal2 can be bumped from postgresql96 to postgresql12 ?

This has already happened in #697

dmacks:
> > Could you check if gdal2 can be bumped from postgresql96 to postgresql12 ?
> 
> This has already happened in #697

Whoops, thanks...didn't notice when updating my checklist.

PovlAbrahamsen:
I've updated libgeotiff2 as requested (trivial change).

Changes to libogdi4 are slightly more substantial:
- plugins are now compiled as bundles rather than dylibs, and still reside in %p/lib/ogdi/ with extension .so; these have been removed from the "shlibs" field. Nobody should be linking directly to them anyway.
- main library is compiled as a dylib, named libogdi.4.dylib, and resides in %p/lib
- link libogdi.dylib is still created in libogdi4
- included Debian patches "spelling-errors" and "hardening"
- in case of future updates, we might need to place plugins in a different directory, to avoid conflicts.

PovlAbrahamsen:
Have also added a new upstream version of libgeotiff: libgeotiff5. Includes appropriate conflicts/replaces with libgeotiff2.

dmacks:
Looking at libgeotiff5...

The Puma patch is no longer necessary (the ./configure script was generated with a newer autotools that has this already included).

```configure: WARNING: unrecognized options: --enable-incode-epsg```

It appears to use pkg-config for some detection, so it needs a BuildDepends for that.

The Conflicts/Replaces between libgeotiff*-bin do not seem symmetric. If libgeotiff2-bin:C/R:libgeotiff5-bin (which makes sense), why doesn't libgeotiff5-bin:C/R:libgeotiff2-bin?

According to NEWS, it now uses the more modern libproj interface, so it might not need the -D for the old API.

PovlAbrahamsen:
I'm not sure if pkg-config is actually being used, since we specify the library paths directly. But I have added a BuildDepends for safety's sake. I've also fixed the other issues (oops!). Regarding the last point: do you think this test should be removed from the configure script for safety's sake? Or shall we leave it as it is for now?

I have also fixed a few other issues in gdal2:
- bumped libgeotiff2 to libgeotiff5
- I had previously incremented libpcre1 to libprcre2 - but this was a mistake. GDAL doesn't support libpcre2, so it was being disabled
- removed dependencies to libdab11 and szip, neither of which are (currently?) being used

PovlAbrahamsen:
DODS/libdap11 support has now been re-enabled in gdal2.

dmacks:
I found an upstream bug that affects the pkg-config detection of lots of things:  https://github.com/OSGeo/libgeotiff/issues/51 I suspect you can allow autodetection (rather than needing any --with-* for tiff or proj), but once you pass --with-libtiff explicitly, detection of proj breaks.

dmacks:
Please add a link to https://github.com/OSGeo/libgeotiff/issues/52 as the upstream bug that we are solving by using flag-sort

PovlAbrahamsen:
Hopefully this works.

In gdal2, I've also added some conflicts/replaces to gdal3 (work in progress), and fixed building against Fink's qhull, rather than the internal one.

dmacks:
I think this looks fine. Ping @babayoshihiko who is maintainer of a few of them.

nieder:
I just pushed a fix for Xcode12 for libogdi3 (implicit functions). I don't think it'll affect merging this because the changes to libogdi3.info are in different locations that what's done here.

dmacks (approved):

nieder:
Ping for @babayoshihiko again. Might be worth emailing him at Maintainer address.

dmacks:
> Ping for @babayoshihiko again. Might be worth emailing him at Maintainer address.

I just emailed him. Will wait a few days then push if no response.

PR #713, opened 2020-12-20
Spatialite update

dmacks:
This set of commits cleans up the builds of libspatialite2 and libspatialite5, converts the remaining users of 5 to use the latest (7), and fixes a bunch of build flaws along the way. With 2 and 5 both no longer used, they are converted to only their -shlibs package.


dmacks:
libspatialite7 possibly also needs build fixes, and could be converted to a parallel layout to these older libs for easier maintenance in the future. Once PR #712 lands, I'll look at it.

babayoshihiko:
Thank you!

Issue #650, opened 2020-07-25
librtmp->librtmp1 migration


dmacks:
Our librtmp (v2.3) has dependencies on openssl100 and FTBFS with openssl110, whereas librtmp1 (v2.4) does not have openssl dependencies at all. This is a tracker issue for updating package-suites with dependencies on the older libversion (see also #644)...

- [x] gst-plugins-bad-0.10 (Ben Reed; up for grabs)
- [x] libavcodec56-2.4-shlibs (snaggle)
- [x] libavcodec57-3.0-shlibs (snaggle)
- [x] libavcodec58-4.0-shlibs (snaggle)
- [x] libcurl4 (danielj)
- [x] mplayer (snaggle)

nieder:
Feel free to update libavcodec* and mplayer to librtmp1 as needed.

dmacks:
Just heads-up to @danielj gentoo has patch for librtmp openssl110 support:
https://gitweb.gentoo.org/repo/gentoo.git/tree/media-video/rtmpdump/files/rtmpdump-openssl-1.1-v2.patch

dmacks:
librtmp (old libversion) is orphaned now. Should I convert it to a -shlibs stub (as usual for this situation) or nuke it altogether (since it's an openssl100 user)?

nieder:
I say nuke. There's no reason to keep an unused libN that exposes a security issue.

dmacks:
*poof*

PR #716, opened 2020-12-23
libcurl4: new version, librtmp upgrade

dmacks:
Fixes a bunch of CVEs, and also in support of Issue #650


nieder:
Failed tests 1055 & 1383 on 10.13, but libcurl always fails some tests for me (current fails 1501), but otherwise seems ok.

dmacks:
It passed all tests on my 10.13. For the record, this commit removes some fink-specific disabling of certain tests because they were removed altogether from the new %v dist.

Issue #682, opened 2020-11-10
All pythons broken on Big Sur 11.0.1RC


nieder:
On the latest release for Big Sur (11.0.1RC), all pythons are broken:

python38 output:
```
Python build finished successfully!
The necessary bits to build these optional modules were not found:
ossaudiodev           spwd                  zlib               
To find the necessary bits, look in setup.py in detect_modules() for the module's name.
(Fink package build should have 2 missing)

make[3]: *** [sharedmods] Error 1
```

This is the python27 version:
```
Python build finished, but the necessary bits to build these modules were not found:
bsddb185           dl                 imageop         
linuxaudiodev      ossaudiodev        spwd            
sunaudiodev        zlib                               
To find the necessary bits, look in setup.py in detect_modules() for the module's name.
(Fink package build should have 7 missing)
```

In both cases, -lz is correctly found during ./configure and there are no errors during the build unlike in other python build failure reports.

nieder:
I made my own local zlib package for Fink and now the python packages build completely.

I wonder if this is related to this release note:
> New in macOS Big Sur 11.0.1 Release Candidate 2, the system ships with a built-in dynamic linker cache of all system-provided libraries. As part of this change, copies of dynamic libraries are no longer present on the filesystem. Code that attempts to check for dynamic library presence by looking for a file at a path or enumerating a directory will fail. Instead, check for library presence by attempting to dlopen() the path, which will correctly check for the library in the cache. (62986286)

For the record, libpng16 builds fine on 11.0.1RC1 w/ system-zlib, so the pythons are doing something differently.

Proposal: finally make our own zlib package, but hide it into a private prefix. If any package wants/needs to use fink-zlib, they can add the private prefix to *FLAGS as necessary. Otherwise, they can keep using system-zlib. This will prevent opportunistic linking.

@fink/core

dhomeier:
Python suggests this:
> There is a workaround: explicitly specify the SDK location to ./configure (you also need to specify the universal archs setting):

```./configure \
    --enable-universalsdk=$(xcodebuild -version -sdk macosx Path) \
    --with-universal-archs=intel-64 \
```
https://bugs.python.org/issue41116

nieder:
OK, that seems to work with at least python27. But compress-raw-zlib-pm is failing too:
```
t/01version.t ...... 1/3 
#     Failed test (t/01version.t at line 23)
#     Tried to use 'Compress::Raw::Zlib'.
[...]

PR #717, opened 2020-12-27
python37 - fix build on 10.15

PovlAbrahamsen:
Python37 wasn't building for me on 10.15.7: once it reached the stage where the modules were being compiled, setup.py was moving the system headers ahead of Fink's, causing compilation of the sqlite3 and readline modules to fail.

This fix comments out three offending lines in setup.py. It's been tested on 10.15.7, but should probably also be tested on 11.0 to ensure it doesn't break #682


nieder:
Running tests now. Really wish the python packages weren't such moving targets. Feels like we had just fixed them and then they broke again (Xcode changes?)

nieder:
This fixed python37 for me on 10.15. However, I can't test on 11.0.1 because even our current python37 is not working, even though it should according to #682. Now to figure out what's happening there.

nieder (approved):

nieder:
Haven't figured out why python* is again broken on my 11.0.1 VM (it's a fresh VM). But that shouldn't block fixing what we do support.

PR #718, opened 2020-12-28
libopenexr24 and ilmbase24 - fix pkgconfig files

PovlAbrahamsen:
Bug fixes for OpenEXR and IlmBase 2.4.2: the pkgconfig files omitted the prefix from the include/library paths...

Maintainer is @nieder


nieder (approved):
lgtm

PR #719, opened 2020-12-28
Gdal3 - and a few updates to packages that depend on it

PovlAbrahamsen:
Here is a new package for gdal version 3.2.0. This PR depends on #712 (for changes to gdal2) and #718 (for bug fixes in openexr24 pkgconfig files, which otherwise break compilation).

gdal-py now includes an extensive suite of tests (for gdal itself and gdal-py)! I've tested these with 2.7 and 3.7. There is one test in the transformations where a slightly larger error margin is required; and two places where there are known issues with MacOS that needed a slight tweak in the test scripts. And one script with a unicode character that causes an error with 2.7 (but not with later versions). In addition, the pytest version needs to be specified in the makefile if more than one type is installed. With these small caveats (four tests out of thousands), the library appears to work correctly and pass its tests.

I have also bumped the gdal dependence of gmt 4, 5, and 6, as well as mbsystem.

In gmt6-shlibs, "-DCOPY_DCW=FALSE" has been also added to the cmake flags, removing the file collision with dcw-gmt (which is a dependence - so the files will be there either way).

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgis.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgis.info
@@ -1,7 +1,7 @@
 Info2: <<
 Package: postgis%type_pkg[postgresql]
 Version: 3.1.0
-Revision: 1
+Revision: 2
 Description: PostgreSQL geographic object support
 
 # This version only supports postgresql >=9.6 (upstream has a moving
@@ -15,8 +15,8 @@ License: BSD
 Maintainer: Benjamin Reed <postgis@fink.raccoonfink.com>
 
 Depends: <<
-	gdal2-shlibs,
-	json-c-shlibs,
+	gdal3-shlibs,
+	json-c5-shlibs,
 	libgeos3.6.1-shlibs,
 	libgettext8-shlibs,
 	libiconv,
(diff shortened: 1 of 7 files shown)


dmacks:
Feel free to update postgis.info dependency gdal2->gdal3 also.

PovlAbrahamsen:
Done. I also changed json-c to json-c5.

dmacks:
I fixed (486a8762b83112c74661690c829ae883dbfcc9c4) a packaging weakness in gdal2 (surprise! Fink breaks upstream tests that use heuristics and indirect path detection). The first symptom was that the ./configure transcript mis-reported the version of libdap11. The configure.ac section in gdal3 appears the same, so this same fix to configure might be useful. It probably means the existing patchscript that fiddles with -ldap++ is no longer needed, as it would become an unused fallback.

dmacks:
gdal3 itself, and postgis12 and gdal-py38 on top of it built fine for me on 10.13. By eye, the others look like simple dep-bumps, so they should be fine too. In gmt6, you shouldn't need to bump the versioning in the Shlibs entries. The final components ("%n (>= 5.4.1-1)") define a dependency for when the listed filename at the given compatibility_version were first available. Since the filename and c_v are not changing, the dependency can remain the same as well.

PovlAbrahamsen:
But wasn't this a mistake in the original 6.1.1-1? The library versions with .6. didn't exist before gmt6 (and the first version in Fink was 6.1.1-1).

dmacks:
Good catch, yes! That mistake does go back to the original packaging of gmt6.

dmacks (approved):
lgtm

PR #720, opened 2021-01-01
Gettext 0.20.2

nieder:
Update our libgettext to 0.20.2 (including the private versions used by libiconv and dpkg).

This has the potential to fix the annoying build failures that pops up for some people that are using a 'weird' locale/language mix (such as en_AT).
The [release notes](https://savannah.gnu.org/forum/forum.php?forum_id=9430) say this:

> The interpretation of the language preferences on macOS has been improved, especially in the case where a system locale does not exist for the combination of the selected primary language and the selected territory.

On my 10.15 VM, I switched the region to Spain, but kept my language as English. During any build process, I got a lot of warnings about failure to set locale category LC_* to en_ES (but couldn't replicate the failures others were seeing). The warnings happened both during the actual compiling steps as well as while dpkg was creating buildlocks and building debs.
Once I updated my local libgettext8-shlibs 0.20.2, all the locale warnings went away.

I think this is a rather low risk update.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/base/dpkg.info
+++ b/10.9-libcxx/stable/main/finkinfo/base/dpkg.info
@@ -1,11 +1,11 @@
 Package: dpkg
 Version: 1.10.21
-Revision: 1250
+Revision: 1251
 GCC: 4.0
 BuildDepends: fink (>= 0.30.0)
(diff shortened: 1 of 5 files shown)


dmacks (approved):
lgtm (10.13)

PR #721, opened 2021-01-05
Kill ancient scribus

dmacks:
It doesn't recognize OS X > 10.10, builds fat or 32-bit-only on various platforms, and is quite old. And it has a BuildDepends:boost1.53.systempython that also hasn't existed since 10.10 either.

Along the way, also tweaking podofo9 so that its -dev (headers and .dylib symlink) has a dependency on its -shlibs (runtime code and the actual .dylib file). Upstream is planning to make a new release shortly.


dmacks:
This is an alternative to #125 (I'd rather see an update, but as an alternative I'd rather have nothing than a broken mess).

dmacks:
#722 includes the podofo dep fix and more.
#125 fixes scribus rather than killing it.

PR #722, opened 2021-01-07
Podofo fix

dmacks:
Existing package FTBFS due to incompatibilities with recent versions of cmake and newer cppunit. There's also a -I ordering error (global before local); I don't know how to fix it, so BuildConflicts to prevent a bad effect.

There is a newer upstream that includes at least the cmake fix, but this older version scratches my itch of actually having packages be compileable.


nieder:
Can you build with cppunit1.15.1 if you add to `-std=c++11` to CXXFLAGS ?

dmacks:
> Can you build with cppunit1.15.1 if you add to `-std=c++11` to CXXFLAGS ?

Ayup, with some upstream patches to compensate. This would also be cured by updating our package to the latest upstream version.

dmacks:
The library interface is completely unstable, so each new version needs a newly-namaed package suite. A new upstream version is reported to be coming soon, so if that can be packaged and suffices for building scribus (the only dependency), we could simply kill podofo9 as a "broken, old, unused" thing. Martin, if you're not interested in maintaining this lib (having dropped scribus in PR #125), I can pick it up.

costabel:
On 9/01/21 12:50, Daniel Macks wrote:
> The library interface is completely unstable, so each new version needs
> a newly-namaed package suite. A new upstream version is reported to be
> coming soon, so if that can be packaged and suffices for building
> scribus (the only dependency), we could simply kill podofo9 as a
> "broken, old, unused" thing. Martin, if you're not interested in
> maintaining this lib (having dropped scribus in PR #125
> <https://github.com/fink/fink-distributions/pull/125>), I can pick it up.

Please do.

--
Martin

dmacks:
At upstream report https://sourceforge.net/p/podofo/tickets/121/ another user proposed passing -std=c++11 to the whole build (not just to test/unit/CMakeLists.txt as based on https://sourceforge.net/p/podofo/mailman/podofo-users/thread/1501515201.2032.64.camel@litePDF.cz/)

PR #723, opened 2021-01-10
Cfitsio upgrade

dmacks:
From static cfitsio to shared libcfitsio8-dev/-shlibs. Along the way, overhaul libvips42's .pc to stop leaking dependencies on internal-use-only libs.

Pinging @nieder (libvips/nip2), @akhansen (fits-oct), @dhomeier (astropy)

There are a few other reverse-dependencies on old cfitsio, but they will involve more substantial updates.


nieder:
Fine for my packages. Other packages are simple dep upgrade away from static lib, so shouldn't be a problem if $maintainers don't respond.

PR #725, opened 2021-01-13
patched configure and R.m4 to include <stdlib.h> and <string.h> to av…

sth0:
patched configure and R.m4 to include <stdlib.h> and <string.h> to avoid implicit declaration no longer allowed by compiler in Xcode 12.3.
Updated patch file for R-base 3.6.3 and updated MD5 hash in .info file


dmacks (approved):
lgtm. Same config results with/without this change on 10.13; submitter says it also works on 10.14 and 10.15 ([Fink-users] R-base install fails on Mac OS 10.15). The rev-up isn't technically needed, but no serious harm.

PR #727, opened 2021-01-18
root5: Dep upgrades

dmacks:
Especially important here is a switch to using external libxxhash/lv4 (fink package dep), to avoid build-time
downloading of sources.


dmacks:
ping? Been a few weeks and it's just libversion bumps...I'll push it soon if no objections.

mommsen (approved):
Thanks for the ping. It slipped through the cracks, sorry.

It looks okay to me. Although this root version is very old.

dmacks:
Thanks! It is indeed old...I'm just doing a bunch of dependency audits and this popped up. I'm sure users woudl welcome a new version, but I'm not a root user myself.

PR #729, opened 2021-01-18
telnet: add Big Sur

htodd:
Just added Big Sur so I can get telnet.


nieder:
LGTM, as it's a simple enough change, but untested.

akhansen (approved):
LGTM too.  I can't test it right now, though--I need the incantation to have a working fink on Big Sur.

PR #732, opened 2021-01-23
Postgresql updates

nieder:
@wodev 
Updates for postgresql packages

For postgresql9X, had to patch configure to successfully build when using Xcode12 and it's implicit-declaration enforcements.

For postgresql1X, simple version bumps that contain a fix for a test in earlier releases that failed when not in daylight savings time.


dmacks:
These don't use any special mirrors for the sources, so the '''BuildDepends:fink-mirrors (>=0.28.7.1-2)``` can probably be removed.

wodev (approved):
looks good for me :-)
I check and remove the BuildDepends:fink-mirrors with the next update.

Thanks for the fix and update.

PR #734, opened 2021-01-30
Jpeg8 turbo update

nieder:
Update to 1.5.3
This is the last version that uses autotools.
Current upstream is 2.0.6 which is cmake based


nieder:
v2.0.6 builds and tests fine, but there's a minor change in the install_name.
The old autotool version installs these two dylibs (actual install_names):
`$PREFIX/libjpeg.8.dylib and $PREFIX/libturbojpeg.dylib`
But the new cmake version has:
`$PREFIX/libjpeg.8.dylib and $PREFIX/libturbojpeg.0.dylib`

The libturbojpeg dylib is no longer unversioned.

nieder:
Merging.
Minor %v bump

PR #735, opened 2021-02-01
pianobar: new version, major dep audit

dmacks:


htodd (approved):
Works for me, which is all the testing I usually do.

PR #736, opened 2021-02-01
Nieder upgrades (security-related deps)

dmacks:
Bump deps on libgcrypt and gnutls to their latest libversions.

There are large trees of packages with deps on older gpgme lib, not sure if it's a problem to update them piecemeal, so I did not touch here (or check to see if the ton of BDep without accompanying Dep are even needed). Also not touching boost deps for similar reason.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/mplayer.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/mplayer.info
@@ -1,6 +1,6 @@
 Package: mplayer
 Version: 1.4
-Revision: 2
+Revision: 3
 Description: Unix movie player
 License: GPL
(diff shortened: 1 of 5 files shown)


nieder (approved):
Seems fine assuming they've been tested.

nieder:
* For BDeps w/out matching Dep, kde4*-buildenv pushes `-Wl,-dead_strip_dylibs`, which often reduces overlinkage, but headers are still picked up.
* Bumping boost to any KDE package will require updating kde4*-buildenv, and a matching %r BDep on the users.
* IIRC, there were issues w/ updating gpgme (via qgpgme?), but it's been a long time since I looked.

dmacks:
Those other comments make sense. The kde build systems are fairly opaque to me.

PR #737, opened 2021-02-04
Grass proj nad2bin

babayoshihiko:
This is not finished work.  There still remains several errors regarding proj.

GRASS cannot build with the latest proj.
Fix is based on: https://trac.osgeo.org/grass/changeset/74181


babayoshihiko:
nad2bin is now a separated package. This seems an easy yet effective solution for older grass packages.

dmacks:
Sorry this has been idling so long! I support the approach of making a separate "nad2bin" package, breaking it out of the old version of proj so that it can be used along-side newer versions of proj. It appears to have been in the "proj-bin" package previously. The new nad2bin has:
```
Conflicts: <<
		proj (<< 4.8.0-5),
		libproj9 (<< 4.9.2-5)
<<
```
but I am not sure why. More importantly, if users have proj-bin-4.9.2-4 installed and try to install nad2bin, it seems like there will be a collision. To solve it, the new package should Replaces:proj-bin (<<4.8.0-7)

Unrelated to nad2bin, but given this is a change to grass64 and grass72 also, could you switch those packages from readline6 to readline8?

dmacks:
That looks fine now. Can you bump the readlineX versions?

dmacks (approved):
Looks good to go. There's something weird-looking in one of the readline flags, but easy enough to deal with that later if it even matters at all. Best to solve the main problems now.

PR #742, opened 2021-02-14
libaudiofile 0.3.6

nieder:
update to latest libaudiofile (from 2013).
Adds Ubuntu's patches which contain several fixes for CVEs

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libaudiofile1.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libaudiofile1.info
@@ -1,17 +1,49 @@
 Package: libaudiofile1
-Version: 0.3.4
-Revision: 2
+Version: 0.3.6
+Revision: 1
 Epoch: 1
 ###
 Depends: %N-shlibs (= %e:%v-%r)
+BuildDepends: <<
+  fink (>= 0.32.0),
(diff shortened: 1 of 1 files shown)


dmacks:
Do we need to actually install the static lib, or just have it built iff -m enabled?

nieder:
Probably fine to add `--disable-static` to `ConfigureParams` and `--enable-static` to `TestConfigureParams`.

TheSin-:
Honestly I like having the statics in the -dev package for all libs, local dev needs statics and it's kinda the point of the -dev package IMHO.  Maybe we should start a new policy to have -dev-statics for us devs that like and use them locally and make them so they can't have any deps on them or something.

I build a few pieces of software locally that I use statics for (not specifically this package) but I always have to mod info files which sucks.  I need to binary that I build to be statically built and not use dynamics to be easier portable.

nieder:
@TheSin- It's your pkg so can keep static enabled if you want. Review and approve and can merge.

TheSin- (approved):

TheSin-:
I haven't merged a package from the ui yet so I don't want to click the button.  If that how it's done for future requests though?

nieder:
Yep, just click the green "Merge pull request" button.

PR #743, opened 2021-02-14
push maven version to 3.6.3

eduardkareletc:
An updated .info file to update maven to the most recent version  (3.6.3).


nieder:
Thanks for the submission.

Validating the .deb fails with this error:
```
Error: package contains the shared library
          /sw/share/maven/lib/jansi-native/osx/libjansi.jnilib
       but the corresponding install_name and compatibility_version
          /Users/gnodet/work/git/jansi-native/target/native-build/target/lib/libjansi-1.8.jnilib 0.0.0
       are not listed in the Shlibs field.  See the packaging manual.
```
The prebuilt jnilib needs to 1) have it's install_name fixed to the actual install location, and 2) be listed as a [private library](https://www.finkproject.org/doc/packaging/policy.php?phpLang=en#sharedlibs) in the Shlibs: field. This would be fixed in InstallScript.

Also, searching through the installed files, `bin/mvn` and `share/maven/bin/mvn` reference the location `/etc/mavenrc`. This should be edited to point to %p/etc/mavenrc. See youtube-dl.info and youtube-dl.patch for a common way that the Fink prefix is added into files.

Also, feel free to change the Maintainer field to yourself. The previous maintainer has said it's OK to do that for his packages.

nieder:
Closed/outdated by dd5f7b1 (and later updated to v3.8.4).
dd5f7b1 borrowed heavily from here

PR #748, opened 2021-03-11
New package cppy-py as build dependency for kiwisolver; update to upstream 1.3.1

dhomeier:
I found that the kiwisolver 1.2.0 build fails if pip is not installed, and that it pip-installs a dependency `cppy` during the build phase, if it is. I think this goes against packaging rules, so I have added that package as `BuildDepends` (I checked that the tests are still running successfully after cppy has been removed).
Updated to the latest upstream along the way and also removed a patch to the tests that no longer seems required.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/cppy-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/cppy-py.info
@@ -1,0 +1,37 @@
+Info2: <<
+Package: cppy-py%type_pkg[python]
+Version: 1.1.0
+Revision: 1
+Description: C++ headers for C extension development
+License: BSD
+Maintainer: Derek Homeier <dhomeie@gwdg.de>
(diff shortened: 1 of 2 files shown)


dmacks:
Kocking out ```setup_requires...``` from setup.py will prevent auto-installation of deps.

dhomeier:
> Kocking out `setup_requires...` from setup.py will prevent auto-installation of deps.

I don't think that will be necessary as it is not going to install anything when the BuildDep is already installed.
Or do you think it might call `pip --upgrade` once a newer version shows up on PyPI, even with
`setup_requires=['cppy>=1.1.0']`?

dmacks:
I know zilcho about setuputils, just that's a way to stop auto-installing (trick I use when porting stuff to see where things fail later:). If you have the appropriate (Build)Depends, then I don't think there's harm to leave it.

nieder (approved):
Explicit BDep means we don't need to mess with setuptools dependency code.
LGTM

PR #750, opened 2021-03-14
Add libsigsegv2-shlibs to the depends list for grep.

dak180:
Should fix lib not found errors:
```
dyld: Library not loaded: /opt/sw/lib/libsigsegv.1.dylib
  Referenced from: /opt/sw/bin/grep
  Reason: image not found

```

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/grep.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/grep.info
@@ -1,6 +1,6 @@
 Package: grep
 Version: 3.6
-Revision: 1
+Revision: 2
 Maintainer: None <fink-devel@lists.sourceforge.net>
 
 BuildDepends: <<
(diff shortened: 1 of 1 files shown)


nieder:
1. Will sigsegv3 work?
2. sigsegv is an optional dep (grep compiles fine if not found). Do we want to have an extra dep for it? I'm agnostic.

dak180:
@nieder 
1. not sure, have not tested.
2. `sigsegv` will get pulled in if it is present when grep is compiled so it needs to be listed as a dependency, a build conflicts or explicitly configured off; I would be happy with any of these, this just seemed the simplest to do.

dak180:
@nieder initial investigation points to `grep` not using `sigsegv3` when present.

nieder:
I pushed a change to explicitly disable use of libsigsegv in grep to keep the dependencies down. If someone explicitly needs/wants it, we can revisit. Thanks for finding this issue.

PR #763, opened 2021-04-02
pasdoc: update to upstream version 0.16.0

kamischi:
pasdoc-gui: use lazarus-cocoa and the included script to create the app-bundle.


dmacks:
```
Source: https://github.com/%n/%n/archive/refs/tags/v%v.tar.gz
SourceDirectory: pasdoc-%v
```

Lots of github sources have this URL style, and it means we get a ton of /sw/src/ files that are just version-numbers. I think it would be nice (more clear to users, and less likely to generate collisions among "save version number" of different packages) to use ```SourceRename``` so that the tarball winds up including %n. If you do:
```
SourceRename: pasdoc-%v.tar.gz
```
then you also no longer need the ```SourceDirectory```.

dmacks (approved):
lgtm

PR #769, opened 2021-04-04
Autotrace update

dmacks:
New upstream with new version that is same install_name but seems to have backward-incompatible interface changes. So I offloaded to a subdir for install_name.

Also fixed some reverse-depends from long ago, when "autotrace" was monolithic instead of having the runtime programs in a -bin splitoff (for ImageMagick suites).


nieder:
Ping @ballapete

nieder:
No response after a month. Seems OK, but untested.

PR #770, opened 2021-04-06
cernlib2006: fix FTBFS: no bin/makedepend or libXp.dylib in xquartz-2.8.0

dmacks:
So disable header-tracking and avoid linking against a lib that isn't directly used anyway.


dmacks:
cernlib2006 is no longer being developed, but at some point might want to look into bumping from gcc9 to gcc10. I don't use this thing, just caught it in a check for things broken by the latest XQuartz release.

dmacks:
Quick-fix for FTBFS against current XQuartz.

PR #772, opened 2021-04-17
grace: updates for Xcode12

nieder:
Fix implicit declaration errors during configure
bump to netcdf-c18 because -c7 deps on hdf5.10, which doesn't build with Xcode12 (more implicit declarations).


nieder:
@akhansen ping
Will push these in a day or two since it's a simple build fix.

akhansen (approved):
Looks fine to me.

PR #775, opened 2021-04-22
New GRASS version (grass78)

dmacks:
This is the latest upstream release, which is now compatible with proj19 (where nad2bin no longer exists, avoiding need for PR #737). Assuming this package works (I only tested that each variant runs as far as opening a file), we could scrap the older grass64 and grass72 rather than having to add more hacks to keep them buildable and/or retain old versions of yet other packages to support them.


babayoshihiko:
I've successfully built grass78-mac.

When I clicked GRASS-mac-7.8.app, nothing happened.

I ran grass.sh and got an error:
Failed to load image from file ".gui/images/startup_banner.png" but the file existed.

I ran grass78 and got an error:
ERROR: Path '/Volumes/HDD/opt/sw/Applications/GRASS-mac-7.8.app/Contents/MacOS/<UNKNOWN>/<UNKNOWN>' doesn't exist

nieder:
On 10.13:
1) install grass78-x11
2) double click .app does nothing.
3) run grass.sh from Terminal: gui starts (ugly python3.8 icon in dock). Interestingly, Xquartz doesn't start.
4) run GRASS from Terminal, which is what Info.plist calls:
```
$ GRASS-x11-7.8.app/Contents/MacOS/GRASS 
2021-04-25 06:08:56.972 GRASS[43201:74738785] No Info.plist file in application bundle or no NSPrincipalClass in the Info.plist file, exiting
```
5) run grass78:
```
Starting GRASS GIS...
WARNING: GRASS GUI not found. Please check your installation.
Switching to text based interface mode.

Hit RETURN to continue.


ERROR: Path '/sw/Applications/<UNKNOWN>/<UNKNOWN>' doesn't exist
Exiting...
```

Try switching to grass78-mac and same behavior as above.

grass72-mac does work by double clicking the .app icon.

PR #778, opened 2021-04-26
Cran 20210426

babayoshihiko:
Regular updates. Removed .so from pkgload.


nieder (approved):
Untested but look OK.

PR #780, opened 2021-05-03
latex2html: new version

dmacks:
Fix for "[Fink-users] Latex2html too old version ?"


sth0:
LGTM, basically what I had plus implementing the patch file.  The one thing I did not understand was the ppdf variant.  I didn't see how it built the package with a different option.  It looked to me like that type descriptor did nothing.

sth0 (approved):

sth0:
Should have mentioned that it converted a simple latex file into readable html....  Not extensively tested, but it didn't crash and produced expected output.

nieder (approved):
LGTM

PR #781, opened 2021-05-04
flex transition

nieder:
Moves our old flex-2.5.4a to new package flex-legacy.
Updates flex to latest flex-2.6.4 (same as flex-devel).
Transitions all packages using flex-devel to flex @ v2.6.4.
Removes unused package flex-devel.

Tested that all packages currently depending on flex-2.5.4a will build with flex-2.6.4.
Tested that all packages using flex-devel will also build with flex-2.6.4

Nothing currently uses flex-legacy, but keeping it just in case (similar to autoconf2.13-legacy, texinfo-legacy, bison-legacy)


dmacks:
I support this re-orgnization of the flex packages. Thanks for doing the heavy-lifting and rebuild-testing. Should be good to go unless someone else speaks up soon...

PR #785, opened 2021-05-07
xtide: use latest harmonic data files

PovlAbrahamsen:
A trivial update to xtide: use the latest data files (released in January). Maintainer is @akhansen


dmacks (approved):
$maintainer okayed via IRC

PR #786, opened 2021-05-09
enblend-enfuse 4.2

nieder:
bump deps
switch from cmake to autotools because upstream forgot a couple CMakeLists.txt files in subfolders

@pete-woods


nieder:
Ping. Will push soon if no response.

pete-woods (approved):

nieder:
thanks

PR #788, opened 2021-05-11
libplist: Update to upstream version 2.2.0

kamischi:
Please test on other systems than mine (11.3).

libusdmux and libimobile will follow.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libplist-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libplist-py.info
@@ -1,84 +1,54 @@
 Info2: <<
 Package: libplist-py%type_pkg[python]
-Type: python (2.7)
-Version: 1.10
-Revision: 2
+Type: python (2.7 3.4 3.5 3.6 3.7 3.8)
+Version: 2.2.0
+Revision: 1
 Description: Python bindings for libplist
 License: GPL/LGPL
 
 # Dependencies:
 BuildDepends: <<
   fink (>= 0.24.12),
-  cmake, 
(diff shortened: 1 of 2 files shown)


dmacks:
It's more portable to have the package-name of a library have the version value from the install_name of the library rather than the compatibility_version. Fink policy is a 1:1 correpondence of filenames to package-names, whereas c_v is allowed to increase within the same package-name.

dmacks (changes requested):
It's more portable to have the package-name of a library have the version value from the install_name of the library rather than the compatibility_version. Fink policy is a 1:1 correpondence of filenames to package-names, whereas c_v is allowed to increase within the same package-name. So "libplist2.0.3" rather than "libplist7", and likewise for the -shlibs. It's a coincidence that the previous package had "1" for both i_n and c_v.

nieder:
The symlink `libplist-2.0.dylib` is just the equivalent of `libFOO.dylib` with other libraries. Upstream seems to consider `libplist-2.0` to be the name of the library. That leads to libplist2.0.**3**-shlibs, where `3` is the install_name library value.

(FWIW, Debian calls the package libplist3)

kamischi:
Oh. i just noted that accidentally i did not use separating dots. Is it OK or do you prefer with dots. I do not really mind.

nieder:
Sorry, lost track of this.
Either of `libplist2.0.3` or `libplist20.3` would be preferable over `libplist203`. It makes it easier to see what parts of the package name represent the library name and the install_name.
Also, don't forget to add the new library to the Conflicts/Replaces fields of libplist1

PR #789, opened 2021-05-11
iOS tools: Update to upstream version

kamischi:
usbmuxd and libusbmuxd have separate sources.


kamischi:
I need to work on the version numbers of the info files and wait for libelist to be finished.

dmacks:
If possible, could it switch from libzip2 to libzip5?

kamischi:
Propably. I have to admit that i have somewhat abandoned libplist,  usbmuxd and friends. There are probably upstream updates. You want to get rid of libzip2 and usbmuxd is the only one left over?

kamischi:
It now builds.

nieder:
seems fine at first glance but needs #788 to change the package name from libplist203 to libplist2.0.3 as discussed there.

PR #790, opened 2021-05-12
New pkg perl5302 to match system-perl on 11.3

nieder:
Tested on 10.14.6 and passes all tests. 
dist/Time-HiRes/t/stat occasionally fails like this:
```
dist/Time-HiRes/t/stat ......................................... #   Failed test at t/stat.t line 35.
#     Structures begin differing at:
#          $got->[8] = '1620781011.01701'
#     $expected->[8] = '1620781010.81523'
# Looks like you failed 1 test of 43.
FAILED at test 27
```
but it is not reproducible.

I've temporarily lost access to my VMs, so please check with other systems, especially macOS 10.12 and macOS11.0/1/2.


nieder:
I've also built extutils-makemaker-pm5302, module-build-pm5302, and moose-pm5302 and their accompanying dep trees against perl5302 on 10.14.6 w/ no errors (only 2 version bumps to deal with the loss of '.' from `@INC`).

nieder:
Ping

nieder:
I'll commit this at some point today unless there's an explicit objection. We can try to figure out the intermittent test failure, but need to get perl5302 going to allow many system-perl packages on `Dist:11.3`

dmacks:
No objection. I don't have free CPU to test it at the moment, but let me know if you need 10.13 and I'll see what I can arrange. I agree we definitely need to get this in.

PR #791, opened 2021-05-15
io-compress update

nieder:
update to various compress perlmods
Bump to latest upstreams
Add pm5184 and pm5282 as needed

io-compress-lzf had a note about newer releases failing tests, but I tested on 10.14.6 and all tests passed for all perl variants.


dmacks (approved):
This suite is usually pretty robust with the exception of the noted problem, so if that's okay now, hooray! All inter-package dependency versions updated according to Makefile.PL or META.yml? lgtm by eye

nieder:
Yes, I confirmed there were no changes to deps according to the build/packaging descriptions from upstream.

PR #795, opened 2021-05-23
FreePascal and Lazarus update to new upstream versions.

kamischi:
Besides FreePascal and Lazarus it includes
avr-embedded-binutils
qt4pas
qt5pas
xdev-toolkit


dmacks:
Looks good by eye, especially the 32-bit platform restriction. Don't have time to check in detail at the moment. In the future, could you avoid using '&' in branch-names? Makes it a little trickier to use command-line tools.

dmacks (approved):
Seems safe enough.

PR #796, opened 2021-05-25
build fixes and cleanup for r-base3[3-6]

nieder:
* Add libxt as an explicit dep to avoid using libxt-flat.
* Add upstream patch to fix implicit warning errors during the detection of Accelerate.framework on 10.15. Otherwise, package tries to build libRblas.dylib.
* Restrict r-base[3-5] through Dist=10.14.5 to match what we're already doing for the rmods.


babayoshihiko:
Looks good.

babayoshihiko (commented):
Confirmed.

PR #800, opened 2021-06-18
xrootd4: new version

dmacks:
This is likely the last xrootd v4 release upstream.


dmacks:
Ping @mommsen. Will push in a few days if no response.

PR #801, opened 2021-06-18
icon: new version, switch to 64-bit arch

dmacks:
The .so is built flat installed in bin/ as in previous versions; I
have no idea how this file might get used so I don't know if the
location and/or binary library flags are correct.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/icon.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/icon.info
@@ -2,7 +2,7 @@ Package: icon
 Version: 9.5.20i
 Revision: 1
 Depends: x11
-BuildDepends: x11-dev
+BuildDepends: fink-package-precedence, x11-dev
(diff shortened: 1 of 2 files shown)


dmacks:
Ping @microtherion. Will push in a few days if no response.

microtherion (dismissed):
👍 this all looks plausible to me (haven't used it in years).

I don't know the reason for the `.so`, nor do I think I ever knew (possibly having to do with an extension mechanism?).

Only question I have is why the dependency on readline was dropped; ISTR that icon has a REPL, so readline is potentially useful. Did our policies on this change?

PS: Feel free to take over as maintainer, if you wish.

dmacks:
Doesn't access readline headers or link against anything except x11. I added some instrumentation to the build to help catch it.

I don't actually use this thing at all, I'm just working on upgrading everything that used older versions of readline.

dmacks:
Maintainer seems okay; I don't want to maintain it, just minimal work to keep existing packages modernly buildable

PR #802, opened 2021-06-19
CRAN: New upstream version

babayoshihiko:
Regular updates for CRAN packages.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-ade4-r.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-ade4-r.info
@@ -33,14 +33,14 @@ Distribution: <<
 	(%type_pkg[rversion] = 35 ) 10.14.5
 <<
 Type: rversion (3.6 3.5 3.4 3.3 3.2 3.1)
-Version: 1.7-16
+Version: 1.7-17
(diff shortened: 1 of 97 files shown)


nieder:
Do we need to add R-3.2 to cran-devtools-r-2.0.1.info?

nieder:
* cran-recipes-r.info changes the Depends field, so needs to bump %r.
* cran-testthat-r32 went from 3.0.2 to 2.2.1 when you bumped the main .info file to %v=3.0.3. Will need a separate .info for 3.0.2 that's r32 only.

Looks fine for the other ones, but untested.

babayoshihiko:
>     * cran-recipes-r.info changes the Depends field, so needs to bump %r.
DONE


>Do we need to add R-3.2 to cran-devtools-r-2.0.1.info?
>     * cran-testthat-r32 went from 3.0.2 to 2.2.1 when you bumped the main .info file to %v=3.0.3. Will need a separate .info for 3.0.2 that's r32 only.

As for devtools and testthat, -r32 variants did not get built. And the same for the previous version (devtools 2.4.0, testthat 3.0.2). So I reverted to the version that could be built (devtools 2.0.1, testthat 2.2.1).

nieder (approved):

PR #803, opened 2021-06-20
Add mips-binutils.

kamischi:


dmacks (approved):
Plugs a dep-tree hole

Issue #831, opened 2021-07-07
Please update vulnerable Expat 2.3.0 to 2.4.1


hartwork:
https://blog.hartwork.org/posts/cve-2013-0340-billion-laughs-fixed-in-expat-2-4-0/

Thank you!

dhomeier:
Thanks for the alert and updates! Linked PR is in.

hartwork:
@dhomeier thank you!

dmacks:
Thanks for the upstream-alert!

PR #832, opened 2021-07-07
Update expat1 to fix Billion Laughs vulnerability

dhomeier:
To fix DNS vulnerability [CVE-2013-0340](https://marc.info/?l=oss-security&m=136580776324285&w=2,) as reported in #831 updates (incrementally) to the upstream versions 2.4.0/2.4.1.
Build otherwise unchanged (except for switch to tar.xz source), passes both tests on Mojave 18.7.0.


dmacks (approved):
lgtm on 10.13

PR #834, opened 2021-07-11
postgresql updates

wodev:
- update to latest upstream version
- changed readline dependency to readline8
- moved pkgconfig files to dev split off
- added next postgresql major version 13 in replaced by and conflict section (dev splittoff)
- removed dependency to fink-mirrors (>= 0.28.7.1-2)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
@@ -226,6 +226,7 @@ SplitOff: <<
 		postgresql13-dev
 	<<
 	Replaces: <<
+		postgresql10 (<< 10.17-1),
 		postgresql92-dev,
 		postgresql93-dev,
 		postgresql94-dev,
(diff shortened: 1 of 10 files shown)


nieder:
Good catch on the pkgconfig files.
Because files are migrating from one splitoff to another, you need to add `Replaces: postgresqlXX (<< ver-rev)` to the -dev packages (don't need to tweak `Conflicts`). Where ver-rev are the pkg version and revision where the switch happened.

For example (for postgresql94):
```
	Replaces: <<
		postgresql94 (<< 9.4.26-2),
		postgresql92-dev,
		postgresql93-dev,
		postgresql94-dev,
		postgresql95-dev,
		postgresql96-dev,
		postgresql10-dev,
		postgresql11-dev,
		postgresql12-dev
		postgresql12-dev,
		postgresql13-dev
	<<
```
This will let Fink install the latest postgresql94-dev for this package and not complain that the previous postgresql94 already owns the files in `%p/opt/postgresql-94/lib/pkgconfig`.

wodev:
Thanks for the hint. I added the postgresqlXX  to the Replaces list in the dev splittoff with the  latest commit 8917be2 
Also I moved the order lib/pgxs to the dev splitoff, because  the contains Makefiles and testifies for Postgres extension development.

Also thanks to @dmacks  for pointing out the readline dependency update and the misplaced pkgconfig files.

dmacks (approved):
lgtm. Sometime we should think about turning off the -dev of some of the very old versions, as they become completely unused except for library bindings themselves (see #834).

PR #836, opened 2021-07-12
Kf5 5.44

nieder:
Bump KDE Frameworks from 5.8 to 5.44 (last version to support our Qt5.7).
Bumps 3rd party deps where possible


dmacks:
No idea if it's true in this new version, but in the existing (10.9/10-only, in stable) package-set, I had a ton of FTBFS because of missing qmake. I solved it by adding ```BuildDepends:qt5-mac-qtbase-dev-tools```. Affected every kf5 package I had tried.

nieder:
Been a month w/ no further comments.

PR #839, opened 2021-07-17
Libzip overhaul

dmacks:
Main functional change is that the userland binaries package (zimcmp, zipmerge, ziptool) is now not lib-versioned and uses ">=" not "=" on -shlibs (avoids several deadlock situations). Is "libzip-bin" the best name (was libzipN-bin), or should it instead be "zip-bin" or "zip"? Debian has a separate package for each of the three binaries, which seems excessive to me.


nieder:
`zip` is unclear since none of the executables are called zip. And it makes it seem like it's the opposite of unzip, which it's not.

Hanspeter

dmacks:
Keeping it libzip-bin then.

PR #840, opened 2021-07-19
imlib2: new version, propagate x11 flag (in public API)

dmacks:


nieder:
Seems fine, but I thought we were supposed to avoid `Requires.private` because pkg-config forces their presence but the package itself doesn't use it?

dmacks:
The package itself does use x11 in its public API, not just an internal-implementation detail. Without -I for x11 in CPPFLAGS, the installed .h for this lib cannot be used. So I think it exactly should force its presence. It just avoids others from having to know to detect and pass that flag themselves. ```Requires.private``` means the -I of the dependency get propagated for compiling but not the -L or -l for linking (so it's better than ```Requires``` that also propagates linker details).

nieder (approved):

PR #841, opened 2021-07-19
libopusfile: fixing some deps and -I ordering

dmacks:


nieder:
Looks fine. Does it actually use openssl headers or only checks for it (superfluous ./configure check) but doesn't do anything with the package?

dmacks:
According to otool -L, libopusurl.0.dylib really does link against libssl and libcrypto. The libopusurl0-shlibs splitoff had a pre-existing ```Depends:openssl110-shlibs``` but no BDep to match.

nieder:
Helps if I read the entire .info

nieder (approved):

PR #843, opened 2021-08-01
Texlive 2017

nieder:
This is a replacement to #798 that updated Texlive to 2021. Because that release had problems with how the configuration utilities worked, I've decided to do an incremental update to 2017 from our current 2014.

In the [original thread](https://sourceforge.net/p/fink/mailman/fink-users/thread/A1BEBC23-312E-4727-BF53-4827B1EFBA44%40math.purdue.edu/) describing the issues with texlive2021, the primary failure was in texconfig, namely, running the command `texconfig paper letter`. This now works as expected, both using `texconfig-sys` and `texconfig-user`.

All tested on 10.14.6. Please test elsewhere.


nieder:
I'm getting an undeclared INT_MAX error on 12.1. Does the following code compile and run on not-macOS 12?
```C
#include <stdio.h>

int main(void)
{
    printf("INT_MAX = %d\n", INT_MAX);
    return 0;
}
```

dmacks:
On 10.13:
```
foo.c:5:30: error: use of undeclared identifier 'INT_MAX'
    printf("INT_MAX = %d\n", INT_MAX);
                             ^
1 error generated.
```
Resolved by <limits.h>, and the output is `INT_MAX = 2147483647`

nieder:
OK, that's the expected error and solution. I wonder why I wasn't getting it when building texlive2017 on 10.14.5? I guess it could have been an indirect include on <limits.h> that now needs to be explicit.

dmacks:
ptexenc1 needs `BuildDepends:pkgconfig`. I added f-p-p to libkpathsea6, which turned up nothing additional.

dmacks:
I just pushed poppler63, which should be new enough to use (per "system-poppler needs >= 0.30") note. And poppler63-xpdf (drop-in replacment for some xpdf uses) supplies GfxFont.h.

nieder:
> ptexenc1 needs `BuildDepends:pkgconfig`. I added f-p-p to libkpathsea6, which turned up nothing additional.

Fixed.

> teckit doesn't actually use expat for anything because the "sfconv" is program is not compiled (disabled via patch). Any idea why it's disabled? Even if it were enabled, it would only be a Depends of the main package, not the -shlibs.

This was the failure when building sfconv:
```
../SFconv/SFconv.cpp:49:10: fatal error: 'expat/xmlparse/xmlparse.h' file not found
```
That file dates back to expat-v1 from the 90's. Code hides `#include "config.h"` for finding current <expat.h> behind an `#ifndef platformUTF16`. https://github.com/silnrsi/teckit/issues/29
So I just patched the file name and it works.

> texlive `BuildConflicts:luaXX` and then rolls its own lua52. Can it use the external fink one?

I don't see an easy way. There's no ./configure flag for system-lua and its activation (for luatex use) is via nested configure calls.

dmacks:
Great detective-work on sfconv! Lua52 is already not-current, so I don't think it's a blocker to roll its own for internal use.

dmacks (approved):
lgtm

PR #845, opened 2021-08-14
PostgreSQL updates 12.8, 11.13, 10.18, 9.6.23

wodev:
PostgreSQL updates to latest upstream version

The upstream updates contains a fix for security issue [CVE-2021-3677](https://www.postgresql.org/support/security/CVE-2021-3677/) Memory disclosure in certain queries

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgresql10.info
@@ -1,6 +1,6 @@
 Info4: <<
 Package: postgresql%type_pkg[postgresql]
-Version: 10.17
+Version: 10.18
 Revision: 1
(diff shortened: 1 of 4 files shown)


dmacks:
I'm happy to see ongoing security fixes for oooold versions, and I wouldn't be surprised if some servers still use those old versions (and likewise previously-compiled programs use the older shared libraries). Do we need to continue to support compiling against old libs, or maybe we could scrap their -dev portions? Within fink, everything has migrated to pg12. The only question would be if there are non-fink uses of fink libraries that need that legacy. It seems like pg12 clients can still talk to older servers.

dmacks (approved):
lgtm, but waiting a day or two more in case @wodev wishes to respond to question about fate of older versions.

PR #848, opened 2021-08-22
Fix regexps for Big Sur darwin-version

dmacks:
Thanks Nieder for noticing the problem and finding a quick-fix!


dmacks:
Hold on...there's a regexp mistake and also a filestamp-timing inconsistency problem.

nieder:
Looks fine. Do the changes propagate into a package if you run autoreconf?

dmacks:
That patched m4/libtool.m4 is what gets copied into a build and then used to generate ./configure. It seemed to give a configure that had the same changes as the changes to libtool's own ./configure, which was copied from mysql's.

PR #850, opened 2021-08-26
New version for libgeos

babayoshihiko:
The latest version (3.9.1) of GEOS as a new package.


dmacks:
We're accumulating a ton of libgeos variants. Time to strip down the older versions that are not used any more? See #854

babayoshihiko:
I agree. 

I will rewrite .info
I will check all the packages that depend on geos.

dmacks:
The .info specifies `Package:libglwgeom`? Also, please see the most recent libgeos*.info in Fink (changed via #854). It would be useful for future upgrades if new libgeos versions followed that .info pattern (the -shlibs is the parent, the headers package is the splitioff).

dmacks:
@babayoshihiko ping?

dmacks (changes requested):
```
Validating package file libgeos3.9.1.info...
Warning: Incorrect filename 'libgeos3.9.1.info'. Should be one of:
	libglwgeom.info
	libglwgeom-3.9.1.info
	libglwgeom-3.9.1-1.info
```
Please see my comments from January 6.

PR #852, opened 2021-09-02
Perl5340

nieder:
New package perl5340.
The intent is to use this as a constant perl across all distributions for packages such as icon-naming-utils or intltool40 that require a perl tool, but we are currently forcing to use system-perl. The current policy leads to upgrade confusion and .info duplication and is a pain to maintain.
This perhaps could eventually be used for the fink tool itself: the Catalina release notes say that "[s]cripting language runtimes such as Python, Ruby, and Perl are included in macOS for compatibility with legacy software. Future versions of macOS won’t include scripting language runtimes by default"


dmacks:
Building on 10.13:
```
Checking your choice of C compiler and flags for coherency...
I've tried to compile and run the following simple program:

#include <stdio.h>
int main() { printf("Ok\n"); return(0); }

I used the command:

	gcc -o try -O3 -I/sw/include -fno-strict-aliasing -pipe -fstack-protector-strong -I/sw/include -L/sw/lib -fstack-protector-strong try.c -lpthread -lgdbm -ldbm -ldb -ldl -lm -lutil -lc
	 ./try

and I got the following output:

ld: library not found for -lgdbm
I can't compile the test program.
You have a BIG problem.  Shall I abort Configure [y]  
Ok.  Stopping Configure.
```

dmacks:
The perlXXX-core Provides list is out-of-date (also on some of our existing ones). We need to un-list:

* CGI: deprecated in 5.19.7, removed in 5.21.0
* CPANPLUS and CPANPLUS::Dist::Build: deprecated in 5.17.9, removed in 5.19.0
* Module::Build: deprecated in 5.9.4, removed in 5.21.0
* Module::Pluggable: deprecated in 5.17.9, removed in 5.19.0
* Package::Constants: deprecated in 5.19.6, removed in 5.21.0
* Pod::Parser: removed in 5.31.1

I'm not entirely sure what "deprecated" means; it's annotated "will be CPAN-only". We also don't have an exhaustive list of what is included. I'm not sure how this subset was chosen long-long-ago. Since we have a brand new perl version, should we just scrap the list altogether?

dmacks:
Focusing on build-fail, adding BuildDepends:gdbm6 obviously resolves it. But then its self-test fails:
```
ext/GDBM_File/t/count ............................................ skipped
ext/GDBM_File/t/fatal ............................................ skipped
ext/GDBM_File/t/gdbm ............................................. gdbm_firstkey: Item not found at ../../t/lib/dbmt_common.pl line 52.
Compilation failed in require at t/gdbm.t line 6.
# Tests were run but no plan was declared and done_testing() was not seen.
# Looks like your test exited with 255 just after 5.
FAILED--no leader found
ext/GDBM_File/t/opt .............................................. skipped
```
The GDBM_File perlmod is reported to be part of perlcore since forever but fink has never had deps for it. According to otool-L, GDBM_File.bundle is properly linked to /sw/lib/libgdbm.6.dylib

nieder:
On my 10.14.6, perls 5182, 5282, 5302, and 5340 are linking to db53-aes and gdbm4 (because those were the dev packages that were present) and passed all tests. I've also never gotten a build failure, but obviously haven't paid attention to what gdbm and db were around.

```
Checking for optional libraries...
What libraries to use? [-lgdbm -ldbm -ldb -ldl -lm -lutil -lc]  
```
I think by chance I've always built with gdbm4 around. With gdbm6, perl5182 and 5340 fail tests, but 5282 passes. Ugh.

Looks like just removing gdbm from this line
https://github.com/Perl/perl5/blob/e5a468d33677d8d809b0b67b8f21b9ad022c96c9/Configure#L1504

skips checking/building/testing for gdbm.

Does using gdbm4 work for you with no other patching?

Our options seem to be: block gdbm support (and prob db as well), or be explicit using Depends.

dmacks:
Now it passes tests against gdbm6 (still need to decide if we want to haul in those deps vs keep perl as a completely self-contained package such that it could be the lowest-level dep for the fink tool itself).

dmacks:
Starting fresh with a totally fink-controlled perlversion (no apple) lets us fix some legacy problems. MakeMaker installs a ".packlist" file for each package, which contains the fink temproot (%d) pathnames. Not sure if it's a bug in ExtUtils::MM_Unix that it uses the install-path instead of the runtime-prefix, but I think we should either start un-bugging that, or else ([as debian does](https://sources.debian.org/patches/perl/5.32.1-4/debian/no_packlist_perllocal.diff/)) hack that .pm to skip doing .packlist altogether. Or maybe fink should pass whatever env-var or parameter propagates NO_PACKLIST?

nieder:
There's a new gdbm-1.21. Feel free to update it to see if it helps things.
For deps, I think it's OK having linkage to gdbm and db. I would worry about losing functionality of whatever those two libraries bring, but we'll probably need to trim it if we want to make an essential package.

nieder:
NO_PACKLIST gets added when running `perl Makefile.PL NO_PACKLIST=1 ...`. Adding it as part of Fink's CompileScript command will change all debs (but it's not like .packlist is actually used for anything). If we don't want to fix it (why bother?), then axe it and borrow Debian's patch.

dmacks:
I guess .packlist hack would have to be "perl >= 5.34.0" (and also in extutils-makemaker-pmXXX separate package). Looks like some linux distros scrap them in various ways, but there's at least one tool (by google, no idea its real-world uptake) that uses them.

dmacks:
Debian does include File_GDBM.pm and dependency on libgdbm in their main perl package but Apple does not, and our older perl-versions have not. I'd advocate not adding (or not making deterministically added:) new deps for what we might want to remove later and that could be deterministically made external if someone wanted it.

nieder:
This is my current install of perl5182-core (when it opportunistically detects gdbm4). perl5282 and 5302 give similar output.
```
nieder $ dpkg-deb -c perl5182-core_5.18.2-1002_darwin-x86_64.deb | grep -i gdbm
-rw-r--r-- root/admin     5160 2021-05-31 19:32 ./sw/lib/perl5-core/5.18.2/man/man3/GDBM_File.3pm
drwxr-xr-x root/admin        0 2021-05-31 19:32 ./sw/lib/perl5-core/5.18.2/darwin-thread-multi-2level/auto/GDBM_File/
-r--r--r-- root/admin        0 2021-05-31 19:32 ./sw/lib/perl5-core/5.18.2/darwin-thread-multi-2level/auto/GDBM_File/GDBM_File.bs
-r-xr-xr-x root/admin    24756 2021-05-31 19:32 ./sw/lib/perl5-core/5.18.2/darwin-thread-multi-2level/auto/GDBM_File/GDBM_File.bundle
-r--r--r-- root/admin     1551 2021-05-31 19:32 ./sw/lib/perl5-core/5.18.2/darwin-thread-multi-2level/GDBM_File.pm
```
I'm fine with deterministically knocking it out. I don't know what we lose by not having gdbm database access.

dmacks:
My .patch here fixes perl against gdbm6 (v1.21 is PR #853 if you want it).

nieder:
Was the new list determined programatically or manually? Thanks for checking older perls too.

dmacks:
I manually ran `corelist` on each entry in Provides. That tool comes as part of Module::CoreList, a database of every version of every module supplied with every perlversion.

nieder:
On CentOS7, which has /usr/bin/perl-5.16.3:
```
$ perl -V:libperl -V:useshrplib
libperl='libperl.so';
useshrplib='true';
```
This has about 90 .so files in the perl5 directory (all part of the default. I don't think there's been any manually installed mods). ldd doesn't report any of them linking to libperl.so

nieder:
Closing this because full-featured perl5341 is now in fink-dists and @dmacks is working on perl5400 as a minimal perl.

PR #853, opened 2021-09-07
Gdbm update

dmacks:
New version of gdbm6, gettext fixes for gdbm4 and gdbm6


dmacks:
gdbm6 self-test fails for me on 10.13 iff dejagnu is installed, both the existing gdbm6-1.20 and this new 1.21 version. But it passes if I then run it manually in the fink build-dir. So TestConflicts:dejagnu inhibits the test that fails.

nieder (approved):

nieder:
Missed this one. sorry

PR #854, opened 2021-09-11
Libgeos cleanup

dmacks:
Turn off the headers packages of older version that are not used by any other package.
Invert the .info layout so -shlibs is the parent (headers package as splitoff), which enables turning off the headers packages. I even did this for the recent (in-use) versions, so that it is easier to look at differences among the versions (easier maintenance) and turn off the headers packages when the time comes.


dmacks:
Seems straightforward cleanup, no response from maintainer.

PR #862, opened 2021-12-25
New packages python39 and python310

nieder:
Adds new packages python39 and python310.
Python-3.9 is the first upstream to directly support macOS 11 and macOS 12.
We haven't been able to get python38 and earlier to work on  macOS 11/12, so this is the first python that builds and runs on these systems. Latest python-3.8.12 *might* support macOS11/12, but not sure yet.

I removed a lot of the OS-versioned hacks (mostly around finding the SDK and system library cache) that had accumulated into the earlier pythons to work with newer macOS because the underlying issues had been taken care of upstream.

Also cleaned up the lists of tests to just those that currently are known to fail (on macOS12).

Tested on just macOS 12.1.
Please test on anything before 10.14.5.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/python310.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/python310.info
@@ -1,0 +1,268 @@
+# -*- coding: ascii; tab-width: 4; x-counterpart: python310.patch -*-
+Info2: <<
+Package: python%type_pkg[python]
+Version: 3.10.1
+Revision: 1
+Type: python 3.10
+# Free to modify, update, and take over
+Maintainer: Hanspeter Niederstrasser <nieder@users.sourceforge.net>
+Depends: <<
+	%N-shlibs (= %v-%r),
+	blt-shlibs,
+	bzip2-shlibs,
+	expat1-shlibs (>= 2.2.8-1),
+	fink (>= 0.34.8),
+	gdbm6-shlibs,
+	libffi6-shlibs,
+	libgettext8-shlibs,
+	liblzma5-shlibs,
+	libncursesw5-shlibs,
+	openssl110-shlibs (>= 1.1.1g-1),
+	readline8-shlibs,
+	sqlite3-shlibs  (>= 3.30.1-1),
+	tcltk (>= 8.4.1-1),
+	x11
+<<
+BuildDepends: <<
+	blt-dev (>= 2.4z-15),
+	bzip2-dev,
+	expat1 (>= 2.2.8-1),
+	fink (>= 0.32), 
+	gdbm6,
+	gettext-bin,
+	gettext-tools,
+	libffi6,
+	libgettext8-dev,
+	liblzma5,
+	libncursesw5,
+	openssl110-dev (>= 1.1.1g-1),
+	pkgconfig,
+	readline8,
+	sqlite3-dev (>= 3.30.1-1),
+	tcltk-dev (>= 8.4.1-1),
+	x11-dev
+<<
+# See https://github.com/fink/fink-distributions/issues/193
+BuildConflicts: uuid
+# Python 3.4+ wants to install these packages but that conflicts with fink's.
+Recommends: <<
+	pip-py%type_pkg[python],
+	setuptools-tng-py%type_pkg[python]
+<<
+Provides: <<
+	argparse-py%type_pkg[python],
+	enum34-py%type_pkg[python],
+	futures-py%type_pkg[python],
+	pathlib-py%type_pkg[python]
+<<
+
+Source: https://www.python.org/ftp/python/%v/Python-%v.tar.xz
+Source-Checksum: SHA256(a7f1265b6e1a5de1ec5c3ec7019ab53413469934758311e9d240c46e5ae6e177)
+
+PatchFile: %n.patch
+PatchFile-MD5: accc53203fdb4c50b45aea0d9d6ecf5e
+
+PatchScript: <<
+	#!/bin/sh -ex
+	sed 's|@PREFIX@|%p|g' < %{PatchFile} | patch -p1
+	# /tmp and /var/tmp are symlinks to /private/tmp and /private/var/tmp
+	# on OS X. That's fine except it can cause tests in some packages to
+	# fail since '/tmp' != '/private/tmp'. Default to using /private/tmp.
+	perl -pi -e "s|'/tmp', '/var/tmp'|'/private/tmp', '/private/var/tmp'|" Lib/tempfile.py
+	# reason for removing SystemStubs?
+	perl -pi -e 's/ -lSystemStubs//' ./configure
+<<
+
+ConfigureParams: <<
+	--enable-shared \
+	--enable-loadable-sqlite-extensions \
+	--enable-optimizations \
+	--with-dbmliborder=gdbm \
+	--with-system-expat \
+	--with-system-ffi \
+	--with-tcltk-includes="-I%p/include -I/opt/X11/include" \
+	--with-tcltk-libs="-L%p/lib -ltcl -ltk" \
+	--with-dtrace \
+	--without-ensurepip \
+	--enable-ipv6
+<<
+
+CompileScript: <<
+#!/bin/sh -ex
+	SDK_PATH="$(xcrun --sdk macosx --show-sdk-path)"
+	# for OS X 10.14, ensure full SDK is available, but include it after any fink specific versions.
+	#if [ "$(uname -r | cut -d. -f1)" -ge 18 ]; then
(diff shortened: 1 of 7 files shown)


dmacks:
On 10.13...

Python39 [coming shortly...github doesn't have a 'save as draft']

Python310: did not detect support for:
* ossaudiodev
* spwd
which is the correct number of missing, and built even though detected, as configured:
* _abc
* pwd
* time
The first set of self-tests report SUCCESS, but there were some "failed (env changed)" messages along the way. The second set of self-tests had hard failure:
```
0:08:10 load avg: 7.30 [383/420/1] test_pkgutil failed (1 error)
test test_pkgutil failed -- Traceback (most recent call last):
  File "/sw/build.build/python310-3.10.1-1/Python-3.10.1/Lib/test/test_pkgutil.py", line 263, in test_name_resolution
    mod = importlib.import_module(uw)
  File "/sw/build.build/python310-3.10.1-1/Python-3.10.1/Lib/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1050, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1027, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1004, in _find_and_load_unlocked
ModuleNotFoundError: No module named 'é'
```

nieder:
test_pkgutil test failure : https://bugs.python.org/issue41154
It seems to be an HFS+ failure which is why I didn't see it on APFS. We could just add it to the exclusion list for 3.9 and 3.10 since upstream has deprecated the relevant module.

nieder:
This is my module detection list on 12.1, so your results match:
```
The necessary bits to build these optional modules were not found:
ossaudiodev           spwd                                     
To find the necessary bits, look in setup.py in detect_modules() for the module's name.
(Fink package build should have 2 missing)


The following modules found by detect_modules() in setup.py, have been
built by the Makefile instead, as configured by the Setup files:
_abc                  pwd                   time               
```
And I too see the `failed (env changed)` diagnostics, but I don't think they matter

nieder:
FWIW, python is moving to annual release cycles: https://www.python.org/dev/peps/pep-0602/
Each new release will still be security supported for 5 years.
And python36 reached EOL yesterday.

dmacks:
> test_pkgutil test failure : https://bugs.python.org/issue41154 It seems to be an HFS+ failure which is why I didn't see it on APFS. We could just add it to the exclusion list for 3.9 and 3.10 since upstream has deprecated the relevant module.

Disabling that test gets both python39 and python310 fully built. There is a small .deb difference if I build with vs without tests: with tests, I get a bunch of __pycache__ subdirs in various pythonXX/Tools subdirs. These subdirs are empty, but they do not exist at all if tests=off.

nieder:
Python38 has the same thing with empty `Tools/.../__pycache__` folders. Can use something like `find %i -name *__pycache__* -empty | xargs rmdir` (I've never been good with the -exec or -delete options to find).

10.9-libcxx/stable/main/finkinfo/languages/python310.info
```diff
@@ -173,6 +173,10 @@ InstallScript: <<
 	# install some docs and other useful tidbits
 	rm -rf Misc/RPM
 	/bin/cp -R Misc Tools %i/lib/python%type_raw[python]
+
+	# remove leftovers from testing
+	# these dirs are empty and only created if tests are run
+	find %i/lib/python%type_raw[python]/Tools -name *__pycache__* -empty | xargs rmdir
```
dmacks, line 180: Safer to:
```
find %i/lib/python%type_raw[python]/Tools -name *__pycache__* -empty -type d -delete
```

dmacks:
Safer to
```
find %i/lib/python%type_raw[python]/Tools -name *__pycache__* -empty -type d -delete
```

nieder:
Ugh. For some reason, running `/usr/bin/find ./Tools -name __pycache__ -empty -type d -delete` manually in %b works, but fails in TestScript. Even just `find ...` w/out the -delete doesn't find anything. Definitely in the right directory, etc.

nieder:
The `-empty` flag makes it fail when run in TestScript. And `find ./Tools -name __pycache__ -type d -delete` fails. Makes me think they might not be empty (as far as find is concerned).

sth0:
hidden dot files such as .DS_Store?  Usually find works just fine for me.  It really seems that it is not empty as find seems to not trigger.  Also if there is a symbolic link that has the destination missing it will not show up with some ls flags.  (I think the -L)

TheSin-:
instead of -delete use -exec rm -rf {} \;

sth0:
actually use -delete -exec rm -rf {} \;
The backslash is important and there must be the space after the closing curly bracket.
(and take out the spurious "use" as a typo.

nieder:
Directories are very much empty according to `ls` (and various flags to find hidden files).
```
+ /usr/bin/find ./Tools -name __pycache__ -type d -exec rmdir '{}' ';'
rmdir: ./Tools/c-analyzer/cpython/__pycache__: Directory not empty
rmdir: ./Tools/c-analyzer/c_analyzer/__pycache__: Directory not empty
rmdir: ./Tools/c-analyzer/c_analyzer/parser/__pycache__: Directory not empty
rmdir: ./Tools/c-analyzer/c_analyzer/common/__pycache__: Directory not empty
rmdir: ./Tools/c-analyzer/c_analyzer/variables/__pycache__: Directory not empty
```
This eventually worked:
```
/usr/bin/find %b/Tools -name __pycache__ -type d -depth -exec rm -rf "{}" \;
```
`-depth` needed to keep find from trying to go into the directory after deleting it.

dmacks:
LGTM. I took a look at users of pythonXX and none seem to sort by "major.minor", so there is no priority problem of 3.9 being higher than 3.10. Older pythonXX could benefit from the improved find...rm... identified here, but that's not a reason to wait here.

nieder:
OK. We will also want to restrict py34-py36 to 10.9-10.15.

nieder:
Validation failures with python3.10 (3.9 is OK). Happens on multiple py310 packages.
```
Validating .deb dir /sw/build.build/root-termcolor-py310-1.1.0-1...
Error: File in a language-versioned package does not have a pathname specific to that version.
	Offending file: /sw/lib/python3.10/site-packages/termcolor.py
	Offending file: /sw/lib/python3.10/site-packages/termcolor-1.1.0-py3.10.egg-info
```

dhomeier:
> ```perl
> 			# numbers are all "small" (one-digit)
> 			$langver_re =~ /^(\d)(\d)(\d)$/;
> 			# -pyXY is pythonX.Y
> ```

Yes, I needed something like
```perl
			# numbers are all "small" (one-digit) up to Python 3.10
			$langver_re =~ /^(\d)(1??)(\d)$/;
			# -pyXY is pythonX.Y, -pyX1Y is pythonX.1Y
```
(if we want to support up to Python 3.19 for now ;-)
Otherwise builds and works on 10.14.6 and 10.13.6, thanks!

dmacks:
> Validation failures with python3.10 (3.9 is OK). Happens on multiple py310 packages.
> 
> ```
> Validating .deb dir /sw/build.build/root-termcolor-py310-1.1.0-1...
> Error: File in a language-versioned package does not have a pathname specific to that version.
> 	Offending file: /sw/lib/python3.10/site-packages/termcolor.py
> 	Offending file: /sw/lib/python3.10/site-packages/termcolor-1.1.0-py3.10.egg-info
> ```

Fixed in https://github.com/fink/fink/commit/f7ca63721d485c49d36043e399e99391baf7b121

dmacks:
Heh...my commit landed at the same time @dhomeier was commenting. In terms of code, this is just a string-parsing regexp like scan() to decipher major vs minor, not any sort of validation of the value itself. So it seemed clearer to generalize it as "one digit followed by one or two digits" rather than stricter "...and if three digits, middle must be '1'". If we ever get as high as python10.0, we can figure something else out:)

dhomeier:
Just a suggestion in case we wanted to take the "small numbers" still half-seriously – otherwise would have gone with `(\d??)` as well, but `(\d{1,2})` is rather more elegant.
I have not heard even from a tentative timeline for Python 4.0, so this should last for a while. ;-)

nieder:
Putting it here since it's probably a similar code failure as the validation check:
This is in TestDepends:
`(%type_pkg[python] >= 39)` doesn't evaluate to true with foo-py310. But `(%type_pkg[python] >= 310)` works as expected. Confused because 310 > 39, so I would expect the first to work.

dmacks:
Those conditionals are string-compare not number-compare.

PR #864, opened 2021-12-31
libaprutil.0: explicitly link to libapr and libaprutil to resolve und…

nieder:
…efined symbols in the database drivers

On macOS12, building aprutil.0 and it's various database drivers fails with missing symbols from libapr.dylib and libaprutil.dylib.
The system uses a partly homebuilt autoconf setup, so need to patch directly.

The current solution is to add `-lapr -laprutil` to the LDADD flags for the various database drivers. For the database drivers included in the base package (commoncrypto, dbm_db, dbm_gdbm), I used the .la file from the in source build.

I don't know how this worked in earlier macOS releases. Perhaps there was some level of implicit dynamic lookup (there's no such explicit flag in the code that I can see).

Do these packages build on 10.14 or earlier after the patch?

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libaprutil.0-mysql.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libaprutil.0-mysql.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: libaprutil.0-shlibs.patch -*-
 Package: libaprutil.0-mysql
 Version: 1.6.1
-Revision: 4
+Revision: 5
 
 Depends: <<
 	libaprutil.0-shlibs (>= %v-4),
@@ -12,6 +12,7 @@ Depends: <<
 BuildDepends: <<
 	db60-aes,
 	expat1,
+	fink (>= 0.30.0),
 	fink-package-precedence,
 	gdbm6,
 	libapr.0-dev (>= 1.7.0-1),
@@ -30,6 +31,8 @@ Source-MD5: 8ff5dc36fa39a2a3db1df196d3ed6086
 
 PatchFile: libaprutil.0-shlibs.patch
 PatchFile-MD5: eb4274f2998a4146164237056265fa6e
+PatchFile2: libaprutil.0-shlibs-macos12.patch
+PatchFile2-MD5: 5ee7a42deddbb9b8c1cacbd8ba1f3979
 PatchScript: <<
   ### Fix layout
   perl -pi -e 's,/usr/local,%p,g' config.layout
--- a/10.9-libcxx/stable/main/finkinfo/libs/libaprutil.0-odbc.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libaprutil.0-odbc.info
@@ -1,7 +1,7 @@
 # -*- coding: ascii; tab-width: 4; x-counterpart: libaprutil.0-shlibs.patch -*-
 Package: libaprutil.0-odbc
 Version: 1.6.1
-Revision: 4
+Revision: 5
 
 Depends: <<
 	libaprutil.0-shlibs (>= %v-4),
@@ -12,6 +12,7 @@ Depends: <<
 BuildDepends: <<
 	db60-aes,
 	expat1,
+	fink (>= 0.30.0),
 	fink-package-precedence,
 	gdbm6,
 	libapr.0-dev (>= 1.7.0-1),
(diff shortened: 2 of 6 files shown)


dmacks:
apr has its own libtool via libapr.0-dev's share/apr/build/libtool, which defaults to hiding its actual commands because of `LTFLAGS = --slient` in share/apr/build/apr_rules.mk. Hacking that out on my 10.13, I see that it's passing `-Wl,-undefined -Wl,dynamic_lookup` to the linker. I don't know why it's deciding to do that, but it's possibly okay (since the *.so are runtime-loaded into a binary that presumably has those linked; perlmods and some pythonmods are the same way IIRC), but it's also possibly a symptom of a libtool that doesn't know modern OS X.

nieder:
Interesting. I don't see anywhere where it's pulling in -Wl,dynamic_lookup for you. The flag is not in any other files in share/apr/build.

This is what I get on 12.1 when I remove the `--silent` flag and try to build the current stuff we have:
>/bin/sh /opt/sw/share/apr/build/libtool  --mode=compile gcc -g -O2   -DHAVE_CONFIG_H  -DDARWIN -DSIGPROCMASK_SETS_THREAD_MASK   -I/opt/sw/build.build/libaprutil.0-shlibs-1.6.1-4/apr-util-1.6.1/include -I/opt/sw/build.build/libaprutil.0-shlibs-1.6.1-4/apr-util-1.6.1/include/private  -I/opt/sw/include/apr  -I/opt/sw/include  -o crypto/apr_crypto_commoncrypto.lo -c crypto/apr_crypto_commoncrypto.c && touch crypto/apr_crypto_commoncrypto.lo
libtool: compile:  gcc -g -O2 -DHAVE_CONFIG_H -DDARWIN -DSIGPROCMASK_SETS_THREAD_MASK -I/opt/sw/build.build/libaprutil.0-shlibs-1.6.1-4/apr-util-1.6.1/include -I/opt/sw/build.build/libaprutil.0-shlibs-1.6.1-4/apr-util-1.6.1/include/private -I/opt/sw/include/apr -I/opt/sw/include -c crypto/apr_crypto_commoncrypto.c  -fno-common -DPIC -o crypto/.libs/apr_crypto_commoncrypto.o
/bin/sh /opt/sw/share/apr/build/libtool  --mode=link gcc  -g -O2      -L/opt/sw/lib -release 1 -module -rpath /opt/sw/lib/apr-util-1 -o crypto/apr_crypto_commoncrypto.la crypto/apr_crypto_commoncrypto.lo 
libtool: link: gcc  -o crypto/.libs/apr_crypto_commoncrypto-1.so -bundle  crypto/.libs/apr_crypto_commoncrypto.o   -L/opt/sw/lib  -g -O2  
  "_apr_crypto_clear", referenced from:
      _crypto_block_encrypt_init in apr_crypto_commoncrypto.o
      _crypto_block_encrypt in apr_crypto_commoncrypto.o
      _crypto_block_decrypt in apr_crypto_commoncrypto.o
<many more symbols>

So it's grabbing it from some where else (or yours is different than mine?)
%p/share/aclocal/libtool.m4 has it, but I don't think apr is rebuilding its copy.

dmacks:
share/apr/build/libtool is from apr.0-dev, which has old-libtool bugs. Compare that file (or the libtool.m4 in its build-dir) with share/aclocal/libtool.m4 around line 1079. Maybe the usual "BigSur and beyond" patch against apr.0-dev's ./configure will solve it?

nieder:
I see now. %p/share/apr/build/libtool is built from apr-1.7.0/libtool.m4, and that has faulty macOS12 detection. If I autoreconf libapr, then dynamic_lookup is available in share/apr/build/libtool, and then current aprutil builds with it.

dmacks:
https://answers.launchpad.net/ubuntu/+source/apr-util/+question/143914 is a long-ago question from a different platform that found the same sort of linking is needed.

TheSin-:
is there any objection, cause if not I'd like to update the libtool.m4 in apr, unless someone has a better idea

TheSin-:
So I modified libapr.0-shlibs.info, added a BD on libtool2 and autoconf2.6 then on the first line of PatchScript I ran ./buildconf and rebuilt libapr, then tried to build libaprutil.  Same error, and I checked after buildconf and the libtool.m4 was the one from libtool2, the issue is the libtool2's libtool.m4 had a header

`Copyright (C) 1996-2001, 2003-2015 Free Software Foundation`

so I don't think it's recent enough.  Thoughts?

TheSin-:
I just committed a fix to libapr which fixes this issue.

PR #870, opened 2022-01-17
Texlive 2021 try2

nieder:
This is a 'fork' of #798.
That PR got difficult to merge when texlive-2017 was merged in #843.

The changes here are all consistent with those from #798, just slightly reorganized and squished when necessary to make a nicer history.

Important changes from 2017:
1) bump TL to 20210325
2) bump kpathsea and ptexinc to the relevant internal versions
3) remove external xpdf/poppler support since it's now a mess (and upstream is abandoning it)
4) Add fixes for getting tlmgr and sysconfig working.


nieder:
Builds as expected with external libs patch. But I don't know texlive enough to test functionality.

dhomeier (dismissed):
Builds for me as well, with `texconfig` working under `sudo`. This seems all based on calling `tlmgr`, and directly attempting to run this confirms that behaviour
```sh
> tlmgr --usermode paper a4                                                                                                                       
tlmgr: action `paper' not supported in usermode
tlmgr: An error has occurred. See above messages. Exiting.
```
so apparently a user-level configuration of this setting is no longer foreseen.

Running `latex` and `dvipdf` on some older papers worked without any obvious problems.

Perhaps it might be worth mentioning this migration from `texconfig` to `tlmgr` in `DescDetail` or `DescUsage` (e.g. that making available usermode typically requires a `tlmgr init-usertree` first).

dmacks:
I'm going to push this, as the only unreviewed detail is a DescUsage that we can tweak later if necessary. Let's get this functionally moving along...

PR #876, opened 2022-01-29
Build fftw3 with gcc11

dhomeier:
Built and tested on 10.13 and 10.14.

This also enables the building of `sse2`, `avx` and `avx2` instructions, which compiles the corresponding sources, to be used where available. Note that I had also `--enable-avx512` added, which did build and run on Mojave with Xcode 11.3 as well (on an i5 with AVX, but neither AVX2 nor AVX512), but failed on compilation on High Sierra with Xcode 10, trying to use some avx512 assembly or machine code; same succeeded with `--enable-avx2` (on a Core 2, with no AVX of any kind).
Might be a bug in gcc11 that it is not correctly detecting the available instructions, but I left it at these flags for now – might perhaps check for `xcode` version >= 11 to put `--enable-avx512` back in, but maybe the other options should also be checked for even older systems.

ETA by @nieder:
List of fftw3 packages:
- [x] fftw3
- [x] gmic (@nieder) [fftw3] (**needs %r bump**)
- [x] gracegtk (@akhansen) [fftw3]
- [x] libmeep17-shlibs [fftw3]
- [x] libmeep24-shlibs [fftw3]
- [x] libmpb1-shlibs (Aurelien Chanudet) [fftw3]
- [ ] octave-3.6.4 (still using gcc5) [fftw3]
- [ ] octave-3.8.2 (still using gcc5) [fftw3]
- [x] relax (Sylvain Barbot) [fftw3] (**needs %r bump**)
- [ ] root5 (@mommsen) [fftw3]
- [x] tinker (@wgscott) [fftw3]
- [x] xcrysden (@nieder) [fftw3]


nieder:
Built successfully on 12.1/intel.

There are 10 packages using currently using gcc9 that directly depend on fftw3. We should add them to this branch so that they can all transition at the same time.

nieder:
Also, %r for fftw3 needs to be bumped if you're changing the available instructions (deb change).

dhomeier:
Sure, thanks for marking the dependent packages. Should they get dependencies on `fftw3 (>= 3.3.10-2)`?

nieder:
Yes. probably should get an explicit %v-%r dep if we're trying to make sure they are working against a gcc11 fftw3 when updating themselves to using gcc11.

dmacks:
I pushed an update to gracegtk to gcc11 as part of my lapackXXX upgrade sweep. Feel free to upgrade it further; hopefully it's not too broken for the moment if it's already at gcc11 but still using an older-gcc fftw3.

dhomeier:
Pushed some packages (ngspice slipped in by accident before I noticed that it does not use any gcc-fsf).
Where applicable I have also moved to hdf5.200.v1.12 or the latest netcdf; all tests passing where existent.

`relax` turned out a special case as it had a dep on `docopt-py27`. AFIACS this is only used in the two provided Python scripts, which also require numpy. Don't know what the odds are of getting a `numpy-py27` to work on macOS 11 or later, so I moved those to `Recommends` that will make the scripts functional. `python3` is just necessary to find an appropriate installation of the Python `Tools/scripts` and let the installed scripts run with  `/usr/bin/env python3` – they could of course always be called as `python%type_raw[python]  %p/bin/obsres.py`.
An alternative could be to decide on one Python version now and make that a `Depends` and hardcode it.

dhomeier:
The `tinker` update may also need some discussion – linking everything also against the system `JavaVM` made the executables unusable without the legacy Java runtime installed, so I installed those versions to a separate bindir.
But they don't seem to have any use outside of the Tinker FFE suite, which has its own standalone macOS installer, so we may as well dump the java versions altogether.

Apart from `root5-cernlib` this collection should be complete then. I am happy to push my hacked .info for that, if anyone wants to give it a try on 10.13 or older...

nieder (approved):
lgtm. we can do octave (#881) and root5 elsewhere

nieder:
Let's close this and deal with root5 and octave elsewhere.

dhomeier:
Thanks. Yes, review of #881 is probably going to take a bit longer.

PR #877, opened 2022-01-30
Update fortran programs to gcc11

kamischi:
Update Fortran source code from gcc9 to gcc11. Actually untested, but I do not expect much problems. In case, bug reports will be acted on.


dmacks:
ortep3 Source-MD5 is not correct. It looks like it's using a non-specific revision from a git repo, so any time a commit is made, it would 1) change the MD5 and 2) change the .deb result. Probably just need to switch to a specific hash instead of master.

dhomeier:
> Probably just need to switch to a specific hash instead of master.

There haven't been any commits in 6 years, so it's unlikely to make a difference, but using
https://github.com/ornl-ndav/ortep/blob/76959cab2d/ortep.f
should avoid any surprises.

dmacks:
Thanks for the git-blob, @dhomeier.

Everything looks good *except* ifeffit. There are some "implicit declaration" messages (not related to gcc11, but while we're working on these packages...). They are only warnings on my 10.13, but would be errors on some newer OSX. Probably just need to find the right system .h to #include.

dmacks:
It looks like ifeffit is building its own readline, a *very* old version of it, which is leading to the warnings. Debian has hacked the build system to use an external readline package. We might want to use [their configuration_patches](https://sources.debian.org/patches/ifeffit/2:1.2.11d-11/configuration_patches/).

dmacks (approved):
The main objective here looks well satisfied. Other improvements or long-standing bugs found along the way need not block it.

PR #878, opened 2022-01-31
Unmaintain python stuff

kamischi:
I am not really working with these packages any more. They were part of a project, I abandoned a while ago.


dmacks (approved):
Thanks for all your work on them!

Issue #888, opened 2022-02-12
openssl latest updates breaks postgresql on macOS 10.12


wodev:
with the latest openssl updates 1.1.1.l/1.1.1m or openssl 3.0,x a start of postgesql is no longer possible on macOS 10.12. 
Reason for this is that openssl switched to the Apple Crypto Random API which is not correctly implemented on macOS 10.12 and prior.

Disabling the Apple Crypto Random API in OpenSSL with parameter OPENSSL_NO_APPLE_CRYPTO_RANDOM solves the issue. on macOS 10.12.

see also the issue from the PostgresApp: https://github.com/PostgresApp/PostgresApp/issues/643

dmacks:
#889 closed by f49c59e8481c5ff6f958ba783d6a6c1c32cab31e

PR #889, opened 2022-02-12
OpenSSL -  disable apple  crypto random api on macOS 10.12 and lower 

wodev:
a PullRequest to fix issue #888 which causes start issue of PostgreSQL on macOS 10.12 with OpenSSL 1.1.1l/1.1.1m or OpenSSL 3.0.


dmacks:
According to #888 (and upstream) the problem is that it's implemented (detected and called) but broken. As this gets committed, should we change the comments in the .info that instead say it's "not available"?

wodev:
I changed the comments in the .info files to the correct problem description.

dmacks (approved):

PR #890, opened 2022-02-14
Scipy update to 1.8 (for Python 3.8+) and new BuildDeps

dhomeier:
From  1.7 onwards Scipy is using the `pythran` package for build (in theory this can be overridden my setting `SCIPY_USE_PYTHRAN=0`, but likely at some performance cost, and I've been unable to get the tests to work that way), so pulling  a couple extra BuildDeps in. Legacy versions moved to the `scipy-py-1.2.1` package.


nieder (changes requested):
* gast, beniget, and pythran need a `Distribution` field for py36 (and lower).
* pythran has a py35 variant, but has Deps on beniget-py35, which starts at py36.

PR #883, opened 2022-02-07
Numpy update with OpenBLAS and hypothesis test depends

dhomeier:
Running the hypothesis tests with tox installed failed on 
> 'tox config file (either pyproject.toml, tox.ini, setup.cfg) not found'

In addition the PyPI source does not contain any tests that could be run, and tox would likely try to pip-install its own dependencies even if got to run, apart from the circular dependency.
I've therefore switched to the github source and running pytest directly. This still involves a circular dependency with numpy, the presence of which is now checked instead.

Numpy can in fact now be built against Accelerate again (see notes), but OpenBLAS should still be preferable; also the latter is detected and selected automatically, if present. So I made that a [Build]Depends and the preferred `NPY_BLAS` – otherwise a deterministic build against Accelerate would require deinstallation of `openblas` during the build.
The tests are also running on pytest; the nose-based `runtest.py` has not been recommended in numpy for some time, and appears to be omitting a number of tests compared to pytest. Pytest also prints clearer results. Tested on 10.14 and in mostly the same configuration on 10.13.

PR #891, opened 2022-02-14
PyTables 3.7.0 update and build against hdf5.200.v1.12

dhomeier:
Updating to the latest upstream version and moving to the `hdf5.200.v1.12` library set. This version is dropping support for older version; officially 3.6 is still supported, but building against `hdf5.200.v1.12` or `hdf5.200.v1.10` fails on an `UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2` on `H5public.h`.
If the py27-py36 types are still needed, they are probably best moved out to a 3.5.1 version.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/pytables-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/pytables-py.info
@@ -1,81 +1,55 @@
 Info2: <<
 
 Package: pytables-py%type_pkg[python]
-Version: 3.5.1
+Version: 3.7.0
 Revision: 1
-Distribution: <<
-	(%type_pkg[python] = 34 ) 10.9,
-	(%type_pkg[python] = 34 ) 10.10,
-	(%type_pkg[python] = 34 ) 10.11,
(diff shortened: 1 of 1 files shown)


dmacks:
pytables-py is a reverse-dependency of a few other packages, so we do need to keep py27 and py34-36. So I assume offloaded to a new old-versions .info as is becoming annoyingly standard for -pyXX packages.

dhomeier:
Yes, with 3.6 past its end-of-life and 3.7/8 in the security release phase, cannot expect any recent package releases to guarantee support for py36 or below.

dmacks:
The -py36 and -py37 variants have a dependency on numpy-py >= 1.19 but those variants of numpy-py are only at v1.14.5.

dhomeier:
> Everything builds fine now (10.13). Can the old-version remnant (3.5.1) move up to newer hdf?

No idea. But py36 and py37 are supported up to numpy 1.19.x and 1.21.6 (just came  out), so I have added those as as LTS to #883 ; updating them all here to 3.7.0 seems easier getting stuck with 3.5.1.

Issue #887, opened 2022-02-12
Update HDF* packages


nieder:
This is to centralize updating the hdf packages.
1) remove any versions that are no longer in use (or at least make them -shlibs only stubs)
2) transition them to latest dependencies (such as gcc9 -> gcc11) See #874 

- [ ] hdf (static only; dx and grads are only dependees)
- [ ] hdf5.9-gfortran (unused)
- [ ] hdf5.9 (gdal, octave-3.6, and octave-3.8 are only dependees) #881 for octave-3.8
- [ ] hdf5.9-oldapi-gfortran (unused)
- [ ] hdf5.9-oldapi (unused)
- [ ] hdf5.10-gfortran (unused)
- [ ] hdf5.10 (used by netcdf-c7/11, pytables-py, and libmatio2)
- [ ] hdf5.10-oldapi-gfortran (unused)
- [ ] hdf5.10-oldapi (unused)
- [ ] hdf5.100-gfortran
- [ ] hdf5.100
- [ ] hdf5.200-gfortran
- [ ] hdf5.200

dhomeier:
> @dhomeier pytables-py is yours. There is a new upstreram (3.7.0, vs our 3.5.1). Along the way, support for py27/34/35 were dropped at various points.

#891 is moving that version to hdf5.200.v1.12.

dmacks:
"hdf" is v4, which seems quite different than hdf5 in terms of what features they supply.

dmacks:
Given hdf is different enough thing that hdf5* isn't a replacement, we'll keep it and try to push to gcc11 so we don't have to block its reverse-dependencies. If we change If we can push it to gcc11, we should propagate a versioned build-dependency and rev-up on its reverse-dependencies because it's a static lib:

- [ ] dx
- [ ] h5utils
- [ ] grads

nieder:
grads and h5utils Bdep on both hdf and hdf5. I can't see if that's for backwards compatible access to old data formats or what.

dmacks:
> grads and h5utils Bdep on both hdf and hdf5. I can't see if that's for backwards compatible access to old data formats or what.

All I know is that they detect it, they link against it, and according to `nm -m` they actually use symbols from it.

sth0:
If I install hdf5.200.v1.10-shlibs.  I can't access the 1.10 API since the package has returns:
#error "Can't choose old API versions when deprecated APIs are disabled"

[...]

Issue #874, opened 2022-01-28
gcc9 migration tracker


dmacks:
We're up to gcc11 now and gcc9 is reported to have build problems on OSX 11 (see fink-beginners [Unable to update gcc9](http://permalink.gmane.org/gmane.os.apple.fink.beginners/30395)) and maybe even back to 10.15 (#844 comment on August 15)

[ ] apbs
[ ] arpack-ng
[ ] atlas
[ ] cd-hit (@nieder)
[ ] cdo
[ ] cernlib2006 (@mommsen)
[ ] clipper (@wgscott)
[ ] ddscat
[ ] eccodes
[ ] fasttree (@nieder)
[ ] fftw
[ ] fftw3
[ ] floppy (@kamischi)
[ ] flow (@kamischi)
[ ] freehelix (@kamischi)
[ ] ftidy (@kamischi)
[ ] geant4.9 (@mommsen)
[ ] gmic (@nieder)
[ ] gracegtk (@akhansen)
[ ] grib-api-fortran
[ ] hdf
[ ] hdf5.10-gfortran
[ ] hdf5.10-oldapi-gfortran
[ ] hdf5.100-gfortran.v
[ ] hdf5.200-gfortran.v
[ ] hdf5.9-gfortran
[ ] hdf5.9-oldapi-gfortran
[ ] ifeffit (@kamischi)
[ ] lapack350-shlibs
[ ] lapack360-shlibs
[ ] lapack361-shlibs
[ ] lapack371-shlibs
[ ] lapack380-shlibs
[ ] libccp4 (@wgscott)
[ ] libcfitsio8-shlibs
[ ] libctl7-shlibs
[ ] libharminv3-shlibs
[ ] libmeep17-shlibs
[ ] libmeep24-shlibs
[ ] libmpb1-shlibs (Aurelien Chanudet)
[ ] libnc-dap3 (@akhansen)
[ ] mathgl6 (Jack Fink)
[ ] mopac7 (Jack Fink)
[ ] netcdf-fortran7
[ ] netcdf-fortran8
[ ] openblas (@dhomeier)
[ ] openmpi
[ ] ortep3 (Jack Fink)
[ ] patchy4 (@mommsen)
[ ] pgplot
[ ] phaser-mpi (@wgscott)
[ ] qrupdate (@akhansen)
[ ] r-base36
[ ] r-base36 (@babayoshihiko)
[ ] refmac (@wgscott)
[ ] relax (Sylvain Barbot)
[...]

PR #893, opened 2022-02-24
Netcdf updates

nieder:
Update netcdf packages to use hdf5.200 (v1.12.0) and gcc11 as possible.
This will help with #887 and #874


dmacks:
Generally looks good by eye. While grepping for uses of the older packages, I noticed that a few netcdf-c* descriptions mention "netcdf-fortran5".

nieder:
Only kept the notes referencing the fortran7 and fortran8 packages that match the active netcdf-c# packages.

dmacks (approved):
lgtm (glad we hadn't accidentally removed a -fortran package we needed!

PR #894, opened 2022-03-01
qhull8.0: revive non-reentrant lib, unified in existing -shlibs/-dev

dmacks:
Upstream still supports libqhull even though it now also has libqhull_r, and the API changes mean it's not always trivial to migrate existing users of the older interface. Also flip libqhullcpp to shared (both changes per Debian patches).

This allows eventual killing of qhull6.3.1.


dmacks:
Debian puts the three shared libs in separate packages but a unified -dev. But we've already committted the simple package-name to the _r package, so the naming could get confusing if we follow them. They each have their own headers set, so we could some day retain just some of them if upstream abandons them entirely, but the install_names are in sync and upstream seems to move slowly and keep them in sync.

nieder (approved):

PR #895, opened 2022-03-01
xtide: bump to version 2.15.4 with harmonics 20220109

PovlAbrahamsen:
A very simple update: a handful of bug fixes in the main code, and updated harmonics files.

Tested on MacOS 10.15.7, where it builds and appears to work correctly. Maintainer is @akhansen


dmacks (changes requested):
lgtm, except that xtide needs `GCC: 4.0` (pre-existing nit I just noticed).

PovlAbrahamsen:
While I'd normally be happy to make this change, I'm currently at sea (for
several weeks), and can't access the internet through my own laptop. So if
you could add this line, it would be really helpful!

Thanks,

Povl

dmacks:
Done. Safe travels!

PR #897, opened 2022-03-07
CRAN packages for R 4.1 and 4.0 (20220304)

babayoshihiko:
This is a first work to bring CRAN packages for R 4.1 and 4.0.

The packages here are built both on 4.1 and 4.0. For those that require R (>= 4.0), CRAN .info files are created for older Rs.

The revision numbers stay the same unless:
* the CRAN package is a new version (revision = 1)
* a missing dependency (libgettext8) are added to the CRAN (revision++)
* the license is fixed (revision++)

cran-pspline-r.info:
The homepage says its license is 'Unlimited'. The License field of the .info is GPL, which is wrong. Should be OSI-Approved or Public Domain?


10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-tree-r.info
```diff
@@ -1,7 +1,7 @@
 Info2: <<
 
 Package: cran-tree-r%type_pkg[rversion]
-Type: rversion (3.6)
+Type: rversion (4.1 4.0 3.6)
```
nieder, line 4: Needs `Distribution:` field.

nieder:
Looks fine except for that cran-tree-r.missing the Distribution field.

babayoshihiko:
> Looks fine except for that cran-tree-r.missing the Distribution field.

Done.

nieder (approved):

Issue #903, opened 2022-04-01
Scrapping java components


dmacks:
Java's not free for all users but is a low-level dependency of huge portions of our distro via the 'bison' package and the 'libidn-java' component of the 'libidn' package, not just packages actually for java tools. This is a tracker item for removing them from the main packages if possible:
- [ ] bison
- [ ] libidn (libidn-java is a splitoff, can we push it into a separate .info?)
- [ ] libgraphviz

(list not yet complete...please update as you find them)

akhansen:
IIRC libiconv builds differently whether it finds Java or not.  This is why we make people install Java as part of the bootstrap.

nieder:
dpkg1.16 eliminates the java dep in base. Don't recall how exactly.

sth0:
Isn’t OpenJDK GPL v2 licensed?  I use that to satisfy the Java dependencies in an OS manner.

dmacks:
I have no idea about alternatives, just what happened when I tried to install fink (dpkg1.16) from scratch on a brand new computer. I could bootstrap and selfupdate, but not really install anything major due to explicit deps on system-java. If there's a free alt, it would be great to have it packaged, and either hard-code it (for consistency, which would solve a lot of older path chaos?) or as an alternative (if we just need "some javac").

nieder:
Sample openjdk17.info. License is GPL2+.
It requires a change to Validation.pm to allow %p/Library/Java. Each major JDK release will mean a new package, but there won't be a moving target.
`fink list system-java` detects the provided files, and `/usr/libexec/java_home -V` also detects it.

[openjdk17.info.txt](https://github.com/fink/fink-distributions/files/8405601/openjdk17.info.txt)

cooljeanius:
>    * [ ]  libgraphviz (java classes/bindings are compiled and made part of the -shlibs package)

For searchability purposes, this also affects other graphviz splitoffs, i.e., the one that's just "graphviz", and libgraphviz238-shlibs

PR #904, opened 2022-04-02
Offload libidn-java to own .info

dmacks:
Pursuant to Issue #903, this helps remove java deps from mainline dep tree


nieder:
Untested but looks fine and a good idea. Assuming the standalone libidn-java deb contents are otherwise identical to the SplitOff contents?

dmacks:
I don't know squat about java, but the package's contents are the same in terms of which filenames are present in each.

PR #910, opened 2022-04-17
Deprecate texi2html

nieder:
Remove texi2html from packages that currently BDep on it. It's long been dead and supplanted by texinfo for quite a while.
Fontforge (doesn't use texi2html) had to be updated because it wasn't building on 10.14.5 and it was needed for lilypond.


dmacks (approved):
Sounds like a great idea to me!

PR #913, opened 2022-05-08
New libhunspell

dmacks:
This is a new libversion, and upgrades of some packages that were using older versions, all listing @nieder as maintainer. Will work on the other unmaintained ones later, didn't want to step on an active one's toes.


nieder (approved):
Untested but LGTM

Issue #917, opened 2022-05-16
uncertainties-py3X FTBS


nieder:
Couple issues:
1: TestScript uses `%p/bin/nosetests-%type_raw[python]`, but that only exists for nose-py27. The py3X only have `%p/bin/nosetests3.X` (py27 also has the file name without the hyphen).

2. On py35, tests fail at the very start
```
  File "/sw/lib/python3.5/site-packages/nose/__init__.py", line 1, in <module>
    from nose.core import collector, main, run, run_exit, runmodule
  File "/sw/lib/python3.5/site-packages/nose/core.py", line 153
    print "%s version %s" % (os.path.basename(sys.argv[0]), __version__)
                        ^
SyntaxError: invalid syntax
```

3. on py36, build fails at the very beginning
```
/sw/bin/python3.6 setup.py build
error in uncertainties setup command: use_2to3 is invalid.
```

dmacks:
Our package is a quite old version (https://pypi.org/project/uncertainties). Current 3.1.6 is reported to be compatible with py27 and py3x up through 3.9.

dmacks:
The errors reported at /sw/lib/python3.5/site-packages/nose is a problem with our nose-py3x packages: they can give different .deb depending if that package's TestScript runs or not. Oooops. Fixed in its 1.3.7-4

dhomeier:
> Our package is a quite old version (https://pypi.org/project/uncertainties). Current 3.1.6 is reported to be compatible with py27 and py3x up through 3.9.

The 3.0 update got somewhere stuck with #204 (and still had the wrong nosetests calls); created a new PR for 3.1.6.

PR #919, opened 2022-06-11
Uncertainties 3.1.6

dhomeier:
Update to current upstream, fixing the `nosetests` call for #917.
This version requires `future-py`, which comes in with the missing pyversions as well; also trying to get its tests to run again. As far as I could check, the enabled tests are running without network access now.
Tested all variants on 10.14.5, py39/310 on 12.4 and 10.13.6 as well.


nieder (approved):
LGTM on 10.14

PR #920, opened 2022-06-11
Python 3.9.13 + 3.10.5 updates with build against openssl300

dhomeier:
New upstream versions for the two current Python versions. I gave this a try and bumped the OpenSSL version from 1.1.1 to 3.0.1; built and tested successfully on 10.13, 10.14 and 12.4 (arm64).
Also added a patch to various scripts installed under `Tools` to make them directly executable with the right Python interpreter.


dmacks:
Can python310 switch from pkgconfig to ppkgconfig? Please double-check the build transcript and flag ordering. The reason is that pkgconfig->glib2, and new glib2 will use meson->python310, so python310->pkgconfig is a dep-cycle.

nieder:
@dhomeier ping

dhomeier:
Superseded by #996

PR #921, opened 2022-06-18
Htodd fdupes

htodd:
Adding new package for fdupes.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/editors/vim.info
+++ b/10.9-libcxx/stable/main/finkinfo/editors/vim.info
@@ -1,13 +1,13 @@
 Info2: <<
 Package: vim%type_pkg[-nox]
 Type: -nox (boolean)
-Version: 8.2.5010
+Version: 8.2.5126
 Revision: 1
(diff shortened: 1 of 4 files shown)


nieder:
What's the purpose of the `Replaces: %{ni}` and `Provides: fdupes` fields?
Also, don't need to `BuildDepends:` on `fink (>= 0.24.12-1)` any more. All supported distributions have only had fink greater than that version.

htodd:
My mistake. I used another info file as an example and I never know what to keep. I'll fix that.

nieder:
Forgot to notice: There's a BDep on fink-package-precedence, but it's not used in CompileScript.

nieder:
I would add a check with f-p-p to CompileScript (or to the very beginning of InstallScript so that you don't have to add to the default CompileScript). Important since this uses ncurses and we want to make sure to not accidentally grab system-ncurses.

nieder (approved):
fdupes looks good. not sure how bowtie2 and vim made it into the branch, but shouldn't affect merging.

Issue #918, opened 2022-05-26
Migrating off pykg-config


dmacks:
'pkg-config' is written in C and has dependencies including glib2 (which wants pkg-config or equivalent tool as a BuildDepends). For places where we don't want such a heavy toolchain (and to avoid a dep-cycle in glib2, something upstream is somewhat working on solving), we have 'pykg-gconfig' that is written in python (currently hardcoded to use /usr/bin/python) as an alternative. But now Apple is dropping that python interp. And python has a heavy dependency chain and and also has pkg-config or equivalent tool as a BuildDepends, so that's a different dep-cycle. So now we also have 'ppkg-config', written in perl and no deps except perl-core itself. So that should be a portable solution since we are guaranteed to have a perl interp at some known location.

I propose deprecating pykg-config, since it is currently little-used and has the most intractable dep-cycle, and switching to ppkg-config for places we don't want to use the regular pkg-config. Current pykg-config users:
 - [ ] arpack-ng
 - [x] glib2
 - [x] gphoto2
 - [x] libgphoto2.6
 - [x] libxml2
 - [ ] octave-3.8.2
 - [x] octave-3.6.4 (killed)
 - [x] pkgconfig
 - [ ] qrupdate
 - [x] xmkmf

nieder:
Sounds reasonable. Make it so.

dhomeier:
FWIW 1.3.0 seems to work with system python3 just out of the box, just needs a check for the executable in /usr/bin.
But if the perl alternative is easier to drop in, that's probably more maintainable. Otherwise I can open a PR to keep those other packages supported until they are migrated to ppkg-config.

dmacks:
> FWIW 1.3.0 seems to work with system python3 just out of the box, just needs a check for the executable in /usr/bin. But if the perl alternative is easier to drop in, that's probably more maintainable.

If it were just that we needed to find the right executable-name on different platforms, that would be doable. But recent OSX dropped /usr/bin/python* altogether.

dhomeier:
> If it were just that we needed to find the right executable-name on different platforms, that would be doable. But recent OSX dropped /usr/bin/python* altogether.

Uh, which ones are those? I tested this on Monterey 12.3 both for x86_64 (Rosetta) and arm64, and basically
```bash
        [ -x /usr/bin/python ] || pyversion=3
        /usr/bin/python$pyversion setup.py build_py --with-pc-path="$pc_path" build_scripts
```
[...]

PR #923, opened 2022-07-02
Add pkgconf package as an alternative to pkgconfig

dhomeier:
As pointed out in https://github.com/fink/fink-distributions/issues/918#issuecomment-1163887385 some of the circular dependency problems caused by Freedesktop's `pkgconfig` are not resolved by switching to `ppkg-config`, since e.g. meson builds do not accept the latter as a pkg-config replacement.
This is an initial attempt to package the suggested alternative https://github.com/pkgconf/pkgconf.
Could not run tests due to a missing dependency, and shlibs layout may need discussion.
`pkgconf` was tested to build and work in building `python310` and the meson build of scipy 1.9.0rc1 on 12.4/arm64 and 10.14.6.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/pkgconf.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/pkgconf.info
@@ -1,0 +1,79 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info2: <<
+Package: pkgconf
+Version: 1.8.0
+Revision: 1
+Source: https://distfiles.dereferenced.org/%n/%n-%v.tar.xz
+Source-Checksum: SHA256(ef9c7e61822b7cb8356e6e9e1dca58d9556f3200d78acab35e4347e9d4c2bbaf)
+Depends: %N-shlibs (= %v-%r)
+BuildDepends: fink (>= 0.32), pkgconfig-common
+BuildDependsOnly: true
+
+ConfigureParams: --mandir=%p/share/man
+
+# tests require `kyua`
+#InfoTest: TestScript: make check || exit 2
+
+InstallScript: make install DESTDIR=%d
+
+SplitOff: <<
+  Description: Pkgconf - Shared library
+  Package: %N-shlibs
+  Files: lib/libpkgconf.3.dylib
+  Shlibs: %p/lib/libpkgconf.3.dylib 4.0.0 %n (>= 1.8.0-1)
+  DocFiles: AUTHORS COPYING NEWS README.md
+<<
+
+DocFiles: AUTHORS COPYING NEWS README.md doc
+
+Description: Drop-in pkg-config replacement w/o [B]Deps
+DescDetail: <<
+pkgconf is a program which helps to configure compiler and linker flags for
+development libraries. It is similar to pkg-config from freedesktop.org.
+
+libpkgconf is a library which provides access to most of pkgconf's
+functionality, to allow other tooling such as compilers and IDEs to discover
+and use libraries configured by pkgconf.
+<<
+DescUsage: <<
+You can run 'pkgconf' directly (it's in fink's default PATH), or,
+for most build systems set PKG_CONFIG=%p/bin/pkgconf in the build
+environment to have it use this C implementation instead of the
+Freedesktop pkg-config program.
+
+pkgconf uses a strict "first match" search scheme to find .pc
+files. The default search-path list is hardcoded into the pkgconf
+binary at compile-time using data from the 'pkgconfig-common' package.
+See the pkgconf manpage for env vars that can alter the default.
+A dependency on 'pkgconf' (at the version where things work for you)
+is the only BuildDepends needed in packages that use others' .pc
+files.
+
+As of 0.22-4, lists of -I and -L flags are re-sorted to place all
+fink paths before any system (/usr or X11) paths; other paths come
+first, so PKG_CONFIG_PATH should allow users to over-ride fink, but fink
+always over-rides system-supplied paths). This operation is controlled
+by a wrapper script supplied by the 'pkgconfig-common' package.
+
+## using `pkgconf` with autotools
+
+Implementations of pkg-config, such as pkgconf, are typically used with the
+PKG_CHECK_MODULES autoconf macro.  As far as we know, pkgconf is
+compatible with all known variations of this macro. pkgconf detects at
+runtime whether or not it was started as 'pkg-config', and if so, attempts
+to set program options such that its behaviour is similar.
+
+In terms of the autoconf macro, it is possible to specify the PKG_CONFIG
+environment variable, so that you can test pkgconf without overwriting your
+pkg-config binary.  Some other build systems may also respect the PKG_CONFIG
+environment variable.
+
+To set the environment variable on the bourne shell and clones (i.e. bash), you
+can run:
+
+    $ export PKG_CONFIG=/usr/bin/pkgconf
+<<
+Homepage: https://github.com/%n/%n
+License: OSI-Approved
+Maintainer: None <fink-devel@lists.sourceforge.net>
+<<


nieder:
* pkgconf and pkgconfig both install `%p/share/aclocal/pkg.m4`.
* There are also some packages that Dep (not just BDep) on pkg-config (I think because they wrap foo-config around pkg-config). Do we want to allow that with pkgconf?

dhomeier:
I just added the `BuildDependsOnly` since the validator complained about dylibs and headers in the same package. Could of course set it to `false`, not sure what is preferable from the point of packaging policy.
I had @eli-schwartz's remark about meson in mind, which might use pkgconfig or a substitute at runtime, but actually they are rather BDeps of the packages using meson for the respective build.

I guess one could also create a -dev package as second splitoff including `pkg.m4` etc., which then might also install a symlink to `%p/bin/pkg-config` as suggested in the README.md, to make it a full Replaces/Conflicts of `pkgconfig`.

The `pkg.m4` here is slightly older (_# serial 11 (pkg-config-0.29.1)_) than the _serial 12_ from `pkgconfig-0.29.2-2`.

eli-schwartz:
The serial number difference is mostly irrelevant. pkg-config bumped it in this commit: https://gitlab.freedesktop.org/pkg-config/pkg-config/-/commit/677e9248753d31c4b7ceed7805ccdc3fc508e980

> There aren't any significant changes here, but this will cause aclocal to get the latest version.

In fact, other than whitespace (and post-serial-bump comment typo fixes) there's exactly one actual change, to use the lookup name rather than the saved variable name for logging. It doesn't affect usage though.

I guess pkgconf simply hasn't synced a handful of changes + the serial number.

dhomeier:
I've removed `pkg.m4` here for now; perhaps another option is to move it out off `pkgconfig` as well and install the latest version with `pkgconfig-common` – or would that confuse autoconf, expecting to find a 'pkg-config' even if none is installed?

dmacks:
We currently have each pkg-config-like program at its own name, so callers know exactly which implementation they are getting.

Debian is fairly conservative in terms of avoiding breakage during transitions. [Today, they made a formal proposal that pkgconf replace pkgconfig.](https://lists.debian.org/debian-devel/2022/07/msg00252.html) As part of that, they've done a full-distro rebuild and didn't see substantial problems. With that, I think we can do that switch also, having the pkgconf install ```bin/pkg-config``` as well, which means a clean Conflicts/Replaces:pkgconfig. But that would be blocked by any installed package that has Depends:pkgconfig. It feels like the clean solution is to have pkgconf:Provides:pkgconfig. However, we've had breakage at the debian level with the triplet Conflicts/Replaces/Provides:foo when foo is itself also a real package.

So we could have a final actual pkgconfig package that is a dummy on top of pkgconf ([fink's usual "obsolete" mechanism](https://wiki.finkproject.org/index.php/Fink:Packaging:Obsolete_Packages) for when a package is drop-in replaced by another work-alike with a different package-name). That would mean pkgconf:Replaces:pkgconfig (not also Conflicts).

eli-schwartz:
pkgconf 1.9.2 is out.

BTW it includes this change: https://github.com/pkgconf/pkgconf/commit/59a56dfa64e8bb7ea430daeffac18af18647e5b6
Which re-syncs the pkg.m4 with the freedesktop pkg-config.

dhomeier:
I've updated to 1.9.3 and after some struggles managed to build `glib2-shlibs` using that; but it seems it really needs to be installed as a full replacement of `pkgconfig` for that (including `pkg.m4`).
What does not work in packages tested so far is xft2 (failing validation on building against XQuartz fontconfig etc.) and guile22 (does not compile, possibly getting wrong `--std=` flags).
Both are building with our old pkgconfig, so that could be a show-stopper – have to try yet if ppkg-config could serve as a replacement there.
@dmacks when making this just a `Replaces` but not a `Conflicts` to pkgconfig, Fink is still happy installing both in parallel – would the latter need to be made a dummy/deprecated package at the same time?

eli-schwartz:
I wonder what the diff is between the generated buildsystem Makefiles / ninja files when using pkg-config vs. pkgconf?

dhomeier:
The `pkgconf` build of `guile22` is failing with
```
  CC       guile-guile.o
[gcc -std=gnu11 -DHAVE_CONFIG_H   -DBUILDING_LIBGUILE=1 -I../.. -I.. -I../../lib -I../lib -I/opt/sw3/include -I/opt/sw3/src/fink.build/guile22-2.2.7-2/guile-2.2.7 -I/opt/sw3/include -I/opt/sw3/include -D_THREAD_SAFE  -Wall -Wmissing-prototypes -Wdeclaration-after-statement -Wpointer-arith -Wswitch-enum -fno-strict-aliasing -fwrapv -fvisibility=hidden -I/opt/sw3/include -g -O2 -MT guile-guile.o -MD -MP -MF .deps/guile-guile.Tpo -c -o guile-guile.o ../../libguile/guile.c]
In file included from ../../libguile/guile.c:34:
In file included from ../../libguile.h:81:
../../libguile/pairs.h:182:1: error: static declaration of 'GC_is_heap_ptr' follows non-static declaration
GC_is_heap_ptr (void *ptr)
^
/opt/sw3/include/gc/gc.h:523:20: note: previous declaration is here
GC_API int GC_CALL GC_is_heap_ptr(const void *);
                   ^
1 error generated.
make[3]: *** [guile-guile.o] Error 1
```
and its differences in `build/Makefile` vs. the `pkgconfig` one are missing Fink paths in
```
< BDW_GC_CFLAGS = -I/opt/sw3/include
< BDW_GC_LIBS = -L/opt/sw3/lib -lgc
---
> BDW_GC_CFLAGS = 
> BDW_GC_LIBS = -lgc 
397c397
< CFLAGS = -I/opt/sw3/include -g -O2
---
> CFLAGS =  -g -O2
```
Although I could not spot any difference in the generated libguile headers, manually executing `gcc` with the same flags fails in the `pkgconf` build dir and succeeds in the `pkgconfig`-generated one; so something about dealing with the system `gc/gc.h` is not detected correctly by `pkgconf`. First I thought the Fink wrapper script (which was directly copied from the `pkgconfig` package description to `pkgconf` was not doing the right mojo with the latter, but even calling the former's binary directly with `PKG_CONFIG = %p/bin/pkg-config.real` gets the paths right.
The package _can_ be built by adding
```
        BDW_GC_CFLAGS=-I%p/include \
        BDW_GC_LIBS="-L%p/lib -lgc" \
        LIBFFI_CFLAGS=-I%p/include \
        LIBFFI_LIBS="-L%p/lib -lffi" \
```
to `ConfigureParams`, but that's probably not very satisfactory when trying to fix the builds of future packages.

eli-schwartz:
pkgconf probably thinks for some reason that that path is part of the "system" includes that the compiler uses even without specifying any flags. Both freedesktop pkg-config and `pkgconf` are supposed to skip printing those unless an option is passed to the pkg-config invocation, but the latter may have been configured differently (or autodetected the configuration differently when building the tool itself).

dhomeier:
Yes, it's obviously not needed in `CFLAGS` since those still come with plenty of `-I%p/include` anyway, but somewhere in the configuration of the gc and libffi subsystems it is missed that `gc/gc.h` will eventually be in the system (Fink) include path. These bits from config.log may be relevant:

pkgconf:
> configure:53755: $PKG_CONFIG --exists --print-errors "$bdw_gc >= 7.2"
configure:53758: $? = 0
configure:53816: result: yes
configure:53829: checking for GC_pthread_exit
configure:53829: gcc -std=gnu11 -o conftest  -g -O2 -I/opt/sw3/include  conftest.c -lgc    >&5
ld: library not found for -lgc

pkg-config:
> configure:53755: $PKG_CONFIG --exists --print-errors "$bdw_gc >= 7.2"
configure:53758: $? = 0
configure:53816: result: yes
configure:53829: checking for GC_pthread_exit
configure:53829: gcc -std=gnu11 -o conftest -I/opt/sw3/include -g -O2 -I/opt/sw3/include  conftest.c -L/opt/sw3/lib -lgc   >&5
  "_GC_pthread_exit", referenced from:
      _main in conftest-28e443.o
ld: symbol(s) not found for architecture x86_64

So in the first case pkgconf tries to compile without the Fink library path, probably concluding that `gc` is not available at all and then including the conflicting header definitions, while pkg-config finds the library at `%p/lib/libgc.1.dylib`, but determines that it is missing a number of pthread symbols.

Indeed the pkgconf build can also be fixed with just `SetLDFLAGS: -L%p/lib` instead of setting the BDW_GC_ and LIBFFI_ vars (so at leasts that's easier to figure out than the latter), but it's strange that the linker path is incomplete at all within a Fink build. Perhaps this can be fixed after all in the `pkgconf` wrapper script...

dmacks:
I just hacked on guile22's build system a bit: c8a6882009f4c631a478634007d07bbf6368ebc2 might solve the problem. But if nothing else, it will give a more verbose output to help diagnose it.

dhomeier:
> Edit: This is probably the same upstream bug whose same symptom was reported years ago: https://debbugs.gnu.org/cgi/bugreport.cgi?bug=35427

Or at least its incomplete fix. `pairs.h` now in fact has the `#ifndef HAVE_GC_IS_HEAP_PTR` mentioned there, but since the macro is not defined in `gc/gc.h`, pkgconfig needs to add it to `build/config.h`
> #define HAVE_GC_IS_HEAP_PTR 1

The pkgconf build doesn't, obviously incorrectly determining from not finding the `gc` library that the header is also not present. Without that gcc even fails with the include paths sorted correctly, probably because `libguile.h` includes `gc/gc.h` ahead of `libguile/pairs.h`.

I'd agree that this build is broken, and since we know how to work around it, it does not have to keep us from switching to `pkgconf`.

> The bug typically results from upstream confusion between *_CFLAGS variables returned from pkg-config's autoconf macros and *_CFLAGS automake variables. The pkg-config value is supposed to be passed via *_CPPFLAGS, but here it's loaded into CFLAGS in the configure script.

In this case it's only done for the BDW_GC and LIBFFI versions (by pkg-config):
> pkg_cv_BDW_GC_CFLAGS=${pkg_cv_BDW_GC_CFLAGS=-I/opt/sw3/include}
pkg_cv_BDW_GC_LIBS=${pkg_cv_BDW_GC_LIBS='-L/opt/sw3/lib -lgc'}
pkg_cv_LIBFFI_CFLAGS=${pkg_cv_LIBFFI_CFLAGS=-I/opt/sw3/include}
pkg_cv_LIBFFI_LIBS=${pkg_cv_LIBFFI_LIBS='-L/opt/sw3/lib -lffi'}

Again, `pkgconf` does not add any paths here at all.
The reason probably being that by default it does not print any system paths without the extra flags
`--keep-system-cflags`, `--keep-system-libs`; e.g. `pkgconf --libs bdw-gc` prints only `-lgc`.
Taking `-L%p/lib` as system path would perhaps work with one of Fink's gcc-fsf compilers, but obviously not with clang.
That is actually more or less how it's [documented](https://github.com/pkgconf/pkgconf/#compatibility-with-pkg-config).

Could try to add those extra flags through the wrapper script, but if instead, following the hints in their README, I configure pkgconf with
```
ConfigureParams: <<
  --with-system-libdir=/usr/lib \
  --with-system-includedir=/usr/include:/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include
<<
```
it no longer considers the Fink paths system paths to be stripped, which lets guile22 build without any further modifications, and also fixes the validation failures in xft2-dev!
Not sure if the system-include is sufficiently general for 11.0 and later (on 12.6 the above seems to contain everything that is also in /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/usr/include).

dhomeier:
> I just hacked on guile22's build system a bit: [c8a6882](https://github.com/fink/fink-distributions/commit/c8a6882009f4c631a478634007d07bbf6368ebc2) might solve the problem. But if nothing else, it will give a more verbose output to help diagnose it.

That came in just with my last comment on the way out.
I tested the build with that commit, and the include paths are indeed properly sorted
> gcc -std=gnu11 -DHAVE_CONFIG_H   -DBUILDING_LIBGUILE=1 -I../.. -I.. -I../../lib -I../lib   -I/opt/x64/include -I/opt/x64/include -D_THREAD_SAFE  -Wall -Wmissing-prototypes -Wdeclaration-after-statement -Wpointer-arith -Wswitch-enum -fno-strict-aliasing -fwrapv -fvisibility=hidden -g -O2 -MT guile-guile.o -MD -MP -MF .deps/guile-guile.Tpo -c -o guile-guile.o `test -f 'guile.c' || echo '../../libguile/'`guile.c

but with the original pkgconf that command still fails on the conflicting declarations, probably confirming that it's the order in which they are included within `libguile.h`. It builds with the system paths fixed in pkgconf as described in the last comment.
Note that there is also a new set of debian patches available:
> Source2: http://archive.ubuntu.com/ubuntu/pool/main/g/guile-2.2/guile-2.2_2.2.7+1-4.debian.tar.xz
Source2-Checksum: SHA256(eca2e36c3ce91bd4fcadd7827d1738df19f84f63d560f751c573bc01088d6e6e)

The package builds with them the same way, but I did not notice any other differences.

eli-schwartz:
pkg.m4 seems simple, they're the same file anyway. Just create another package called `pkg-m4`, have it use the pkgconf source tarball as its source, and implement the install routine as a single `mkdir && cp`.

Then have all pkg-config implementations depend on that.

dmacks:
It's time to act on this (sorry it fell off everyone's radar!). We've now found ppkg-config has a serious bug (and is a poor crutch anyway) and debian seems to have had no trouble migrating from pkg-config to pkgconf. The two concerns are handling BDO (related to libpkgconf) and pkg.m4 (colliding with pkg-config's file). Rolling this .info as a regular libary+userprogram suite rather than an all-in-one package solves the former easily if we use 'pkgconf' rather than 'pkgconfig' as the package-name. The latter is solved by Replaces. Debian has a nice approach of a separate 'pkgconfig' package with symlink, so that "installing the latest pkgconfig' migrates users to pkgconf and there is never a time when things are in an incomplete/inconsistent state.

Looking at Debian, they're using 1.8.1 and I see no effort to jump to the 2.x series that has been out for a while now.

dmacks:
Please try the pkgconf in #1254.

PR #582, opened 2020-03-07
First version of meson installation

leighsmith:
First version of meson. This can be used to build GNOME libraries such as the latest versions of pango.

PR #924, opened 2022-07-02
New package 'meson-py'

dhomeier:
Packaging the `meson` build system building on the earlier (closed) attempt from #582 and homebrew recipes.
Built, tested (with a few failures) and used to build scipy 1.9.0rc1 on 12.4/arm64 and 10.14.6.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/meson-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/meson-py.info
@@ -1,0 +1,103 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info2: <<
+Package: meson-py%type_pkg[python]
+Version: 0.62.2
+Revision: 1
+Type: python (3.7 3.8 3.9 3.10)
+Depends: python%type_pkg[python], ninja (>= 1.7-1)
+BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python]
+
+Source: https://files.pythonhosted.org/packages/source/m/meson/meson-%v.tar.gz
+Source-Checksum: SHA256(a7669e4c4110b06b743d57cc5d6432591a6677ef2402139fe4f3d42ac13380b0)
+
+PatchScript: <<
+    #!/bin/sh -ev
+    HDF5INC="$(ls -d %p/opt/hdf5.v1.12/include 2> /dev/null || echo '')"
+    perl -pi -e 's|(/bin/env python)(3)|${1}%type_raw[python]|;' *.py mesonbuild/*.py tools/*.py test*/*/*/*.py
+    perl -pi -e "s|(python)(3)(')|\${1}%type_raw[python]\${3}|;" test\ cases/python/[1345]*/meson.build
+    perl -pi -e "s|(python)(3)(')|\${1}-%type_raw[python]\${3}|;" test\ cases/python/2*/meson.build
+    if [ -n "$HDF5INC" ]; then
+        HDF5LIB=${HDF5INC%%/include}/lib
+        perl -pi -e "s|(h5c =)|add_project_arguments('-I${HDF5INC}', language: ['c', 'cpp'])\n\$1|;" "test cases/frameworks/25 hdf5/meson.build"
+        perl -pi -e "s|(h5c =)|add_project_link_arguments('-L${HDF5LIB}', language: ['c', 'cpp'])\n\$1|;" "test cases/frameworks/25 hdf5/meson.build"
+    fi
+<<
+
+CompileScript: <<
+    %p/bin/python%type_raw[python] setup.py build
+<<
+
+InstallScript: <<
+    %p/bin/python%type_raw[python] setup.py install --root %d
+    mv %i/bin/meson %i/bin/meson-py%type_pkg[python] 
+    mv %i/share/man/man1/meson.1 %i/share/man/man1/meson-py%type_pkg[python].1 
+    perl -pi -e 's|(/usr)(/bin/python)(3)|%p${2}%type_raw[python]|;' %i/share/polkit-1/actions/com.mesonbuild.install.policy
+    perl -pi -e 's|(/usr)(/bin/meson)|%p${2}-py%type_pkg[python]|;' %i/share/polkit-1/actions/com.mesonbuild.install.policy
+    mv %i/share/polkit-1/actions/com.mesonbuild.install.policy %i/share/polkit-1/actions/com.mesonbuild.install.policy-py%type_pkg[python]
+<<
+
+PostInstScript: <<
+    update-alternatives --install %p/bin/meson meson %p/bin/meson-py%type_pkg[python] %type_pkg[python] \
+          --slave %p/share/man/man1/meson.1 meson_man %p/share/man/man1/meson-py%type_pkg[python].1 \
+          --slave %p/share/polkit-1/actions/com.mesonbuild.install.policy mesonbuild %p/share/polkit-1/actions/com.mesonbuild.install.policy-py%type_pkg[python]
+<<
+
+PreRmScript: <<
+    if [ $1 != "upgrade" ]; then
+        update-alternatives --verbose --remove meson %p/bin/meson-py%type_pkg[python]
+    fi
+<<
+
+UseMaxBuildJobs: true
+InfoTest: <<
+    TestDepends: <<
+        cmake,
+        hdf5.200.v1.12,
+        hdf5-cpp.200.v1.12,
+        hdf5.200-gfortran.v1.12,
+        ninja (>= 1.10.0)
+    <<
+    TestScript: PATH=/usr/bin:$PATH %p/bin/python%type_raw[python] -B run_tests.py || exit 1
+    TestSuiteSize: medium
+<<
+
+DocFiles: COPYING PKG-INFO README.md contributing.md
+Description: Fast and user friendly build system
+DescDetail: <<
+Meson is an open source build system meant to be both extremely fast, and,
+even more importantly, as user friendly as possible.
+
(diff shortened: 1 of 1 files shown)


10.9-libcxx/stable/main/finkinfo/devel/meson-py.info
```diff
@@ -0,0 +1,103 @@
...
+    %p/bin/python%type_raw[python] setup.py build
+<<
+
+InstallScript: <<
+    %p/bin/python%type_raw[python] setup.py install --root %d
+    mv %i/bin/meson %i/bin/meson-py%type_pkg[python] 
+    mv %i/share/man/man1/meson.1 %i/share/man/man1/meson-py%type_pkg[python].1 
+    perl -pi -e 's|(/usr)(/bin/python)(3)|%p${2}%type_raw[python]|;' %i/share/polkit-1/actions/com.mesonbuild.install.policy
+    perl -pi -e 's|(/usr)(/bin/meson)|%p${2}-py%type_pkg[python]|;' %i/share/polkit-1/actions/com.mesonbuild.install.policy
+    mv %i/share/polkit-1/actions/com.mesonbuild.install.policy %i/share/polkit-1/actions/com.mesonbuild.install.policy-py%type_pkg[python]
```
eli-schwartz, line 36: Do you actually need to install a polkit file at all? AFAICT fink doesn't have polkit available anyway.

dhomeier: Good point, this was left from the initial version; the only other polkit file present there on my system was installed by `dpkg`.

dmacks:
I'd been working on meson recently at [dmacks:meson2](https://github.com/dmacks/fink-distributions/commits/meson2) ("2" because it was my second attempt, not a package-name). It has a few additional patches and documentation of some test-fails. Feel free to adopt this thing, I just really want to see it done and finaly had some free time to work on it.

After investigating the meson project goals and features, I decided to merge it back into a single main package rather than having -pyXX variants. Regardless though, there probably needs to be a main "meson" package that is either the thing itself or a ```Type:bundle``` on top of the variants pool (like https://github.com/dmacks/fink-distributions/commit/d31ad1a09810ea0cae62250fc7590401c92f8532. Otherwise every maintainer has to make a variant decision for a single command-line tool and users might wind up having to install 4 pythons just to build a gnome package.

eli-schwartz:
This decision is a lot easier for distros and vendors that have a default python which all command-line utilities that *happen* to be written in python, target by default. I'm not familiar with how fink works though.

I can verify that Meson is supposed to work with any (non-EOL) version of python, and that it simply treats python as:
- an implementation language for writing a CLI utility
- a fallback for `find_program('python3')`, if it cannot find one in `$PATH` and none is specified via a `--native-file`'s "binaries" config section then it knows as a last-ditch attempt it can find "some version" via sys.executable; this is primarily used to allow stdlib-only helper scripts that work on Windows, which may not have an existing python.

dhomeier:
Thanks, I think I had actually gotten a notification about a commit to the old PR that reminded me this was still not available and triggered this PR. Will have a look at the additional fixes for the tests etc.

I agree it makes sense to have at minimum a `bundle` making a generic `meson` available; since its functionality is rather independent of the interpreter, there may be no need for versioned packages at all (kind of similar to `mercurial` I guess, which could certainly get by with a single Python Type). I have not seen any use cases yet where `mesonbuild` would be imported directly from a Python session (I guess it could come up when e.g. linking it to an IDE).
Just needs a little extra preparation for pythonic packages built with meson then (which I think would automatically fall back to the correct `sys.executable` with the versioned packages), but for the Scipy build all it took was explicitly setting `py_mod.find_installation('%p/bin/python%type_raw[python]')` in `meson.build`.

> This decision is a lot easier for distros and vendors that have a default python which all command-line utilities that _happen_ to be written in python, target by default. I'm not familiar with how fink works though.

Settling to Python 3.10 for packages like this, mercurial or libnewt0 would probably a good solution, though a bit of a moving target. I don't think the general Fink system of maintaining several minor versions in parallel is up for discussion at this point.
And I find it actually convenient to have, say all the currently supported versions of my projects, like 3.8-3.11 for Numpy, Astropy etc., available simultaneously on my system. Though having to keep the older versions back to 3.4 alive becomes a real pain when upgrading packages!

eli-schwartz:
> Just needs a little extra preparation for pythonic packages built with meson then (which I think would automatically fall back to the correct `sys.executable` with the versioned packages), but for the Scipy build all it took was explicitly setting `py_mod.find_installation('%p/bin/python%type_raw[python]')` in `meson.build`.

The SciPy build uses a PEP 517 build backend that e.g. when running `python3.X -m build` to create a wheel, passes the python3.x version as a binary path override, so this should work.

On the other hand, it also wants to have a python metadada dependency on Meson, i.e. depend on a version of Meson packaged for that python version... even though it only runs `meson` as a CLI tool. It also depends on a python metadata slot for `ninja` that is packaged on PyPI. Both of these are because `pip install` cannot invoke a package manager such as apt or pacman or emerge or rpm/dnf, so it decides to just depend on additional internal copies.

eli-schwartz:
In that case, the SciPy "dist-info" metadata will be missing, so if users install SciPy through fink, then try to pip install something that depends on fink, pip will not know that SciPy is already installed.

(dist-info metadata generation is one of the planned improvements for Meson's python extension building support, it's just not there yet.)

That may or may not matter to you. Everything else works with native Meson -- in fact, upstream SciPy's development tooling just uses native Meson to build and install SciPy and run the tests, which works fine.

dhomeier:
Added the `bundle` package now, as there seems sufficient potential demand for the py-versioned packages.

dmacks:
Python-varianted packaging sounds like a good idea. Thanks everyone for explaining it so clearly!

10.9-libcxx/stable/main/finkinfo/devel/meson-py.info
```diff
@@ -0,0 +1,103 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info2: <<
+Package: meson-py%type_pkg[python]
+Version: 0.62.2
+Revision: 1
+Type: python (3.7 3.8 3.9 3.10)
+Depends: python%type_pkg[python], ninja (>= 1.9-1)
+BuildDepends: fink (>= 0.24.12), setuptools-tng-py%type_pkg[python]
```
dmacks, line 8: nit: no need for ```fink (>= 0.24.12)``` BDep

dhomeier: No idea how I came up with 0.24.12. But I thought `Info2` still required 0.20.0; validation passes without any fink [B]Dep... – maybe assuming there are no more ancient versions in existence!

dhomeier:
Upstream update came in between; tested on 10.14.6 and 12.4 (arm64+x64_86).
The remaining test failures
```
FAILED run_unittests.py::AllPlatformTests::test_pkgconfig_gen_escaping - AssertionError: Lists differ: ['-I/usr/with -DLIBFOO', 'spaces/include'] != ['-DLIBF...
FAILED run_unittests.py::NativeFileTests::test_python_module - subprocess.CalledProcessError: Command '['/opt/bin/python3.10', '/private/scratch.noindex/fink....
FAILED run_unittests.py::InternalTests::test_pkgconfig_parse_libs - AssertionError: Lists differ: ['-L/private/tmp/tmpdli2n2ia/2', '-L/privat[34 chars]bar'] ...
FAILED run_unittests.py::FailureTests::test_extraframework_dependency_method - RuntimeError: Configure failed
FAILED run_unittests.py::LinuxlikeTests::test_pkgconfig_uninstalled - subprocess.CalledProcessError: Command '['/opt/bin/ninja', '-w', 'dupbuild=err', '-d', ...
```
seem in part related to the flag-sorting issue noted, others to looking for a Framework installation. I've also seen failures like
```cc
cc -Itest1.p -I. '-I../test cases/common/44 pkgconfig-gen/dependencies' -I/private/scratch.noindex/fink.build/meson-py310-0.63.0-1/meson-0.63.0/tmp4by7txff '-I/private/scratch.noindex/fink.build/meson-py310-0.63.0-1/meson-0.63.0/test -I/private/scratch.noindex/fink.build/meson-py310-0.63.0-1/meson-0.63.0/tmp4by7txff' '-I/private/scratch.noindex/fink.build/meson-py310-0.63.0-1/meson-0.63.0/test -DLIBFOO' -fcolor-diagnostics -Wall -Winvalid-pch -O0 -g 'cases/common/44 pkgconfig-gen' 'cases/common/44 pkgconfig-gen' -MD -MQ test1.p/main.c.o -MF test1.p/main.c.o.d -o test1.p/main.c.o -c '../test cases/common/44 pkgconfig-gen/dependencies/main.c'
clang: error: no such file or directory: 'cases/common/44 pkgconfig-gen'
clang: error: no such file or directory: 'cases/common/44 pkgconfig-gen'
```
indicating the trouble handling path names with spaces.
Could not find any details on the `ninja` failure.

dhomeier:
Tested the new release with 3 failures left in the main (pytest) suite, second (cmake tests) part seems to be missing an `appleframeworks` dependency. Also tested to successfully build scipy-py 1.10.0.

dhomeier:
Updated to upstream 1.6.0, tested on 15.1 arm64 and x86_64 (via Rosetta). Also adding an update for ninja, as tests are checking for version >= 1.12, but everything seems to build and run with 1.11 as well.
Changed the `RuntimeDepends` in `meson.info` as `(= %v-%r)` would lead to a conflict on updating `meson-pyXX`. This way one might accidentally update the py-versioned package only, but at least the unflavoured package may still be updated at any time.
The tests are running on pytest if available and may be running a number of different checks in that case, failures are either 5 or 6 vs. ~500 passing.

PR #926, opened 2022-10-07
Pep517 build system

nieder:
Python is trying to get rid of setuptools with new backends and frontends for building pymods (PEP517). Not exactly sure what all that means. But many packages will start not including setup.py anymore.

This creates a bootstrap module that contains several of the necessary packages for building these new PEP517 packages. Because many depend on each other, they're installed all as one Fink package as pure python (and stashed in %p/share).
Other packages that need them can just add to PYTHONPATH as necessary in order to build.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/bootstrap-modules-py-build_env.patch
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/bootstrap-modules-py-build_env.patch
@@ -1,0 +1,29 @@
+https://github.com/pypa/build/pull/401
+
+--- src/build/env.py.orig	2022-05-23 00:59:47.000000000 +1000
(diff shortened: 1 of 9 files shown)


dhomeier (approved):
Thanks; can't say I fully understand this yet, but everything builds for me (including on Monterey both for x86_64 and arm), and I was able to install the typing_extensions package for #883 with this.
I've run into a number of packages already that no longer come with a setup.py, and typically used some invocation of `pip install -I --no-deps . --root=%d` for those, but this looks much leaner and cleaner.

nieder:
TBH, I copied a lot of the logic from MP.
And it was the needed typing-extensions for numpy that finally got me to try anything. pip works great except when you can't download during the build process. So yeah, this seems like a workable option. Glad it works for someone else.

PR #936, opened 2022-12-27
Enable python3 for ninja bootstrap on macOS 11+

### LINKED ISSUE #932: Ninja looks for python
sudo -u fink-bld [ENV] sh -c /tmp/fink.I9puw
./configure.py --bootstrap
python: error: Failed to locate 'python'.

so on 12+ (maybe 11, I skipped 11 for I forget) this needs to be patched in configure to use python3

dhomeier:
Fixes #932 (tested on 12.1-13.1 – no further py3 conversion needed).

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/ninja.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/ninja.info
@@ -8,7 +8,14 @@ Source-Checksum: SHA256(3c6ba2e66400fe3f1ae83deb4b235faf3137ec20bd5b08c29bfc368d
 SourceRename: %n-%v.tar.gz
 # re2c >= 0.15.3 for its --no-version flag
 BuildDepends: re2c (>= 0.15.3)
-CompileScript: ./configure.py --bootstrap
+CompileScript: <<
+  #!/bin/sh -ex
+  if [ -x /usr/bin/python ]; then
+    ./configure.py --bootstrap
+  else
+    /usr/bin/python3 configure.py --bootstrap
+  fi
+<<
 InstallScript: <<
   #!/bin/sh -ex
   mkdir -p %i/bin


nieder:
This made ninja build on 12.6. I tried ledger and libnss3.0e which both use ninja, and they failed similarly. Is the python path hardcoded in the binary?

Can probably be fixed likewise.

Is it worth doing something like this to avoid having to duplicate the final command?
```
  if [ -x /usr/bin/python ]; then
    PYTHON=/usr/bin/python
  else
    PYTHON=/usr/bin/python3
  fi
  $PYTHON configure.py --bootstrap
```

dhomeier:
> This made ninja build on 12.6. I tried ledger and libnss3.0e which both use ninja, and they failed similarly. Is the python path hardcoded in the binary?

I rather don't think so; in meson builds e.g. the Python interpreter that is configured by `meson setup builddir` is then defined in the configuration files inside `builddir` like `build.ninja`, `compile_commands.json` and retrieved from there when running `ninja -C builddir`. I assume similar is done in those packages, but I cannot try to build them currently since I don't even get to build emacsen-common on Ventura.

nieder (approved):
LGTM for now.
I wasn't able to find any simple proof that ninja hardcodes the python used into the binary for later runtime events. If it turns out to be the case, we can always split ninja into /usr/bin/python and /usr/bin/python3 .info files (or merge them to use a single %p/bin/python3.XX.

Edit: never mind about python not being hardcoded. It is in fact [hardcoded](https://github.com/fink/fink-distributions/issues/932#issuecomment-1706397337).

dmacks:
%p/bin/ninja is a compiled program that also contains a python script (it inlines the whole src/browse.py file). Previously, mine had ```#!/usr/bin/env python```. Rebuilding on 10.13 still gives me that. And there's also another place where the python interp determined at build-time is encoded in that compiled program (over-rideable by a cmake variable).

Easy enough to fix, either platform-variant or hardcode a consistent fink python interp. I need advice which way to go with that though. And if the former, we need documented somewhere which platforms have which /usr/bin/python*.

PR #937, opened 2022-12-28
Openldap 2.6 update

nieder:
Creates a new package for openldap-2.6
* The server and executables are in openldap26, which C/R older openldap24. I have no idea if we can just normalize to an openldap2 package.
* The library now uses a normal versioning scheme, and the new library package is openldap2-shlibs/-dev, which C/R the older openldap24-dev.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/openldap24.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/openldap24.info
@@ -2,7 +2,7 @@
 Info4: <<
 Package: openldap24
 Version: 2.4.59
-Revision: 1
+Revision: 2
 Maintainer: Daniel Johnson <daniel@daniel-johnson.org>
 Source: mirror:custom:openldap-release/openldap-%v.tgz
 CustomMirror: <<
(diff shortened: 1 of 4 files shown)


dhomeier:
Both are using `soelim`, so on Ventura will need a BDep on `groff` or some other workaround (see https://github.com/fink/fink-distributions/pull/939#pullrequestreview-1241072179).
With `groff` installed both build on 13.1.

nieder:
`soelim` seems to be needed to rebuild the man pages, as set in `build/man.mk`. openldap already comes with manpages in the tarball, so we can probably get away with patching it out. testing...

nieder:
Not so simple. Some manpages are generated, others exist. grr

nieder:
Feel free to add groff as a BDep to openldap24 (and to this branch). It'll take some time to figure out if it's possible to knock it out as a requirement.

dhomeier:
Had to add extra headerpad for the `%b/libraries/...` install names to work on arm64 (sufficient for my build path, although using `-headerpad_max_install_names` does not seem to affect install size either).
1 test segfaulting with that in
```
Testing ldapwhoami as Manager...
./scripts/test076-authid-rewrite: line 177: 80078 Segmentation fault: 11  $LDAPSASLWHOAMI -H $URI1 -Y $MECH -U $ID -w $PASSWD
ldapwhoami failed (139)!
>>>>> 00:09:15 Failed   test076-authid-rewrite for mdb after 2 seconds
```
which would need to be moved out of the way so the remaining 11 tests could run (and install names restored).

nieder:
Might be better to use `--headerpad_max_install_names` since who knows what build paths someone might have.

For the test failure, see https://bugs.gentoo.org/848831#c7
I don't think that provides a fix, but it mentions a log file from the tests that might provide more details, maybe?

nieder:
I think I fixed the test failure

PR #939, opened 2023-01-09
libiconv updates

dmacks:
- [x] Fix for new platforms that don't have /usr/bin/groff
- [ ] New upstream version


nieder:
I'm fine with whatever is needed to fix the loss of groff, but I'm at a loss to understand how removing an HTML tag fixes that. Just trying to understand is all.

dmacks:
The groff tool is part of the Makefile recipes for generating man/iconv.1.html from man/iconv.1, but the source tarball has a pregenerated html file. We have a long-standing patch segment that changes iconv.1, so make sees the newer timestamp on that source file and runs groff to regenerate  iconv.1.html. So I changed the .patch to also include the resulting change in .1.html and therefore not need to have groff available. It's possible our PatchScript may need to `touch` the .1.html to make sure it really is "newer than the .1 dependency file"; I have no idea the granularity of current filesystems.

TheSin-:
Confirmed this fixes libiconv build on 13.x (arm and x86_64)

TheSin-:
please add to bootstrap once approved so I can pull this into my PR please

TheSin- (dismissed):
Works on x86_64 and aarch64 13.x without groff

TheSin-:
I did it with pref as essential in bootstrap when I tested.  It's fairly small didn't add much time at all.

dhomeier (dismissed):
Tested on 13.1/arm64 as well.
`openldap24` wants to run `soelim` from the `groff` suite; adding it as a BDep there seems the easier option.

dmacks:
Sorry about the noise, just re-jiggering the commits and fixing some details so I can cherry-pick.

TheSin-:
still looks good, did you want me, want me to retest? and if so did you want me to retest without pref since I didn't do that?

TheSin-:
I think this can be closed now, libiconv is fixed and committed

TheSin-:
OH I see it's left open for an update, nvm I missed that

PR #942, opened 2023-01-12
Distribution: 13.0 versions of various packages

dhomeier:
Mostly perlmods for `system-perl5303` without further changes required; tested on arm64 as well.

The `xmkmf`-distversions from 10.14 to 13.0 are really identical, wondering if we really need separate package files for all.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/help2man-perl.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/help2man-perl.info
@@ -13,7 +13,8 @@ Distribution: <<
   (%type_pkg[systemperl] = 5184) 10.15,
   (%type_pkg[systemperl] = 5282) 11.0,
   (%type_pkg[systemperl] = 5302) 11.3,
-  (%type_pkg[systemperl] = 5303) 12.0
+  (%type_pkg[systemperl] = 5303) 12.0,
+  (%type_pkg[systemperl] = 5303) 13.0
 <<
 Source: mirror:gnu:help2man/help2man-%v.tar.xz
 Source-Checksum: SHA256(9e2e0e213a7e0a36244eed6204d902b6504602a578b6ecd15268b1454deadd36)
@@ -28,6 +29,10 @@ CompileScript: <<
 #!/bin/sh -ex
  perl -pi.bak -e 's,-shared,-bundle,g' Makefile.in
 #Need to build w/same perl as %type_raw[systemperl]
+  m=%m
+  if [ "$m" = "arm" ] ; then
+    m=arm64
+  fi
   if [ "%type_pkg[systemperl]" = "5100" ] && [ "%m" != "powerpc" ] ; then
     PERL="/usr/bin/arch -%m perl%type_raw[systemperl]"
   elif [ "%type_pkg[systemperl]" = "5123" ] && [ -e "/usr/bin/perl5.12" ] ; then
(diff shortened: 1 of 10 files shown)


dhomeier:
Alternatively, I am seeing the `git` update in 955e39e has both 12.0 and 13.0 in the same file, but trying to build that on 13.1 I am getting a
> Validating package file /opt/sw3/fink/dists/stable/main/finkinfo/devel/git-12.0.info...
Warning: Incorrect filename 'git-12.0.info'. Should be one of:
	git.info
	git-2.34.0.info
	git-2.34.0-1.info
Failed: Please correct the above problems and try again!

@TheSin- have you tested that on 12.0 and 13.x?

TheSin-:
Yes I have, that validator error is annoying, it's cause of the file name, so we need to copy the file to git-13.0.info but I feel that is insane so I just add it.  Maybe the validator needs to be fixes so the filename needs to be the lower version on that line, higher version are okay?

10.9-libcxx/stable/main/finkinfo/devel/help2man-perl.info
```diff
@@ -43,9 +48,9 @@ CompileScript: <<
   elif [ "%type_pkg[systemperl]" = "5282" ] && [ -e "/usr/bin/perl5.28" ] ; then
     PERL="/usr/bin/arch -%m perl5.28"
   elif [ "%type_pkg[systemperl]" = "5302" ] && [ -e "/usr/bin/perl5.30" ] ; then
-    PERL="/usr/bin/arch -%m perl5.30"
```
dmacks, line 46: Should this (and again two lines later) be percent instead of dollar-sign?

TheSin- (changes requested):
With my current PR you do not need to set $m, just use %m like it's already there at %m is arm64 now.

10.9-libcxx/stable/main/finkinfo/devel/help2man-perl.info
TheSin-: It did need it before, as %m was arm, but that is corrected to arm64 in my branch

TheSin-:
everything else looks sane other then help2man

dhomeier:
> With my current PR you do not need to set $m, just use %m like it's already there at %m is arm64 now.

Thanks for catching that, yes that was left from the previous bootstrap.

TheSin-:
> > With my current PR you do not need to set $m, just use %m like it's already there at %m is arm64 now.
> 
> Thanks for catching that, yes that was left from the previous bootstrap.

once it's resolved I'll commit this request I can not see any other issues with it.

10.9-libcxx/stable/main/finkinfo/devel/help2man-perl.info
dmacks: Gotcha

dhomeier:
Done

TheSin-:
shoot quick question on the -12.0 packages.  Is perl 5303 on 12.0 or was it's changed during the 12.x updates?  Anyone know?

dhomeier:
Apparently not the kind of info Apple would put into its release notes. 5.30.3 was released a full year before 12.0 was announced, but then it took 11.3 also nearly a year to update to 5.30.2...

TheSin-:
I'm trying to figure this out to see if we need to do like we did with 11.3 and make a separate file for 12.0 and 12.x (when perl updated)

TheSin-:
Based on this info

https://github.com/apple-oss-distributions/perl/tree/038d8df98240eaa1fb33ec359d06dbd63c4d72ac/5.30

at 12.0.1 it was indeed 5303 I wonder how 5302 made it into some files with 12.0, meh I'm convinced this PR is correct

TheSin- (approved):

nieder:
From https://www.finkproject.org/doc/packaging/reference.php (which will need updating for 13.0), 12.0 came with 5.30.3.

TheSin-:
Nice me and @dmacks on IRC were looking for this docu! thanks @nieder

PR #943, opened 2023-01-12
netpbm11-shlibs build for 13.0/arm64

dhomeier:
Checking for the available system python now to pick the appropriate one.
libjpeg8-turbo is by design Intel-only, so for arm64 this is linking the non-turbo libjpeg9 instead. Tested that the build on x86_64 still works and links the turbo lib even with libjpeg9 installed.


10.9-libcxx/stable/main/finkinfo/graphics/netpbm11-shlibs.info
```diff
@@ -1,7 +1,7 @@
 Package: netpbm11-shlibs
 Version: 10.97.06
 Revision: 1
-Distribution: 10.9, 10.10, 10.11, 10.12, 10.13, 10.14, 10.14.5, 10.15, 11.0, 11.3
+Distribution: 10.9, 10.10, 10.11, 10.12, 10.13, 10.14, 10.14.5, 10.15, 11.0, 11.3, 12.0, 13.0
```
nieder, line 4: If we're merging all the netpbm11 .info files by conditionaling `/usr/bin/python{,3}`, then we can get rid of the Distribution field.

10.9-libcxx/stable/main/finkinfo/graphics/netpbm11-shlibs.info
```diff
@@ -11,27 +11,38 @@ BuildDepends: <<
 	libiconv-dev,
 	libjasper.1,
 	libjbig,
-	libjpeg8-turbo,
+	(%m != arm64) libjpeg8-turbo-shlibs,
+	(%m = arm64) libjpeg9-shlibs,
+	libjpeg8-turbo | libjpeg9,
```
nieder, line 16: Why the BDeps on the -shlibs packages? Shouldn't the %m conditionals be on the header packages here rather than using the `|` for dev pkg alternatives?

dhomeier:
So, libjpeg-turbo 2.x supports arm64 as well. Makes things a bit cleaner, but the Shlibs setup may need some discussion yet.

10.9-libcxx/stable/main/finkinfo/graphics/netpbm11-shlibs.info
dhomeier: Yes; typo (or rather leftovers from copy-and-paste)!

dhomeier:
The main change in packaging here is the shlibs naming scheme as `libjpeg.8.2.2.dylib`, `libturbojpeg.0.2.0.dylib`, with an additional level of symlinks added. Strangely the original install names were `@rpath/libjpeg.8.dylib` etc., which I needed to fix anyway since I found no way to pass validation with that. In principle we might also rename the actual library to `libjpeg.8.dylib` and remove that symlink. But it works with the netlib11 build as is...

nieder:
The new install_names for libjpeg-turbo will require a new package. I've almost got it and will commit that, then transition netpbm11 to that.

nieder:
I've updated libjpeg8 to v2 (new package libjpeg8.2-turbo) and pushed firefox8 and netpbm11 to use that.
sane-backends is the only package left using the older libjpeg8-turbo, but it's failing lots of tests, even updating to latest 5.9.3 (net-snmp/net-snmp#520)

nieder:
Also pushed your gtk-doc commit. I think this can be closed.

dhomeier:
> I've updated libjpeg8 to v2 (new package libjpeg8.2-turbo) and pushed firefox8 and netpbm11 to use that.

Thanks; on arm64 netpbm11 is failing two tests
```
pamfix: You cannot specify both -changemaxval and -clip
/opt/sw/src/fink.build/netpbm11-shlibs-10.97.06-2/netpbm-10.97.06/test/pamfix.test: line 62: 41539 Done                    echo -e "P1\n5 5"
     41540 Segmentation fault: 11  | pamfix -truncate -plain > ${test_out}
pamfix: Cannot read a single row from the image.  Use -verbose to find out why
```
```
tifftopnm: writing PBM file
## If this test fails, the cause may be a problem in the TIFF library.
tiff-roundtrip.test: FAILURE
```
(breaking, since it's in the Install phase), but otherwise I don't see anything critical.

PR #945, opened 2023-01-18
Patches/updates for building with python3  on 12.0+

dhomeier:
Various packages tested on 12.6.1 and mostly also on 13.1, for backward compatibility on 10.14.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/qt5-qtwebengine.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/qt5-qtwebengine.info
@@ -59,10 +59,18 @@ PatchScript: <<
 	perl -pi -e's|^linux:||g' tools/qmake/config.tests/srtp/srtp.pro
 	echo "INCLUDEPATH += %p/include" >> tools/qmake/config.tests/snappy/snappy.pro
 	echo "INCLUDEPATH += %p/include" >> tools/qmake/config.tests/srtp/srtp.pro
+	# Patch to use Fink python27 (on macOS 12+)
+	if [ ! -x /usr/bin/python ]; then
+		perl -pi -e 's|(python)( -c)|${1}2.7${2}|;s|(python)( ..[Qn][Tf][Wi])|${1}2.7${2}|' tools/qmake/mkspecs/features/functions.prf
+		perl -pi -e 's|(/usr/bin/env python)|${1}2.7|' tools/*scripts/*
+		perl -pi -e 's|(python)( ..QTW)|${1}2.7${2}|' src/core/gyp_run.pro
+		perl -pi -e 's|(/usr/bin/env python)|${1}2.7|' src/3rdparty/chromium/build/mac/*.py
+		perl -pi -e 's|(MacOSX.)(10)|${1}1[012]|' src/3rdparty/chromium/build/mac/find_sdk.py
+	fi
 <<
 NoSetCPPFLAGS: true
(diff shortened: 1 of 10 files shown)


TheSin-:
you should be breaking this up so maintainers can approve their own.  There are too many here for me to test, I'll cherry pick the ca-certs change later today.

dhomeier:
Not sure how many maintainers have 12.x or 13.1 to test on. [8ea266b](https://github.com/fink/fink-distributions/pull/945/commits/8ea266be32ed007378355622407609533a67b194) has the rather straightforward ones.

dhomeier:
Note that zziplib here uses the upstream update rather than the patch from 9568691.

nieder:
zziplib decreased compatibility_version when they switched from autotools to cmake. This is a no-go because anything linking it will break if we update zziplib.

https://github.com/gdraheim/zziplib/issues/130

dmacks:
itstool doesn't just need a defined python interp at runtime, but it also needs to have libxml2-pyXX to match it. So we need a separate packaging (revision-separation) for each. Is /usr/bin/python3 always the same python3.x on all platforms we support, or does even the minor-version vary?

dhomeier:
> itstool doesn't just need a defined python interp at runtime, but it also needs to have libxml2-pyXX to match it, so we have to use fink's python interp (as the package already does). No objection to bumping it to use python310 instead of python27, but that seems a separate issue from trying to cope with /usr/bin/python version-changes.
> 
> Note that changing a Depends always requires changing %r.

I agree both that %r should have been updated, and it does not really belong with the systempython updates; I had only included it here as I was unable to build python27 on arm64. As @TheSin- has been successful with that and seems close to open a PR, we could also just keep it as is for now (although generally I'd vote for using a still supported Python version).

10.9-libcxx/stable/main/finkinfo/libs/qt5-qtwebengine.info
```diff
@@ -59,10 +59,18 @@ PatchScript: <<
...
+		perl -pi -e 's|(/usr/bin/env python)|${1}2.7|' tools/*scripts/*
+		perl -pi -e 's|(python)( ..QTW)|${1}2.7${2}|' src/core/gyp_run.pro
+		perl -pi -e 's|(/usr/bin/env python)|${1}2.7|' src/3rdparty/chromium/build/mac/*.py
+		perl -pi -e 's|(MacOSX.)(10)|${1}1[012]|' src/3rdparty/chromium/build/mac/find_sdk.py
+	fi
 <<
 NoSetCPPFLAGS: true
 NoSetCFLAGS: true
-NoSetCXXFLAGS: true
+SetCXXFLAGS: -std=gnu++99
```
cooljeanius, line 73: "`gnu++99`" isn't an actual standard; typo for "`gnu++98`"?

cooljeanius:
```suggestion
SetCXXFLAGS: -std=gnu++98
```

dhomeier: Probably confused with `c99/gnu99`

cooljeanius: > Probably confused with `c99/gnu99`

yeah that's what I was guessing

Issue #909, opened 2022-04-13
Update source checksums across the distro


dmacks:
Fink originally used MD5 for sources. Then we added support for SHA1 and SHA256. Many (maybe even most) packages still use MD5, and a bunch SHA1. Should we work on upgrading the checksum fields to SHA256? The current SHA256 checksum feature requires any one of:
- Apple openssl
- Fink openssl
- Fink coreutils
- Fink md5deep
so I think it's safe to assume it will work everywhere without hauling anything else into fink core or requiring adding a `BuildDepends`.

nieder:
1) If `Source-Checksum` and `Source-MD5` are both present, only `Source-Checksum` is checked.
2) `Source-Checksum: MD5(xxx)` is a valid construct, so it could be used as an easily scriptable replacement for `Source-MD5` w/out having to recalculate checksums for tarballs.
3) 4958 info files use MD5, 267 use SHA1, 960 use SHA256.

nieder:
Doing this by hand is going to be nigh impossible and take forever.

dmacks:
Yup, item (2) is easily scriptable but item (1) is mindless but only partially scriptable at best. Assuming SHA256 is state-of-the-art, should we set a goal of at least always using it for new packages and switching to it whenever we upgrade to a new version?

nieder:
Definitely should use SHA256 on any new packages and preferably on updates as well. Scripting to `Source-Checksum: MD5()` would help with that transition.

nieder:
Also, this seems to work on most things that only have `Source-MD5:`
```bash
for i in `ls -1 *.info`; do
	if ! grep -q -Li "^Source-Checksum:" $i ; then
		# only look in files not using Source-Checksum (can have it and Source-MD5, so don't just search for Source-MD5)
		PKG_NAME=$(grep ^Package: $i | grep -v type_ | cut -f 2 -d:)
		PKG_FULL_SOURCE=$(fink dumpinfo -fsource $PKG_NAME | cut -f 2- -d:)
		if [[ "$PKG_FULL_SOURCE" == " mirror:"* ]]; then
			PKG_SOURCE=$(echo $PKG_FULL_SOURCE | rev | cut -f 1 -d: | rev | xargs basename)
		else
			PKG_SOURCE=$(basename $PKG_FULL_SOURCE)
		fi
		if grep -q -li "^SourceRename:" $i; then
[...]

PR #946, opened 2023-01-18
Update asciidoc to changed upstream source and build for python2.7 + python3

dhomeier:
asciidoc has silently replaced its 8.6.10 sources to (mostly) python3 syntax (https://github.com/fink/fink-distributions/issues/909#issuecomment-1381213080), breaking in turn a few statements not supported in 2.7.
This PR is conditionally applying one patch to restore python2.7 support for those, and another to apply the missing py3 updates (system python3 on 12+ unfortunately does not come with 2to3, so including the output as a patch).
The `testasciidoc.py update` command still fails on 2.7 with a number of exceptions which I have not tried to fix, rather skipping it; on 12.6 I could only test the python3 variant without the proper TestDepends (which do not build on arm64 yet).

Alternatively we might dist-restrict the update to 12.0, 13.0 and try to use the older version for python 2 – have all tests been passing for that one?


nieder:
I don't have the original tarball to see how it's building/passing on 10.14.6 and our mirrors don't have it either. Do you have a copy of the original that I can grab?

dhomeier:
This is  the old 8.6.10 (also found an 8.6.9, but that is > 8 years old):
[asciidoc-8.6.10.tar.gz](https://github.com/fink/fink-distributions/files/10504438/asciidoc-8.6.10.tar.gz)

dhomeier:
On my old 10.14.6 installation all 202 tests are reported as PASSED, although I see a bunch if messages like
```
music2png.py: failed command: abc2ly -o "/private/var/tmp/fink.build/asciidoc-8.6.10-1/asciidoc-8.6.10/tests/data/tmpmsFPfQ.ly" "/private/var/tmp/fink.build/asciidoc-8.6.10-1/asciidoc-8.6.10/tests/data/tmpmsFPfQ.abc" 2>/dev/null
music2png.py: failed command: lilypond --png -o "/private/var/tmp/fink.build/asciidoc-8.6.10-1/asciidoc-8.6.10/tests/data/tmp3wv8gW" "/private/var/tmp/fink.build/asciidoc-8.6.10-1/asciidoc-8.6.10/tests/data/tmp3wv8gW.ly" 2>/dev/null
PASSED: html4: tests/data/filters-test-html4.html
```
With the new 8.6.10 source, all those are SKIPPED, possibly because `testasciidoc.py update` is now broken for python2.

On 13.2 strangely, although the missing source-highlight is reported, all 202 tests are now passing as well (not quite there at building the TestDepends "The following 203 additional packages will be installed"...)

dhomeier:
As the 9.1.1 version from upstream builds pretty much out of the box with python3, I might push a dist-restricted (to 11.0-13.0?) asciidoc-9.1.1 update. Backporting that to python2 is becoming rather unfeasible, so we could try to pin the original 8.6.10 source for the Distribution: < 11.0 version.

nieder:
Since the package hardcodes the python version used, we'll either need to split .info files according to /usr/bin/python[2.7|3], or switch to using a Fink python. Benefit of using a Fink python3 is that everyone will get the latest asciidoc.

Pulling in @clepple since it his package.

nieder:
On 10.14.5 (/usr/bin/python2.7) with your original 8.6.10 tarball:
```
TOTAL PASSED:  202
phase test: passed
```

clepple:
To be honest, I haven't had time to do much with Fink recently, so I should probably not be listed as a maintainer anymore. My main interest here was getting the build dependencies set up for the NUT package (and macOS+USB in userspace is getting harder, but that's another story). We can easily work around whatever py2/py3 invocation is needed in the NUT autoconf scripts.

Please split the info file as needed to comply with Fink policy about py2/3 dependencies.

dhomeier:
> Since the package hardcodes the python version used, we'll either need to split .info files according to /usr/bin/python[2.7|3], or switch to using a Fink python. Benefit of using a Fink python3 is that everyone will get the latest asciidoc.

Both work now with their respective system pythons, but upgrading to 10.x will probably require an installation into `%p/lib/pythonNN/site-packages`. Drawback is perhaps that we then have to decide which of the currently supported 3.x Pythons to select, or drag them all along.

dhomeier:
As there have been comments in favour of getting the updated version from upstream where possible, I have now split the package into python3 and python2 versions. The latter needs a copy of the original 8.6.10 source and passes all tests without patches; 9.1.1 also passes all tests on 13.2 (some are skipped in both, even with graphviz and source-highlight installed).
This mainly needs a safe source path for https://github.com/fink/fink-distributions/files/10504438/asciidoc-8.6.10.tar.gz, and confirmation if python3 is indeed available on all 11.0 systems (I think on 10.15 it needed extra action to install via Xcode) – should work even with 3.8, as the Python code is not very special.

nieder:
I've uploaded the original python2 version of asciidoc-8.6.10 to our SourceForge space since upstream already silently changed it once. The following should work in the Source field for the older version only:
```
Source: mirror:sourceforge:fink/%n-%v.tar.gz
```
For confirmation,
SHA1: 53b9c916bb4e29d2a4b850446be070ef81dcd792
SHA256: 9e52f8578d891beaef25730a92a6e723596ddbd07bfe0d2a56486fcf63a0b983

dhomeier:
Thanks, just tested a fresh download, and it got it already from one of the sf mirrors!

nieder (commented):
LGTM. Tested and passed tests on 10.14.6 and 12.6

nieder (approved):

nieder:
py310 is available everywhere, so if we move up to a pymod in the 10.x series, we can use that until someone asks for another variant.

PR #953, opened 2023-01-25
Allow native build for gawk on arm64

### LINKED ISSUE #951: gawk is x86_64 only
in the gawk configure script there is a place that has

```
*darwin*)
    CFLAGS="${CFLAGS} -arch x86_64"
```

So we need an `aarch64-darwin*) before that, or to update gawk so it'll work on Apple Silicon

dhomeier:
Fix #951


nieder (approved):
Untested but LGTM assuming it doesn't affect x86_64 (which I don't expect it too)

TheSin-:
I plan to test this as soon as I'm done with python27

TheSin-:
Build is good, just testing it as a dep and I'll approve and commit this, thanks @dhomeier

TheSin- (approved):

Issue #954, opened 2023-01-25
db53 and db60


TheSin-:
To work on Apple Silicon both version need to add to the `ConfigureParams`

`--with-mutex=Darwin/_spin_lock_try`

db53 needs it to be changed since one is set and it's wrong anyhow

This will require a rev of since it's a change that will affect the deb in how things are compiled.

Tested on 13.2/arm64 and 13.2/x86_64

TheSin-:
@nieder @dmacks can you confirm sanity before I commit it?

dmacks:
db53 currently has ```--with-mutex=x86_64/gcc-assembly```. Should that be replaced with this new flag?

TheSin-:
yes, I do not believe that wasn't the proper flag, though it did work at the time.  I tested on x86_64 with Darwin/_spin_lock_try and it still builds and works, I just can not test on sub 13.x anymore.

dmacks:
On my 10.13, identical build-transcripts when I use that flag. You had noted a weird TestScript detail, which I also see:
```
cd test/micro; sh test_micro || exit 2
Versions db-[3-9].*.*
test_micro: line 59: test: [3-9]: integer expression expected
test_micro: line 61: test: [3-9]: integer expression expected
test_micro: line 63: test: *: integer expression expected
db-[3-9].*.* run begins: Wed Jan 25 18:05:19 EST 2023
test_micro  -d /sw/build.build/db60-aes-6.0.30-2/db-6.0.30/test/micro/RUN.Macks\
-Factor -i /sw/build.build/db60-aes-6.0.30-2/db-6.0.30/test/micro/configs/run.s\
td...
db-[3-9].*.* run ends: Wed Jan 25 18:05:19 EST 2023
phase test: passed
```
That's because of an attempted shell-variable expansion that wound up using the regex for a range instead of a list that contains the range. Given db53 and db60 are the same in this regard and upstream is dead, I'm fairly sure we don't care.

TheSin-:
So want me to commit the changes then?

dhomeier:
Same on 10.14.6.
As the _spin_*lock mutexes are system functions, I now also tend to think the assembly implementation did not offer any performance benefits.

dmacks:
I pushed 3bcb64a9c1cba67e0e809ad5d0d7c3e177872012, there were some other cleanups I had lurking around for them.

PR #957, opened 2023-01-25
arm64 build fixes

dhomeier:
config.guess updates as required, some additional build patches; pushing now what I've come across so far to keep people from duplicating efforts.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/base/apt.info
+++ b/10.9-libcxx/stable/main/finkinfo/base/apt.info
@@ -15,6 +15,7 @@ PatchScript: <<
  perl -pi -e 's|vector.h|vector|g' apt-pkg/pkgsystem.h include/apt-pkg/pkgsystem.h
  perl -pi -e 's|iostream.h|iostream|g' apt-inst/contrib/extracttar.cc apt-inst/deb/dpkgdb.cc apt-inst/extract.cc apt-pkg/algorithms.cc apt-pkg/contrib/error.cc cmdline/apt-cache.cc cmdline/indexcopy.cc test/versiontest.cc 
  perl -pi -e 's|fstream.h|fstream|g' apt-pkg/sourcelist.cc cmdline/apt-get.cc
+ perl -pi -e 's/(arm-\*  \|)( armbe-\*)/$1 arm64-\* \|$2/' buildlib/config.sub
  sh patch_flush
  for i in `grep -rl '#ifdef __GNUG__' .` ; do perl -pi -e 's/#ifdef __GNUG__/#if defined(__GNUG__) && !defined(__APPLE_CC__)/' $i; done
  echo "x86_64\tx86_64" >> %b/buildlib/archtable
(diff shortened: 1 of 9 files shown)


TheSin-:
You put WAY too many packages in one PR which is why they aren't getting accepted, it's too much to review.

both db53 and db60 are too complex, you only need to add one line to db60 and change one line on db53, that's all, they both need to be forced to use `--with-mutex=Darwin/_spin_lock_try`. This works for x86 and arm64 as it's what should be used for Darwin according to the m4 file.

TheSin-:
I do approve the nettle7 fix though, just waiting on @nieder to approve it since it's his package.

dhomeier:
> both db53 and db60 are too complex, you only need to add one line to db60 and change one line on db53, that's all, they both need to be forced to use `--with-mutex=Darwin/_spin_lock_try`. This works for x86 and arm64 as it's what should be used for Darwin according to the m4 file.

Perhaps `x86_64/gcc-assembly` was used for performance reasons? But I don't have anything to benchmark that on, so the  Darwin one appears good enough. Are you planning to submit your own PR?

dmacks:
> > both db53 and db60 are too complex, you only need to add one line to db60 and change one line on db53, that's all, they both need to be forced to use `--with-mutex=Darwin/_spin_lock_try`. This works for x86 and arm64 as it's what should be used for Darwin according to the m4 file.
> 
> Perhaps `x86_64/gcc-assembly` was used for performance reasons? But I don't have anything to benchmark that on, so the Darwin one appears good enough. Are you planning to submit your own PR?

Because our maintainers are thoughtful enough to document their hackery:
```
mutex is hardcoded to x86_64/gcc-assembly to avoid a detection and build
failure with Xcode 12. Actual failure is unclear but seems linked to
implicit declarations now being errors during ./configure.
```

TheSin-:
I had already seen that note, but I still think it was just the first one found, likely by griping for x86_64. I still think Darwin/_spin_lock_try is the correct answer here.

dmacks:
I agree that having a bunch of unrelated files in a single PR can get confusing, since a bunch of us are working on them or also in other PRs, and no need to wait to push it as a single set.

- [ ] apt
- [x] db53-aes (Issue #954, fixed by 3bcb64a9c1cba67e0e809ad5d0d7c3e177872012)
- [x] db60-aes (Issue #954, fixed by 3bcb64a9c1cba67e0e809ad5d0d7c3e177872012)
- [ ] nettle7
- [ ] cairo (needs pkgconf: PR ##923)
- [ ] fftw3
- [ ] openblas
- [ ] libid3tag

dhomeier:
I did not see what was still in the works at the time; to avoid more duplicate work wanted to push at least the draft here so others can see what is already in development. It's probably best to keep this as a tracker now that packages are being updated in their own PRs.

dhomeier:
> Have you tested cario with older dpkg? I've like to commit that one if it's working with dpkg 1.10

You mean `cairo`? Yes, just built and even uncommented the test suite; there are a couple of failures, as noted in the comments, but config+build passes on 13.2 both archs, 10.14.6 with both dpkg 1.16 and 1.10.

TheSin-:
none of the pkg-config changes are required anymore, I fixed pkg-config tonight, selfupdate and pkg-config will start working.

kristibektashi:
Why is `apt` the only unfixed package that doesn't have a separate pull request?

TheSin-:
I just haven’t had time to move it from my exp since dpkg 1.15+ was accepted and it’s low priority since we don’t have a bindist atm

PR #960, opened 2023-01-26
pykg-config build for 12.0/13.0

dhomeier:
Compatibility with python3

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/pykg-config.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/pykg-config.info
@@ -1,6 +1,6 @@
 Package: pykg-config
 Version: 1.3.0
-Revision: 3
+Revision: 4
 Source: https://files.pythonhosted.org/packages/source/p/%n/%n-%v.tar.gz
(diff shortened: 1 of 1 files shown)


dmacks:
The existing package installs %p/bin/pykg-config.real that is an executable starting with ```#!/usr/bin/python```. On platforms where we have /usr/bin/python3 instead, does that give python3 in the #! line? That's a .deb variation, so we'd need separate revisions on the different distros.

dhomeier:
On Python3 systems it's
```
#!/Applications/Xcode.app/Contents/Developer/usr/bin/python3
```
and I actually thought the two would be pretty much exclusive, but it seems Python 3.8 was already made available in Catalina
https://discuss.python.org/t/right-or-best-way-to-install-python-in-macos-10-15-catalina/6204/2
while 2.7 was only finally pulled with the update to 12.3
https://developer.apple.com/documentation/macos-release-notes/macos-12_3-release-notes#Python
(which I assume already came with 3.9.6 installed).
The way the package is first checking for `/usr/bin/python` it would now still use 2.7 as long as it's available; would probably make more sense to use python3 throughout at least everywhere 3.9.6 is installed. So should this be split into two packages, dist-restricted to (12.0, 13.0) and the rest, respectively?

PR #961, opened 2023-01-26
AquaTerm build update to Xcode 14

dhomeier:
The old revision had `MACOSX_DEPLOYMENT_TARGET=10.7`, and 10.5 in the xcodeproj. Set here to the minimum accepted by Xcode 14 on 13.2, which still builds as is with Xcode 11.3.1 on 10.14.6, but obviously will need to dist-restrict this for still older versions – or set the target depending on the build host?

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/aquaterm-1.1.1-5.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/aquaterm-1.1.1-5.info
@@ -1,8 +1,8 @@
 Package: aquaterm
 Version: 1.1.1
-Revision: 6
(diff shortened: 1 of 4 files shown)


dhomeier:
Note that `xcodebuild` tried to install files into `fink-bld`'s nonexistent home dir at `/var/empty/Library/Developer/Xcode/DerivedData`; similar issue has been reported for `osxfuse` on the mailing list.

nieder:
With that MACOSX_DEPLOYMENT_TARGET set, we'll need to make a versioned copy (`aquaterm-1.1.1-5.info`) that' Dist restricted to 10.9-10.12.

nieder (approved):
LGTM. tested on 10.14.6

Issue #952, opened 2023-01-24
nettle7 fix for Apple Silicon


TheSin-:
Just needs `UpdateConfigGuess: true`

That's all.

sth0:
so does nettle8 and libid3tag.  I can't get my GitHub version to sync with the master and submit a PR. But these additions fix all three packages to build on Apple Silicon.

TheSin-:
nettle7 works fine on Apple Silicon, I even built gnutls3 with it just fine.  Not saying we shouldn't update, but it's not required to support Apple Silicon is all.

Trying to fix db53 and db60 for Apple Silicon is way more important IMHO cause of the depth of depends is has and it's all Mutex issues.

sth0:
It did not build for me on Apple Silicon.  Both nettle7 and nettle8 failed until I edited the dot-info files.  They were complaining about an unknown architecture.  What updated config.guess?  Is that autoconf?  Do you have a different version of autoconf than I do?  I can't think of what the difference would be.

Since UpdateConfigGuess: True worked for builiding the libid3tags package when it complained about an architecture problem I tried it with the nettle series.

TheSin-:
you have to be using my PR and make sure it's up to date.  UpdateConfigGuess is a fink info file param, it updates the config.guess and config.sub commands in the build so that it returns aarch64-apple-darwin22.x

That's the only change I made the nettle7 and it works perfectly.

TheSin-:
@nieder are you okay if I commit this?  My M2 is in today and I'd like to be able to install `gnupg2` and `git` right away on it

TheSin-:
@nieder I'm not sure if you had time to look at this, but I'd really like to add it today, as I want to get python27 updated so I can switch to my new laptop which I can't do without gnupg2 and git for commits ;)

nieder:
Sorry about the delay. If I add `UpdateConfigGuess: true` on 10.14.6 using not-dpkg1.16, the build fails:
```
/sw/bin/m4 ./asm.m4 machine.m4 config.m4 aes-encrypt-internal.asm >aes-encrypt-internal.s
aes-decrypt-internal.s:66:2: error: instruction requires: Not 64-bit mode
 pushl %ebx
 ^
aes-decrypt-internal.s:67:2: error: instruction requires: Not 64-bit mode
```
and many more like it.

nieder:
Would running `autoreconf -vfi` work instead of using `UpdateConfigGuess` ?

[...]

PR #962, opened 2023-01-28
arm64/aarch64 build fixes for nettle7 and libid3tag

dhomeier:
Separating the packages addressed in #952 out here.
`libid3tag` works as proposed with updated `config.guess` (tested on old 10.14 installation with fink 0.45 and dpkg 1.10.21 as well). Both are building as `aarch64-apple-darwin...` now.
`nettle8` from ab0978d builds without further changes.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/crypto/nettle7-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/nettle7-shlibs.info
@@ -37,6 +37,7 @@ PatchScript: <<
 <<
 #
 ConfigureParams: <<
+	("%m" = "arm64") --host=aarch64-apple-darwin`uname -r` \
 	--enable-shared \
 	--disable-openssl
 <<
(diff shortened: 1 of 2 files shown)


TheSin-:
I think the nettle7 fix is likely the best option.  I'll have to cherry pick it though cause I'm not going to test libid3tag.  Hopefully I'll get it in today as I need to get python27 in today

sth0 (approved):
LGTM

TheSin-:
Seems, this isn't enough

```
/usr/bin/install -d -m 755 /opt/sw/src/fink.build/root-nettle7-3.5.1-3/opt/sw/share
/bin/mv /opt/sw/src/fink.build/root-nettle7-shlibs-3.5.1-3/opt/sw/share/info /opt/sw/src/fink.build/root-nettle7-3.5.1-3/opt/sw/share/
mv: rename /opt/sw/src/fink.build/root-nettle7-shlibs-3.5.1-3/opt/sw/share/info to /opt/sw/src/fink.build/root-nettle7-3.5.1-3/opt/sw/share/info: No such file or directory
```

seems a missing dep maybe?  I'm looking into it.

```
  Documentation:     no
```

```
checking for makeinfo... not-found
```

need a BD on `texinfo`

Once tested I'll commit nettle7 with this patch and the BD addition

dhomeier:
> need a BD on `texinfo`

Ah, of course; same as for `nettle8` (at least on Ventura, <= 12.x still has its own makeinfo, but better to use the same build everywhere).

nieder:
I think these are fixed. Just updated libid3tag and nettle7 was fixed in 6c2b7df

Issue #956, opened 2023-01-25
python27 on arm64


TheSin-:
`configure: error: Unexpected output of 'arch' on OSX`

Haven't looked into it, but likely just need to update the version or update configure?

TheSin-:
We also want to check so that we don't try to make fat binaries, in the `CompileScript`

move `--with-universal-archs=intel-64`

```
        # for macOS 11, need help finding libz in new system-library cache
        # https://bugs.python.org/issue41116
        # pthread is still not found, so disable test_ctypes for now as well
        if [ "$(uname -r | cut -d. -f1)" -ge 20 ]; then
                DIST_CONFIG_PARAMS="--enable-universalsdk=$(xcrun --sdk macosx --show-sdk-path) --with-universal-archs=intel-64"
        fi
```

to

```
        # for macOS 11, need help finding libz in new system-library cache
        # https://bugs.python.org/issue41116
        # pthread is still not found, so disable test_ctypes for now as well
        if [ "$(uname -r | cut -d. -f1)" -ge 20 ]; then
                DIST_CONFIG_PARAMS="--enable-universalsdk=$(xcrun --sdk macosx --show-sdk-path)"
                # Only build fat on x86
                if [ "%m" != "arm64" ]; then
                         DIST_CONFIG_PARAMS="${DIST_CONFIG_PARAMS} --with-universal-archs=intel-64"
                fi
        fi
```

dhomeier:
Yes, `configure` only knows of `x86_64` and `ppc64` as (64bit) `MACOSX_DEFAULT_ARCH`s.

TheSin-:
So far this did work, it's compiling while see if it finished

TheSin-:
@nieder can you confirm sanity before I commit it?

TheSin-:
Technically we dont' need universals on x86 even so

```
# for macOS 11, need help finding libz in new system-library cache
        # https://bugs.python.org/issue41116
        # pthread is still not found, so disable test_ctypes for now as well
        if [ "$(uname -r | cut -d. -f1)" -ge 20 ]; then
                DIST_CONFIG_PARAMS="--enable-universalsdk=$(xcrun --sdk macosx --show-sdk-path) --with-universal-archs=intel-64"
        fi
```

can just be removed completely and just add isysroot to CFLAGS

TheSin-:
[...]

PR #983, opened 2023-02-02
TclTk update to 8.6.13 fixing large integer bug

dhomeier:
Fixes a number of test failures like https://github.com/fink/fink-distributions/issues/956#issuecomment-1407521945 and by way of this also the `test_tcl` failures in our Python builds.

Checking the changes made though the last upstream update, all library versions that have changed are again just the non-public ones like
```
		!%p/lib/itcl4.2.3/libitcl4.2.3.dylib
		!%p/lib/sqlite3.40.0/libsqlite3.40.0.dylib
```
so this should be OK to update without packaging as a new version?


dmacks:
blocking on #923

PR #988, opened 2023-02-03
libicu72

dmacks:
This is the latest upstream version, which is yet another new libversion. I also inverted the package-layout (-shlibs as parent) to make it easy to turn off the -dev packages of old libversions.

@TheSin- you recently added libicu67, but it does not appear to be used by any packages in the public distro yet. If this 72 works for you, do we still need that 67, or can we kill it altogether?

I support the idea of offloading the runtime tools to a new -bin package of some sort, which is what Debian already does. But no sense doing that until the libversion chaos settles down.


TheSin-:
I will build and test this tomorrow, sorry I have been pre occupied by glib2

TheSin- (approved):
build is good, used on nodejs and it worked, 67 can be removed after this and and I update nodejs.

I have a new nodejs ready to commit once this is released.

TheSin-:
using icu72 with php7.3 I get

```
/opt/sw/src/fink.build/php7.3-7.3.12-1.2/php-7.3.12/ext/intl/collator/collator_sort.c:349:26: error: use of undeclared identifier 'TRUE'
        collator_sort_internal( TRUE, INTERNAL_FUNCTION_PARAM_PASSTHRU );
                                ^
/opt/sw/src/fink.build/php7.3-7.3.12-1.2/php-7.3.12/ext/intl/collator/collator_sort.c:543:26: error: use of undeclared identifier 'FALSE'
        collator_sort_internal( FALSE, INTERNAL_FUNCTION_PARAM_PASSTHRU );
                                ^
```

so 67 will need to stay for php7.3

dmacks:
Thanks for the check. If nodejs builds on x86_64, I can look at libicu issues.

TheSin-:
once this is committed I'll commit the x64 fix

PR #993, opened 2023-02-09
Patch libnss3.0e to build with Fink Python (not /usr/bin/python)

dhomeier:
The gyp build hardcodes `python` or `python2` in various scripts and configuration files that all need to be fixed; probably something to occur more frequently in gyp-based packages. Fixing build on 12.3+
AFAICT only the build process is changed, no installed files, thus not incrementing revision yet.
Tested with gyp-py27 and gyp-py310 (which should be more stable long-term) on 10.14.6, 13.2.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info
@@ -4,7 +4,7 @@ Revision: 1
 # gyp version to avoid needing BuildDepends:xcode.app (Issue #865)
 BuildDepends: <<
 	fink (>= 0.28),
-	gyp-py27 (>=  0.1+20210831gitd6c5dd5-1),
+	gyp-py310 | gyp-py27 (>=  0.1+20210831gitd6c5dd5-1),
 	ninja,
 	nspr.0f,
 	sqlite3-dev
@@ -20,9 +20,14 @@ SourceDirectory: nss-%v
 PatchFile: %n.patch
 PatchFile-MD5: 47ea72df5a1dc6f6a3a83b7e61341d60
 PatchScript: <<
+#!/bin/sh -ev
 	%{default_script}
 	# fix install_name to have absolute path to libdir
 	perl -pi -e 's,\@executable_path,%p/lib/nss3.0e,g' nss/coreconf/Darwin.mk nss/lib/freebl/config.mk nss/coreconf/config.gypi
+	# patch Python executable to match gyp's (crucial for 12.3+ without system python[2])
+	gyppy=$(awk -F 'python' 'NR==1{print $2}' %p/bin/gyp)
+	perl -pi -e "s,(/bin/env python)(2?)\$,\${1}$gyppy," nss/coreconf/*.py
+	perl -pi -e "s,(python)( |'),\${1}$gyppy\${2}," nss/build.sh nss/coreconf/config.gypi nss/lib/ckfw/builtins/builtins.gyp nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp
 <<
 CompileScript: <<
 #!/bin/sh -ev
@@ -139,6 +144,10 @@ DescPackaging: <<
 	the headers, a pre-installed version's files are not visible.
 
 	nss-config uses pkg-config at runtime
+
+	build script and a bunch of gyp config files hardcoded to use
+	'python' even when running Fink's py-versioned gyp; need to
+	patch all as /usr/bin/python no longer available.
 <<
 License: GPL/LGPL
 Maintainer: Daniel Macks <dmacks@netspace.org>


nieder:
This is setting just `python2.7`. Does it matter if it picks up %p/bin/python2.7 vs /usr/bin/python2.7 on << 12.3?

TheSin-:
Why doesn’t it use fink’s gyp which I patched already?

dhomeier:
> This is setting just `python2.7`. Does it matter if it picks up %p/bin/python2.7 vs /usr/bin/python2.7 on << 12.3?

That's effectively what it has been doing all along when just calling `python` or `/usr/bin/env python`, which apparently had not been an issue so far (mainly calling rather simple scripts). I was more concerned if `python3.10` would work, but the relevant build scripts seem all python3-ready.

dhomeier:
> ahh I see it is using fink's gyp, but it also has it's own hard coded calls. That's what I get for trying to figure it out form my phone ;)

:D
It's using the interpreter from a `'<(python)'` variable in several places, which "helpfully" is not inherited from the gyp-py installation, but set in various .gyp[i] config files.

TheSin-:
yeah at first glance (on my phone, which I hate) I thought it was building a local gpy, that was my bad I spoke out of turn.

dmacks:
It feels fragile to introspect bin/gyp to find python interp and then patch that into a lot of places. Though I don't doubt that you have found lots of places it needs to be in sync, and definitely not rely on 'first "python" in PATH". Can we just pick a specific gyp-pyXX and use it on all platforms?

dhomeier:
The `'<(python)'` is set in the respective project's .gypi configuration files and for all I can tell does not care what the `gyp` program is built with. A lot of those gyp build systems are a horrible mess IMO, and pretty sure I've seen a number with unfixable Python 2-syntax, so I am afraid settling on a single gyp-pyXX is no guarantee those will work automagically.

dhomeier:
1. The issue IMO is not so much figuring out which python the installed `gyp` is using, but finding all the config files where the default interpreter is set. Just patching `build.sh` as in https://github.com/fink/fink-distributions/commit/1ce7d9e still fails on
```
run_scanbuild gyp -f ninja --depth=/opt/x64/src/fink.build/libnss3.0e-shlibs-3.74-1/nss-3.74/nss --generator-output=. -Dnspr_include_dir=/opt/x64/include/nspr -Dnspr_lib_dir=/opt/x64/lib/nspr.0f -Duse_system_sqlite=1 -Dtarget_arch=x64 -Denable_sslkeylogfile=1 -Dnss_dist_dir=/opt/x64/src/fink.build/libnss3.0e-shlibs-3.74-1/nss-3.74/dist -Dnss_dist_obj_dir=/opt/x64/src/fink.build/libnss3.0e-shlibs-3.74-1/nss-3.74/dist/Debug /opt/x64/src/fink.build/libnss3.0e-shlibs-3.74-1/nss-3.74/nss/nss.gyp
/bin/sh: python: command not found
gyp: Call to '"python" ./coreconf/detect_host_arch.py' returned exit status 127 while in /opt/x64/src/fink.build/libnss3.0e-shlibs-3.74-1/nss-3.74/nss/nss.gyp.
```
even when also patching the ninja build to compile `#!/usr/bin/env python3` into the executable.
Those definitions look like `'python%': 'python'`, while references to the defined variable, which should not be changed, look like `'python%': '<(python)'`, so to search-and-cover every potential config file with a regex replacement is not so straightforward.
2. If the point is to only set a single interpreter everywhere; this does not simplify very much – see above, and it is still not done with patching `build.sh` and all the config files, as the build will also call `gyp`, when `%p/bin/gyp` resolves to a higher `gyp-pyNNN` variant, if installed. So to really ensure it is also using the same `python` that `gyp` is using, those straightforward `gyp` calls would also have to be patched, or `gyp-py3NN` made a BuildConflict.
3. If still pinning to a single `gyp-py`, I'd strongly recommend against picking python27 or all things – I think it has been hard enough to revive (more or less) from the dead this time, and fear this will just become harder with future Xcode updates. I think there were packages with so ancient gyp setup that it may be unavoidable to recourse to 2.7, but this is not one of them.

nieder:
We can hardcode a specific gyp version by setting `export GYP=%p/bin/gyp-py39` in CompileScript. nss/build.sh then uses that.
I've also been getting consistent results by changing just this in `coreconf/config.gypi`
```diff
    'variables': {
      'variables': {
        'variables': {
-         'python%': 'python',
+         'python%': '%p/bin/python3.9',
        },
        # chromium uses pymod_do_main, but gyp doesn't set a sensible
        # Python sys.path (gyp_chromium does).
```
I'm going to post the diff caused by the PatchScript changes in the next comment since I find it easier to understand the output than the input perl-pies.

On 13.5 the above starts building until an eventual failure about unused variable, but that's a different issue we can deal with after fixing the python issue.
```
../../lib/pkcs12/p12d.c:2773:12: error: variable 'noInstallCnt' set but not used [-Werror,-Wunused-but-set-variable]
    int i, noInstallCnt, probCnt, bagCnt, errorVal = 0;
           ^
1 error generated.
```

nieder:
This is the output of the perl-pies from here:
```patch
$ diff -ruN nss-3.74{-orig,}                                      
diff -ruN nss-3.74-orig/nss/build.sh nss-3.74/nss/build.sh
--- nss-3.74-orig/nss/build.sh	2022-01-06 05:38:58
+++ nss-3.74/nss/build.sh	2023-09-03 20:16:08
@@ -69,7 +69,7 @@
 ninja_params=()
 
 # Assume that the target architecture is the same as the host by default.
-host_arch=$(python "$cwd/coreconf/detect_host_arch.py")
+host_arch=$(python2.7 "$cwd/coreconf/detect_host_arch.py")
 target_arch=$host_arch
 
 # Assume that MSVC is wanted if this is running on windows.
diff -ruN nss-3.74-orig/nss/coreconf/check_cc.py nss-3.74/nss/coreconf/check_cc.py
--- nss-3.74-orig/nss/coreconf/check_cc.py	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/check_cc.py	2023-09-03 20:15:59
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 import os
 import subprocess
diff -ruN nss-3.74-orig/nss/coreconf/config.gypi nss-3.74/nss/coreconf/config.gypi
--- nss-3.74-orig/nss/coreconf/config.gypi	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/config.gypi	2023-09-03 20:16:08
@@ -7,7 +7,7 @@
     'variables': {
       'variables': {
         'variables': {
-          'python%': 'python',
+          'python%': 'python2.7',
         },
         # chromium uses pymod_do_main, but gyp doesn't set a sensible
         # Python sys.path (gyp_chromium does).
diff -ruN nss-3.74-orig/nss/coreconf/detect_host_arch.py nss-3.74/nss/coreconf/detect_host_arch.py
--- nss-3.74-orig/nss/coreconf/detect_host_arch.py	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/detect_host_arch.py	2023-09-03 20:15:59
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 #
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this file,
diff -ruN nss-3.74-orig/nss/coreconf/sanitizers.py nss-3.74/nss/coreconf/sanitizers.py
--- nss-3.74-orig/nss/coreconf/sanitizers.py	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/sanitizers.py	2023-09-03 20:15:59
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python2.7
 
 from __future__ import print_function
 import sys
diff -ruN nss-3.74-orig/nss/coreconf/shlibsign.py nss-3.74/nss/coreconf/shlibsign.py
--- nss-3.74-orig/nss/coreconf/shlibsign.py	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/shlibsign.py	2023-09-03 20:15:59
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python2.7
 #
 # This Source Code Form is subject to the terms of the Mozilla Public
 # License, v. 2.0. If a copy of the MPL was not distributed with this
diff -ruN nss-3.74-orig/nss/coreconf/werror.py nss-3.74/nss/coreconf/werror.py
--- nss-3.74-orig/nss/coreconf/werror.py	2022-01-06 05:38:58
+++ nss-3.74/nss/coreconf/werror.py	2023-09-03 20:15:59
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2.7
 
 import os
 import subprocess
diff -ruN nss-3.74-orig/nss/lib/ckfw/builtins/builtins.gyp nss-3.74/nss/lib/ckfw/builtins/builtins.gyp
--- nss-3.74-orig/nss/lib/ckfw/builtins/builtins.gyp	2022-01-06 05:38:58
+++ nss-3.74/nss/lib/ckfw/builtins/builtins.gyp	2023-09-03 20:16:08
@@ -30,7 +30,7 @@
         {
           'msvs_cygwin_shell': 0,
           'action': [
-            'python',
+            'python2.7',
             'certdata.py',
             'certdata.txt',
             '<@(_outputs)',
diff -ruN nss-3.74-orig/nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp nss-3.74/nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp
--- nss-3.74-orig/nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp	2022-01-06 05:38:58
+++ nss-3.74/nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp	2023-09-03 20:16:08
@@ -30,7 +30,7 @@
         {
           'msvs_cygwin_shell': 0,
           'action': [
-            'python',
+            'python2.7',
             '../certdata.py',
             'certdata-testlib.txt',
             '<@(_outputs)',
```
I think setting GYP would deal with most of the problems being discussed about reproducibility.

dmacks:
I'm getting confused here about different approaches and different ways of implementing them (or concepts vs actual patches). I really don't care the approach. So if @nieder has something that is fairly robust/self-consistent, feel free to commit instead of this PR. Or if this PR itself is better, feel free to say so and I'll resolve the conflicts and merge it.

nieder:
There are two issues:
1) build sysstem calls naked 'gyp'. Setting $GYP allows picking a specific one that we want (prob gyp-py310).
2) build system calls naked 'python'. This fails when system-python is python3. Setting $PYTHON doesn't work.
2a) Patch with wide coverage perl-pies (see 7545408) to a preferred hardcoded python, or
2b) More targetd patch for `coreconf/config.gypi` to change the gyp variable `python%` to our preferred hardcoded python.

2b works on 10.14, but I'm currently having issues with it on 13.5 (still some vanilla-python left). So the build might need a more brute force approach from 2a, but cleaned up to use a single python, vs current grokking what variant gyp is.

I also had to set `SetCFLAGS: -Wno-error=unused-but-set-variable` so that the build wouldn't fail on 13.5 before it got to any of the other python calls.

nieder:
This is my minimal patchset. No rev up because I think all the python stuff is buildtime. The perl-pies are based on @dhomeier 's but w/out grokking gyp flavors.
```diff
diff --git a/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info b/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info
index 2698ec1975..6e24c780a3 100644
--- a/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/libnss3.0e-shlibs.info
@@ -4,12 +4,13 @@ Revision: 1
 # gyp version to avoid needing BuildDepends:xcode.app (Issue #865)
 BuildDepends: <<
 	fink (>= 0.28),
-	gyp-py27 (>=  0.1+20210831gitd6c5dd5-1),
+	gyp-py310 (>=  0.1+20210831gitd6c5dd5-1),
 	ninja,
 	nspr.0f,
-	python27,
+	python310,
 	sqlite3-dev
 <<
+SetCFLAGS: -Wno-error=unused-but-set-variable
 GCC: 4.0
 Depends: <<
 	nspr.0f-shlibs,
@@ -21,14 +22,17 @@ SourceDirectory: nss-%v
 PatchFile: %n.patch
 PatchFile-MD5: 47ea72df5a1dc6f6a3a83b7e61341d60
 PatchScript: <<
-	%{default_script}
+	sed 's|@FINK_PREFIX@|%p|g' < %{PatchFile} | patch -p1
 	# fix install_name to have absolute path to libdir
 	perl -pi -e 's,\@executable_path,%p/lib/nss3.0e,g' nss/coreconf/Darwin.mk nss/lib/freebl/config.mk nss/coreconf/config.gypi
 	# match executable to python we know we have
-	perl -pi -e 's,\(python ,\(%p/bin/python2.7 ,' nss/build.sh
+	perl -pi -e "s,/usr/bin/env python3\$,%p/bin/python3.10,g" nss/coreconf/*.py
+	perl -pi -e "s,(/usr/bin/env python)(2?)\$,%p/bin/python3.10," nss/coreconf/*.py
+	perl -pi -e "s,(python)( |'),%p/bin/\${1}3.10\${2}," nss/build.sh nss/coreconf/config.gypi nss/lib/ckfw/builtins/builtins.gyp nss/lib/ckfw/builtins/testlib/builtins-testlib.gyp
 <<
 CompileScript: <<
 #!/bin/sh -ev
+	export GYP=%p/bin/gyp-py310
 	pushd nss
 	./build.sh \
 		--with-nspr=%p/include/nspr:%p/lib/nspr.0f \
```

nieder:
By the way, nss-3.78 is the last upstream to build with our nspr-4.32. However, upstream claims that the 3.X series are all backwards compatible, so later releases would keep this libN if we update nspr.
Upstream also moved to python3 with v3.88 https://bugzilla.mozilla.org/show_bug.cgi?id=1766767 (and added a flag to declare a python version at build time).

dmacks:
> This is my minimal patchset....

Makes sense and gets us going for now. Please commit it, but leave the ticket open for future work.

nieder:
@dmacks ping given 557bc88d65ef1dc30c97213d6705315388669ae3

Also, since there's a new libnss3.0f, I guess that upstream was incorrect about backwards compat of the 3.x series?
https://firefox-source-docs.mozilla.org/security/nss/releases/nss_3_96_1.html#compatibility

dmacks:
I compared the public headers. There were enums that changed values of existing tokens and functions that were removed/renamed. My default assumption is "public headers is the definition of the public interactions", and these sorts of changes would obviously break existing binaries.

I'm on 10.13, so I have /usr/bin/python but not /usr/bin/python3. I welcome any and all patches (feel free to commit directly) that improve the system-python portability.

nieder:
I think the python issue is fixed, or at least well worked around by using fink-python310 everywhere. Once we stop supporting any dist that has /usr/bin/python2 we could go back to /usr/bin/python3, but would be in trouble in a couple years when that gets changed.

I'm going to close this because it's original raison d'etre has been dealt with and there's been other updates to the packages, so we now have merge conflicts. If we see something else affecting how the build interacts with python, we can make a new issue and/or PR.

PR #995, opened 2023-02-14
Patch all Python calls in ninja build to use system /usr/bin/python[3]

dhomeier:
Follow-up to #936, addressing https://github.com/fink/fink-distributions/issues/932#issuecomment-1428157720.
I have not found any evidence that the compiled-in `browse.py` script is used anywhere, but as this does change the binary for 12.0, 13.0 now, may need a new revision.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/ninja.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/ninja.info
@@ -8,14 +8,11 @@ Source-Checksum: SHA256(3c6ba2e66400fe3f1ae83deb4b235faf3137ec20bd5b08c29bfc368d
 SourceRename: %n-%v.tar.gz
 # re2c >= 0.15.3 for its --no-version flag
 BuildDepends: re2c (>= 0.15.3)
-CompileScript: <<
-  #!/bin/sh -ex
-  if [ -x /usr/bin/python ]; then
-    ./configure.py --bootstrap
-  else
-    /usr/bin/python3 configure.py --bootstrap
-  fi
-<<
+
+PatchScript: [ -x /usr/bin/python ] || perl -pi -e 's|(/bin/env python)$|${1}3|' configure.py src/browse.py misc/*.py
+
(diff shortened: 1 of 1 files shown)


cooljeanius (approved):

nieder:
_Originally posted by @nieder in https://github.com/fink/fink-distributions/issues/932#issuecomment-1706397337_

Like @dmacks said, the python declared at buildtime gets hardcoded into the binary.
I've tried somethings out on their build system and tested variations: This is sufficient to build with whatever python we want and have that encoded in the binary, with no patching needed. The first python3.9 is to run the configure.py script, the second python3.9 is what ninja will use during its runtime. Tested on 10.14 and 13.5 using system-python and python39.

```
CompileScript: %p/bin/python3.9 ./configure.py --bootstrap --verbose --with-python=%p/bin/python3.9
```
In this case, i have `%p/bin/python3.9`, but that can be set to whatever python we decide on. If we pick `/usr/bin/python[23]`, we'll need a separate .info file for macOS 12 and up because ninja built using system-python2 will fail on an upgrade to macOS 12+.
So what wins: ease of maintenance with a single .info and a hardcoded fink-python (that can be upgraded as needed)? Or remove a heavy dep but need to keep various .info files in sync?

/usr/bin/python2 got removed with macOS12.3. macOS 10.15 CLT included /usr/bin/python3.

dmacks:
> _Originally posted by @nieder in [#932 (comment)](https://github.com/fink/fink-distributions/issues/932#issuecomment-1706397337)_
> 
> hardcoded fink-python...a heavy dep

Just a thinking-ahead...#1002 substantially slims down some upcoming new fink-python.

dhomeier:
> In this case, i have `%p/bin/python3.9`, but that can be set to whatever python we decide on.

310 has entered the security bugfix phase as of 3.10.12, so that 3.11 or 3.12 would certainly be preferable – especially if those can be built as a stripped-down base package without tk – any news on that endeavour?

PR #1003, opened 2023-02-26
openmotif 2.3.8

PovlAbrahamsen:
Here is an updated version of OpenMotif, which should allow us to build Motif programs against the two-level namespace version of libxt!
I've tested this with nedit - and it appears to work as intended. One thing I am less sure about is the format for the source URL: it includes spaces, which don't seem to be escaped correctly by Fink. So I have downloaded the tarball separately and moved it into /opt/sw/src. Any advise on how to fix this would be very welcome.


nieder:
What happens to an app that's linking to openmotif-2.3.4 that links to libxt-flat when openmotif is updated to this version?
Should we put this openmotif in a subdirectory so the two can coexist while we transition the ~15 other packages that currently use libxt-flat and openmotif4-flat?

PovlAbrahamsen:
I've just tried downgrading nedit to the version linked against the flat namespace openmotif/libxt. If I run `export DYLD_LIBRARY_PATH=/opt/X11/lib/flat_namespace` (which was necessary to run some of these programs before, e.g. mbsystem - though not nedit), the program still works. If I run the program without setting `DYLD_LIBRARY_PATH`, it fails with 
> Error: Shell widget nedit has zero width and/or height

The version that is linked to the non-flat-namespace libraries runs regardless of whether or not `DYLD_LIBRARY_PATH` has been set.

So my inclination would be to replace the older library version, rather than keeping parallel libraries. And I'd hope that the "flat-namespace" versions of libxt can be phased out eventually.

nieder:
Geomview built with openmotif4-2.3.4/libxt-flat silently fails when openmotif4-2.3.8 is installed. Setting DYLD_LIBRARY_PATH doesn't help (this is 10.14.6). If I rebuild it with the new openmotif4, then it runs.

DYLD_LIBRARY_PATH stopped working from scripts around 10.10 or 10.11, if I recall correctly, except in some very specific circumstances.

List of packages that use libxt-flat. Only mesa-libglw and orrery don't also use openmotif4 (but orrery uses geomview which uses openmotif4)
- [ ] afni
- [ ] cernlib2006
- [ ] ddd
- [ ] dx
- [ ] geant4.9
- [ ] geomview195
- [ ] grace
- [ ] mesa-libglw
- [ ] molmol
- [ ] nedit
- [ ] openmotif4
- [ ] orrery
- [ ] texlive
- [ ] xbae
- [ ] xephem
- [ ] xmakemol
- [ ] xpdf

dmacks:
I agree with nieder that a change to a .dylib that is not backward-compatible needs a new Shlibs filename in a new Package suite. That way we can take our time cascading the updates that switch other packages to use it.

dmacks:
Regarding the source filename with whitespace, if you can't figure it out (it's true, our Source field has trouble with certain characters!), you can use Debian's mirror of it:
```
Source: mirror:debian:pool/main/m/motif/motif_%v.orig.tar.gz
SourceRename: motif-%v.tar.gz
```

nieder:
One nice thing about Sourceforge files, is that you don't need to include the subdirectories when generating a download URL. SF will find the file.

So `https://sourceforge.net/projects/motif/files/motif-2.3.8.tar.gz` will translate to `Source: mirror:sourceforge:motif/motif-%v.tar.gz` and no need to keep track of version subdirs.

nieder:
@PovlAbrahamsen Can I take your commit f438ed0 and change the library install path to a separate directory so it can coexist with the older flat-namespace openmotif? Then we can transition the needed packages to the new openmotif. Unfortunately, it doesn't look like every package will handle an in-place replacement of flat-openmotif to 2level-openmotif.

@dmacks should openmotif4-bin also be renamed and moved (ie change `--prefix` for the whole suite) and C/R current flat-openmotif4-bin, or only change `--libdir` and have 2level-openmotif4-bin just be a %v bump over current flat-openmotif4-bin? %n-bin has a bunch of bindings in %p/lib/X11, but I'm hoping the library (and its dependees) is smart enough to find them in a new `--bindir`.

ETA: thinking more, we probably want a new named `openmotif4-twolevel-bin` that C/R the existing openmotif4-bin because that's the whole reason we're making a new package: we don't know what will happen if we mix old package built against old flat-openmotif4 with new 2level-openmotif4-shlibs. geant4.9 is the only known user of flat-openmotif4-bin.

PovlAbrahamsen:
@nieder, by all means, please do feel free to redo this PR as you see fit!

PovlAbrahamsen:
Superseded by #1043

PR #1009, opened 2023-03-02
Libvpx update to v1.13.0

nieder:
Update libvpx to latest upstream tag v1.13.0
This knows about macOS through Ventura and also knows Apple silicon.


dmacks:
LGTM on 10.13, if that's worth anything.

nieder:
I don't have ARM but do have 12.6 and it works there. Will merge soon unless an ARM user sees an issue.

dhomeier (approved):
Both building for me on 12.6.3/arm64 fwiw.

PR #1010, opened 2023-03-11
Boost icu upgrade

dmacks:
As part of a sweep to upgrade as many packages as possible to the latest libicuXX libversion, this is my quick pass at the boost* package sets.

I did not investigate further what caused the non-detection in the oldest boost libversions (1.53 and 1.55). But one clue is I'm fairly sure that libicu as of 6x-ish requires c++11, so older boost might not be using that. Given those boost are already dist-restricted and not used, probably not worth any time thinking more about it. We'll just have to dist-restrict (rather than kill off altogether) the libicu55 headers package in the future.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/boost1.53..info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/boost1.53..info
@@ -1,5 +1,6 @@
 Info2: <<
 Package: boost1.53.%type_pkg[python]
+# libicu72 does not get detected
 Version: 1.53.0
 Revision: 8
 Type: python (nopython systempython python2.7)
--- a/10.9-libcxx/stable/main/finkinfo/libs/boost1.53..info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/boost1.53..info
@@ -1,8 +1,7 @@
 Info2: <<
 Package: boost1.53.%type_pkg[python]
-# libicu72 does not get detected
 Version: 1.53.0
-Revision: 9
+Revision: 10
 Type: python (nopython systempython python2.7)
 # Do not propagate past 10.10 if at all possible. Use boost1.55 or later.
 Distribution: <<
@@ -32,7 +31,7 @@ Maintainer: Hanspeter Niederstrasser <nieder@users.sourceforge.net>
 Depends: %N-shlibs (= %v-%r)
 BuildDepends: <<
 	fink (>= 0.34.4),
-	libicu55-dev,
+	libicu72-dev,
 	(%type_num[python]) %type_pkg[python]
 <<
 BuildDependsOnly: True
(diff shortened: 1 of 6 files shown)


nieder:
from boost1.55::config.log
```
darwin.compile.c++ bin.v2/libs/locale/build/darwin-4.2.1/debug/has_icu_obj.o

    "g++"  -ftemplate-depth-128 -O0 -fno-inline -Wall -g -dynamic -gdwarf-2 -fexceptions -fPIC  -DBOOST_ALL_NO_LIB=1  -I"." -c -o "bin.v2/libs/locale/build/darwin-4.2.1/debug/has_icu_obj.o" "libs/locale/src/../build/has_icu_test.cpp"

libs/locale/src/../build/has_icu_test.cpp:12:10: fatal error: 'unicode/uversion.h' file not found
#include <unicode/uversion.h>
         ^~~~~~~~~~~~~~~~~~~~
1 error generated.
...failed darwin.compile.c++ bin.v2/libs/locale/build/darwin-4.2.1/debug/has_icu_obj.o...
...skipped <pbin.v2/libs/locale/build/darwin-4.2.1/debug>has_icu for lack of <pbin.v2/libs/locale/build/darwin-4.2.1/debug>has_icu_obj.o...
```

From boost1.78::config.log:
```
clang-darwin.compile.c++ bin.v2/libs/locale/build/clang-darwin-11/release/cxxstd-14-iso/threading-multi/visibility-hidden/has_icu_obj.o

    "clang++" -x c++ -std=c++14 -fvisibility-inlines-hidden -fPIC -m64 -O3 -Wall -fvisibility=hidden -Wno-inline -MD -I/sw/include -DBOOST_ALL_NO_LIB=1 -DNDEBUG -I"."  -c -o "bin.v2/libs/locale/build/clang-darwin-11/release/cxxstd-14-iso/threading-multi/visibility-hidden/has_icu_obj.o" "libs/locale/src/../build/has_icu_test.cpp"

libs/locale/src/../build/has_icu_test.cpp:26:12: warning: unused variable 'c' [-Wunused-variable]
   UChar32 c = ::u_charFromName(U_UNICODE_CHAR_NAME, "GREEK SMALL LETTER ALPHA", &err);
           ^
1 warning generated.
clang-darwin.link bin.v2/libs/locale/build/clang-darwin-11/release/cxxstd-14-iso/threading-multi/visibility-hidden/has_icu
```

boost1.78 passes `cxxflags="-MD -I%p/include"` during the ./b2 command. ./bootstrap also has `--with-icu` for 1.78.

`--with-icu` has no effect on 1.55. passing cxxflags to 1.55 leads to this:
```
    - iconv (libc)             : no
    - iconv (separate)         : no
    - icu                      : no
    - icu (lib64)              : no
- Boost.Locale needs either iconv or ICU library to be built.
```
which config log says it's about not finding `-licuuc`. but adding `linkflags="-L%p/lib` seems to fix that.

TL;DR: try adding `cxxflags="-MD -I%p/include" linkflags="-L%p/lib"` to the ./b2 command for boost1.53 and 1.55.

nieder:
Well, that helps with better detection of icu55 and iconv, but bumping to libicu72 fails (from config.log):
```
In file included from libs/regex/build/has_icu_test.cpp:15:
In file included from /sw/include/unicode/coll.h:61:
/sw/include/unicode/uobject.h:134:54: error: expected ';' at end of declaration list
    static void * U_EXPORT2 operator new(size_t size) U_NOEXCEPT;
```
So probably a cxxstd thing.

We might want to add `cxxflags="-MD -I%p/include" linkflags="-L%p/lib"` anyway,  but keep it at icu55.

nieder:
Otherwise LGTM if you just want to boost the working ones to newer libicu.

dmacks:
Thanks for that hint about flag-passing. While propagating that into the packages, I saw that boost1.58, the lowest one that accepted libicu72 already had an explicit c++11 in one of those flags. Back-porting that to 1.5 and 1.53 allowed those also to accept it. So now all the boost are at libicu72 and have updated flags for maximal "yes" diagnostics for icu detection.

nieder:
I'm fine with these (untested by me) changes. They shouldn't mess up boost-using things.

dmacks:
Thanks. Libicu drops APIs at times, but doesn't seem to change signatures of existing ones, and I've tested I most of the packages that have dependencies on [boost + libicu] to verify that they still build.

PR #1011, opened 2023-03-15
Miktex tools update

nieder:
Bump miktex-tools to the latest release.
The jpeg typedef fix is no longer needed because the file is gone.
Still not OpenSSL3 friendly.

Asking @dmacks because he recently tested/updated our current .info


dmacks (approved):
LGTM on 10.13

Issue #1007, opened 2023-02-28
Failed: phase compiling: libvpx16-1.6.1-1 failed


dak180:
```
dpkg-deb -b /opt/sw/src/fink.build/root-fink-buildlock-libvpx16-1.6.1-1 /opt/sw/src/fink.build
dpkg-deb: building package 'fink-buildlock-libvpx16-1.6.1-1' in '/opt/sw/src/fink.build/fink-buildlock-libvpx16-1.6.1-1_2023.02.28-09.48.15_darwin-arm64.deb'.
Installing build-lock package...
/opt/sw/bin/dpkg-lockwait -i /opt/sw/src/fink.build/fink-buildlock-libvpx16-1.6.1-1_2023.02.28-09.48.15_darwin-arm64.deb
Selecting previously unselected package fink-buildlock-libvpx16-1.6.1-1.
(Reading database ... 256675 files and directories currently installed.)
Preparing to unpack .../fink-buildlock-libvpx16-1.6.1-1_2023.02.28-09.48.15_darwin-arm64.deb ...
Unpacking fink-buildlock-libvpx16-1.6.1-1 (2023.02.28-09.48.15) ...
Setting up fink-buildlock-libvpx16-1.6.1-1 (2023.02.28-09.48.15) ...
sudo -u fink-bld [ENV] sh -c /tmp/fink.EIFar
env LANG=C LC_ALL=C /opt/sw/bin/tar  --no-same-owner --no-same-permissions -xf /opt/sw/src/libvpx-1.6.1.tar.bz2
sudo -u fink-bld [ENV] sh -c /tmp/fink.4iw2B
perl -pi -e 's/-isysroot \${osx_sdk_dir}//g' build/make/configure.sh
sudo -u fink-bld [ENV] sh -c /tmp/fink.iNu00
/tmp/fink.Y7Cda
#!/bin/bash -ev
	
	let version=`uname -r | cut -d . -f 1`
uname -r | cut -d . -f 1
	if [ ${version} -gt 15 ]; then
		version=15
	fi
	./configure --prefix=/opt/sw --target=arm64-darwin${version}-gcc --extra-cflags="-DGTEST_USE_OWN_TR1_TUPLE=1" --enable-shared --disable-static --disable-unit-tests
  enabling shared
  disabling static
  disabling unit_tests
  enabling vp8_encoder
  enabling vp8_decoder
  enabling vp9_encoder
  enabling vp9_decoder
[...]

PR #1014, opened 2023-03-21
Graphics dep upgrades

dmacks:
These bump some of the last few reverse-depends on libvpx16 (reported problematic on new platforms in #1007) and libx264-118 (noticed along the way). If we can get fully migrated off these, we can dist-restrict or nuke altogether and reduce future maintenance and bit-rotting on new platforms.

The only remaining is libavcodec57-3.0, which FTBFS because it requires "sdl2 >= 2.0.1 sdl2 < 2.1.0" but our sdl2 is 2.26. This old libavcodec is a -shlibs only stub since Jan 2020, maybe time to nuke it altogether?


nieder (approved):
vpx and x264 updates look fine on paper

Issue #863, opened 2021-12-27
libusb-0.1.12 FTBFS on macOS12 (deprecated IOMasterPort)


nieder:
The older libusb-0.1.12 fails to build on macOS 12.1 with this error:
```
 gcc -DHAVE_CONFIG_H -I. -I/opt/sw/include -Werror -no-cpp-precomp -g -O2 -g -Wall -MT darwin.lo -MD -MP -MF .deps/darwin.Tpo -c darwin.c  -fno-common -DPIC -o .libs/darwin.o
darwin.c:1171:5: error: 'IOMasterPort' is deprecated: first deprecated in macOS 12.0 [-Werror,-Wdeprecated-declarations]
    IOMasterPort(masterPort, &masterPort);
    ^~~~~~~~~~~~
    IOMainPort
/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/System/Library/Frameworks/IOKit.framework/Headers/IOKitLib.h:132:1: note: 'IOMasterPort' has been explicitly marked deprecated here
IOMasterPort( mach_port_t       bootstrapPort,
^
1 error generated.
make[2]: *** [darwin.lo] Error 1
```
This old libN is long abandoned. It's used by the following packages:
- [ ] gnupg-unified (@nieder)
- [ ] xpwn (none)
- [ ] openocd (none)
- [ ] avrdude (@mringwal)
- [ ] mspdebug (none)
- [ ] libftdi (none)
- [ ] libgphoto2.6 (@nieder)
- [ ] ztex (none)
- [ ] nut (@clepple )
- [ ] gnokii7 (Ben Reed)
- [ ] gpsbabel (@clepple )
- [ ] sispmctl (none)

dmacks:
This FTBFS can surely be resolved by removing `-Werror`, but we should also work on migrating as much as possible off an old libversion.

dmacks:
I'm tagging "OLD_USB" for ones that cannot easily be migrated.

clepple:
Theoretically, we're close to [upgrading NUT to libusb-1.0](https://github.com/networkupstools/nut/issues/300), but TBH userspace access to USB devices on macOS seems to keep breaking at runtime in ways that aren't easy to debug or fix (despite libusb-0.1 having worked just fine on other platforms for years after it was abandoned upstream).

dmacks:
[...]

PR #1017, opened 2023-03-27
nut: 2.8.0, which is new libversions

dmacks:
This bumps nut to 2.8.0 (clearing an item from #863), and retains the older 2.7.4 at a new .info filename. There are several component -shlibs packages whose install_name jumped, so this preserves the older ones for whatever external packages might use them. But I have no objection to just killing off the old suite altogether. Are there any dependants other than nut itself?


nieder:
nut-2.7.4.info calls `nut-2.7.4-openss110.patch`, but it is using openssl300.

dmacks:
> nut-2.7.4.info calls `nut-2.7.4-openss110.patch`, but it is using openssl300.

Isn't that crazy? It's a 110+ compatibility patch; 300 retains the legacy interfacy (though IIRC there are compile-time warnings that it's an old API...really is 110ish in content). Better to rename it just "...openssl.patch"?

nieder:
rename is probably nicer (or a DescPackaging note that it's using the 110 API from 300).

nieder:
@clepple ping

clepple:
TBH I'm probably going to have to give up maintainership of my packages because I can't seem to find time to debug the Fink installations here, let alone track the complexities of recent NUT and OpenSSL upstream. All of these changes sound good in principle, but I can't test them.

nieder:
@dmacks these look fine but untested locally. If they built for you go ahead since it almost helps close up #863

nieder (approved):
LGTM

PR #1030, opened 2023-05-28
lighttpd: update to lighttpd 1.4.71

gstrauss:
lighttpd: update to lighttpd 1.4.71

lighttpd 1.4.25 was released in Nov 2009, over 13 years ago!!!

There are a number of important security and bug fixes that are missing in fink lighttpd.info, besides many improvements.  

Modern lighttpd releases are (generally) still portable to run on very old systems.  The current version of lighttpd is also available for Macs in https://github.com/macports/macports-ports, under https://github.com/macports/macports-ports/tree/master/www/lighttpd

lighttpd release notes over the years announce deprecation and migration paths of some modules which may have been in lighttpd 1.4.25, but are no longer in current lighttpd releases.

mod_cml -> mod_magnet
mod_compress -> mod_deflate
mod_geoip -> mod_maxminddb

lighttpd mod_magnet works with any version of lua 5.1 or later, but I have left the dependency in lighttpd.info at lua51.

lighttpd builds a number of modules into the base lighttpd executable instead of having separate .so (or .dylib) for each module.  (Each modules takes at least (5) 4k pages in memory, and the 12 small modules now built-in to lighttpd use about the same amount of memory as a single small .so.  I have modified lighttpd.info to remove the .so from the module, though the documentation may still be present.

A long time ago, but after lighttpd 1.4.25, some of the documentation in the lighttpd source tree was moved from `doc/*.txt` to `doc/outdated/*.txt` as some of those docs were not maintained and may be out-of-date.  This might require changes to fink lighttpd.info.

mod_secdownload, mod_trigger_before_dl, mod_usertrack, mod_flv_streaming, mod_evasive have all be removed, but replacements are available using lighttpd mod_magnet and lua scripts https://wiki.lighttpd.net/ModMagnetExamples#lua-examples-of-deprecated-lighttpd-modules

mod_mysql_vhost -> mod_vhostdb, mod_vhostdb_mysql but note that I did not (yet?) add mod_vhostdb and mod_vhostdb_mysql, or the alternative mod_vhostdb_dbi with DBI dependencies and SplitOffs to fink lighttpd.info.  Should I add mod_vhostdb, mod_vhostdb_mysql, mod_vhostdb_dbi to lighttpd.info?  mod_vhostdb_ldap?  mod_vhostdb_pgsql?  I somehow doubt that an old Mac is being used as server and requires these databases instead of using mod_evhost with maybe an additional symlink tree on local disk.

The third-party mod_useronline is not provided by lighttpd, but could possibly be re-implemented with a lighttpd mod_magnet and a lua script.  It is not likely to be widely used, if still used at all.

I am not familiar with fink packaging, so I would like to mention some "newer" modules in lighttpd not in fink lighttpd.info: mod_sockproxy, mod_vhostdb_*, mod_wstunnel.  mod_h2 is a separate module for HTTP/2 support.  TLS support in lighttpd is separate from the lighttpd base (which had openssl as part of the lighttpd executable in lighttpd 1.4.25).  Separate TLS modules include mod_openssl, mod_gnutls, mod_mbdetls, mod_wolfssl, mod_nss.  It is likely that fink lighttpd.info needs to be updated to include a SplitOff for at least mod_openssl.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
@@ -27,6 +27,29 @@ BuildDepends: <<
 Depends: daemonic, libpcre2-shlibs, openssl300-shlibs
 Provides: httpd
 Recommends: lighttpd-access, lighttpd-accesslog
+# C/R mods that have become obsolete
+Conflicts: <<
+  lighttpd-cml (<< 1.4.64-1),
+  lighttpd-compress (<< 1.4.56),
+  lighttpd-evasive (<< 1.4.68-1),
+  lighttpd-flv-streaming (<< 1.4.64-1),
+  lighttpd-geoip (<< 1.4.54),
+  lighttpd-mysql-vhost (<< 1.4.64-1),
+  lighttpd-secdownload (<< 1.4.68-1),
+  lighttpd-trigger-b4dl (<< 1.4.64-1),
+  lighttpd-usertrack (<< 1.4.68-1)
+<<
+Replaces: <<
+  lighttpd-cml (<< 1.4.64-1),
+  lighttpd-compress (<< 1.4.56),
+  lighttpd-evasive (<< 1.4.68-1),
+  lighttpd-flv-streaming (<< 1.4.64-1),
+  lighttpd-geoip (<< 1.4.54),
+  lighttpd-mysql-vhost (<< 1.4.64-1),
+  lighttpd-secdownload (<< 1.4.68-1),
+  lighttpd-trigger-b4dl (<< 1.4.64-1),
+  lighttpd-usertrack (<< 1.4.68-1)
+<<
 
 Source: https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-%v.tar.xz
 Source-Checksum: SHA256(b8b6915da20396fdc354df3324d5e440169b2e5ea7859e3a775213841325afac)
@@ -213,9 +236,11 @@ SplitOff6: <<
 SplitOff7: <<
   Package: lighttpd-compress
   Description: OBSOLETE: use lighttpd-deflate instead
+  Depends: <<
+    lighttpd-deflate (>= %v-%r)  
+  <<
   RuntimeDepends: <<
-    fink-obsolete-packages,
-    lighttpd-deflate (>= %v-%r)
+    fink-obsolete-packages
   <<
   InstallScript: <<
     mkdir -p %i/share/doc/installed-packages 
@@ -443,6 +468,22 @@ SplitOff25: <<
   Package: lighttpd-magnet
   Depends: %N (=%v-%r), lua51-shlibs
   Description: Plugin: mod_magnet
+  Conflicts: <<
+    lighttpd-cml (<< 1.4.64-1),
+    lighttpd-evasive (<< 1.4.68-1),
+    lighttpd-flv-streaming (<< 1.4.64-1),
+    lighttpd-secdownload (<< 1.4.68-1),
+    lighttpd-trigger-b4dl (<< 1.4.64-1),
+    lighttpd-usertrack (<< 1.4.68-1)
+  <<
+  Replaces: <<
+    lighttpd-cml (<< 1.4.64-1),
+    lighttpd-evasive (<< 1.4.68-1),
+    lighttpd-flv-streaming (<< 1.4.64-1),
+    lighttpd-secdownload (<< 1.4.68-1),
+    lighttpd-trigger-b4dl (<< 1.4.64-1),
+    lighttpd-usertrack (<< 1.4.68-1)
+  <<
   DescDetail: <<
     Control the request handling
   <<
@@ -493,9 +534,11 @@ SplitOff28: <<
 SplitOff29: <<
   Package: lighttpd-geoip
   Description: OBSOLETE: use lighttpd-maxminddb instead
-  RuntimeDepends: <<
-     fink-obsolete-packages,
+  Depends: <<
      lighttpd-maxminddb (>= %v-%r)
+  <<
+  RuntimeDepends: <<
+     fink-obsolete-packages
   <<
   InstallScript: <<
     mkdir -p %i/share/doc/installed-packages 
@@ -539,7 +582,7 @@ SplitOff32: <<
     lib/%N/mod_deflate.so
     etc/%N/mods-available/deflate.conf
   <<
-  ConfFiles: %p/etc/%N/mods-available/compress.conf
+  ConfFiles: %p/etc/%N/mods-available/deflate.conf
 <<
 
 SplitOff33: <<
--- a/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
@@ -1,23 +1,19 @@
 Package: lighttpd
-Version: 1.4.25
-Revision: 7
+Version: 1.4.71
+Revision: 1
 DocFiles: AUTHORS COPYING NEWS README
 Description: Lightning fast web server
 License: BSD
 Maintainer: Alexey Zakhlestin <indeyets@gmail.com>
-HomePage: http://lighttpd.net
+HomePage: https://lighttpd.net
 
 BuildDepends: <<
   autoconf2.6,
   automake1.15,
-  bzip2-dev,
   cyrus-sasl2.3-dev,
   fink (>= 0.32),
   fink-package-precedence,
-  gdbm6,
-  geoip-dev,
-  libiconv-dev,
-  libmemcache,
+  libmaxminddb-dev,
   libpcre2,
   libtool2,
   libxml2 (>= 2.9.1-1),
@@ -32,25 +28,18 @@ Depends: daemonic, libpcre2-shlibs, openssl300-shlibs
 Provides: httpd
 Recommends: lighttpd-access, lighttpd-accesslog
 
-Source: http://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-%v.tar.bz2
-Source-Checksum: SHA256(7e803089f18b179097cb33b64b37d8a3b537ce9c196c88e3fb09881b471c88ce)
-Source2: http://redmine.lighttpd.net/attachments/download/716/mod_geoip_for_1.4.c
-Source2-Checksum: SHA256(0847f26150bc205819666016ecd7085fad51261992f5ed566455d8e997c84c40)
-Source3: http://redmine.lighttpd.net/attachments/download/197/mod_useronline.c
-Source3-Checksum: SHA256(4aef05b78ede5d31babbedef23e4bc10965fe62d754b06e979ef4155d33aea64)
+Source: https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-%v.tar.xz
+Source-Checksum: SHA256(b8b6915da20396fdc354df3324d5e440169b2e5ea7859e3a775213841325afac)
 
 PatchFile: %n.patch
-PatchFile-MD5: 2e2d27342faa1ae46235c70845a33b4a
+PatchFile-MD5: cd3d98f5d61e581171f1e51a7ba51dce
 PatchScript: <<
   cat %{PatchFile} | sed 's|@PREFIX@|%p|g' | patch -p1
-  perl -pi -e 's/AM_C_PROTOTYPES/AC_C_PROTOTYPES\nAM_PROG_AR/' configure.ac
   glibtoolize --copy --force
   autoreconf -fi
-  mv ../mod_geoip_for_1.4.c src/mod_geoip.c
-  mv ../mod_useronline.c src/mod_useronline.c
 <<
 
-ConfigureParams: --libdir='${prefix}/lib/%n' --with-openssl=%p --with-ldap --with-mysql=%p/bin/mysql_config --with-bzip2 --with-gdbm --with-attr --enable-dependency-tracking --with-pcre --with-webdav-props --without-webdav-locks --with-lua --with-memcache
+ConfigureParams: --libdir='${prefix}/lib/%n' --with-openssl=%p --with-ldap --with-mysql=%p/bin/mysql_config --enable-dependency-tracking --with-pcre2 --with-webdav-props --without-webdav-locks --with-lua --with-maxminddb
 CompileScript: <<
 	./configure %c
 	make
@@ -123,21 +112,15 @@ fi
 <<
 
 DescDetail: <<
-lighttpd is a secure, fast, compliant and very flexible web-server which has
-been optimized for high-performance environments. It has a very low memory
-footprint compared to other webservers and takes care of cpu-load. Its
-advanced feature-set (FastCGI, CGI, Auth, Output-Compression, URL-Rewriting
-and many more) make lighttpd the perfect webserver-software for every server
-that is suffering load problems.
+lighttpd (pronounced /lighty/) is a secure, fast, compliant, and very flexible
+web server that has been optimized for high-performance environments. lighttpd
+uses memory and CPU efficiently and has lower resource use than other popular
+web servers. Its advanced feature-set (FastCGI, CGI, Auth, Output-Compression,
+URL-Rewriting and much more) make lighttpd the perfect web server for all
+systems, small and large.
 <<
 DescPackaging: <<
 liblightcomp.dylib is a private library only.
-
-Patched to fix http://trac.lighttpd.net/trac/changeset/1899
-
-dmacks patched response.c for openssl110, analogous to code in newer
-upstream version:
-https://redmine.lighttpd.net/projects/lighttpd/repository/revisions/master/entry/src/mod_openssl.c
 <<
 
 # SplitOffs follow
@@ -147,7 +130,6 @@ SplitOff: <<
   Depends: %N (=%v-%r)
   Description: Plugin: mod_access
   Files: <<
-    lib/%N/mod_access.so
     share/doc/%N/doc/access.txt
     etc/%N/mods-available/access.conf
   <<
@@ -171,7 +153,6 @@ SplitOff3: <<
   Depends: %N (=%v-%r)
   Description: Plugin: mod_alias
   Files: <<
-    lib/%N/mod_alias.so
     share/doc/%N/doc/alias.txt
     etc/%N/mods-available/alias.conf
   <<
@@ -204,8 +185,8 @@ SplitOff5: <<
 
 SplitOff6: <<
   Package: lighttpd-cml
-  Depends: %N (=%v-%r), lua51-shlibs, libmemcache-shlibs
-  Description: Plugin: mod_cml
+  Conflicts: lighttpd (>= 1.4.64)
+  Description: OBSOLETE: use lighttpd-magnet instead
   DescDetail: <<
     Cache Meta Language
   <<
@@ -219,8 +200,12 @@ SplitOff6: <<
 
 SplitOff7: <<
   Package: lighttpd-compress
-  Depends: %N (=%v-%r), bzip2-shlibs
-  Description: Plugin: mod_compress
+  Conflicts: lighttpd (>= 1.4.56)
+  Description: OBSOLETE: use lighttpd-deflate instead
+  RuntimeDepends: <<
+     fink-obsolete-packages,
+     lighttpd-deflate (>= %v-%r)
+  <<
   Files: <<
     lib/%N/mod_compress.so
     share/doc/%N/doc/compress.txt
@@ -237,7 +222,6 @@ SplitOff8: <<
     Enchanced Virtual Host support
   <<
   Files: <<
-    lib/%N/mod_evhost.so
     share/doc/%N/doc/evhost.txt
     etc/%N/mods-available/evhost.conf
   <<
@@ -249,7 +233,6 @@ SplitOff9: <<
   Depends: %N (=%v-%r)
   Description: Plugin: mod_expire
   Files: <<
-    lib/%N/mod_expire.so
     share/doc/%N/doc/expire.txt
     etc/%N/mods-available/expire.conf
   <<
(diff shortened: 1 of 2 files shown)


gstrauss:
@dmacks I am marking this as a draft since I have not tested the PR with fink.  Hopefully, this can provide a starting point to get lighttpd updated in fink, and then possibly backported to earlier 10.x trees.

nieder:
Bringing in @indeyets since it's officially his package.

@gstrauss Thanks for this submission and work towards a much needed update.

Question, comments, and suggestions on your changes:
* for the mods that drop a .so file (such as `lighttpd-access` and `lighttpd-alias`), is the mod still functional? Has the functionality moved into the main part of the server or something else?
* I see a lot of external functionality that is not enabled (such as PgSQL, Nettle, zstd, bzip2, etc) but has library support in Fink. If ignoring those libraries is on purpose, is there an explicit `--disable-FOO` flag that can be added to make that clear? This will also help in case ./configure functionality is ever changed to autodetect.
* For the modules that are eliminated (mod_cml, mod_secdownload, etc), they should be marked as obsolete with `Depends: fink-obsolete-packages`. See SplitOff3 in file crypto/gsasl19-shlibs.info as template

https://github.com/fink/fink-distributions/blob/ddd97e0fbb256081309566cad3612ab82926bbf2/10.9-libcxx/stable/main/finkinfo/crypto/gsasl19-shlibs.info#L146-L157

* Relatedly, can add `Conflicts: lighttpd-FOO (<< 1.4.71)` and `Replaces: lighttpd-FOO (<< 1.4.71)` to the base package so that they get automatically removed upon installation of the new lighttpd release.
* Similarly, if a mod is replacing an obsolete one, the new mod should Conflicts/Replaces the old mod to force the removal of the old SplitOff mod package.
* I'm not sure how the `ConfFiles` field gets dealt with when a package is obsoleted and removed through conflicts/replaces.
* Missing comman in the `Depends:` field for lighttpd-maxminddb.
* Can we upgrade to lua53?
* Can we switch `make install prefix=%i` to `make install DESTDIR=%d`

Lastly, the build currently fails for me on 10.14 with this error:
> libtool: compile:  gcc -DHAVE_CONFIG_H -DHAVE_VERSIONSTAMP_H -DLIBRARY_DIR=\"/sw/lib/lighttpd\" -DSBIN_DIR=\"/sw/sbin\" -I. -I.. -I/sw/include -D_REENTRANT -I/sw/include -I/sw/include -D_TIME_BITS=64 -D_FILE_OFFSET_BITS=64 -D_LARGEFILE_SOURCE -D_LARGE_FILES -D_DEFAULT_SOURCE -g -O2 -pipe -Wall -W -Wshadow -pedantic -MT liblightcomp_la-stat_cache.lo -MD -MP -MF .deps/liblightcomp_la-stat_cache.Tpo -c stat_cache.c  -fno-common -DPIC -o .libs/liblightcomp_la-stat_cache.o
```
stat_cache.c:845:59: error: too few arguments to function call, expected 6, have 4
                                attrval, sizeof(attrval)-1)))
                                                          ^
/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/sys/xattr.h:61:1: note: 'getxattr' declared here
ssize_t getxattr(const char *path, const char *name, void *value, size_t size, u_int32_t position, int options);
^
```

gstrauss:
Thanks for the feedback.

As this package has not been "maintained" in fink over 13 years, I would suggest @indeyets start with https://salsa.debian.org/debian/lighttpd/

I would also suggest that almost nobody is running lighttpd 1.4.25 from fink when lighttpd 1.4.70 (and soon lighttpd 1.4.71) is available from both homebrew and macports.

Therefore, I do not think it worth the effort to run through a deprecation process for things that have already run through multi-year deprecation processes on Debian.

> for the mods that drop a .so file (such as lighttpd-access and lighttpd-alias), is the mod still functional? Has the functionality moved into the main part of the server or something else?

Selected small modules are now built-in to the main lighttpd executable.
The functionality and configuration syntax is otherwise the same.

> I see a lot of external functionality that is not enabled (such as PgSQL, Nettle, zstd, bzip2, etc) but has library support in Fink. If ignoring those libraries is on purpose, is there an explicit --disable-FOO flag that can be added to make that clear? This will also help in case ./configure functionality is ever changed to autodetect.

lighttpd is very flexible and modular.  Many features are optional, so if fink would like to package them as options, I recommend looking at the Debian, homebrew, or macports configs for modern lighttpd pacakges.
e.g. Debian lighttpd package https://salsa.debian.org/debian/lighttpd/

> Missing comman in the Depends: field for lighttpd-maxminddb.

Thanks.  I'll add it.

> Can we upgrade to lua53?

You could.  lighttpd supports lua versions 5.1 and later.  However, if a system already has lua51 installed for other reasons, there is no particular need to upgrade lighttpd to use lua53 instead of lua51.  Does fink support multiple simultaneous installations of different versions of lua?

> Can we switch make install prefix=%i to make install DESTDIR=%d

It is better to `./configure --prefix=...`, and I defer those choices to you and @indeyets who have a fink test environment on which to test.

> Lastly, the build currently fails for me on 10.14 with this error:

Looks like `getxattr()` prototype on Mac OS is different than on Linux.  There are also differences in `attr_get()` prototype.

The quickest solution is to build without `./configure --with-attr`.  Most people probably are not using the feature.   Some Xwin desktop software might have used it in the past.

indeyets:
Thank you for the heads up!
Unfortunately, I don't feel qualified to make any decisions here as I didn't use either lighttpd or fink for many years already.

gstrauss:
@indeyets for modular installations of lighttpd, `./configure --with-nettle` is recommended for providing basic digest and hashing functions to the core lighttpd executable.  For TLS, standalone modules mod_openssl, mod_gnutls, mod_mbedtls, mod_wolfssl, mod_nss are available.  However, since lighttpd 1.4.25 only supported OpenSSL, and since that support was built into lighttpd (and not a separate mod_openssl until much later), then we might as well keep  `./configure --with-openssl`.  It may be beneficial to switch to `./configure --with-nettle --with-openssl` *iff* the fink lighttpd package is also going to support `--with-gnutls` or other additional TLS modules for lighttpd.

gstrauss:
Overall, lighttpd 1.4.25 is so out-of-date, I hope we can focus on the minimal effort to get fink lighttpd package updated to lighttpd 1.4.71.  After upgrading the package to lighttpd 1.4.71, then we may consider adding other features.

For anyone running lighttpd 1.4.25 from fink, we might simply suggest uninstalling lighttpd 1.4.25, installing new package lighttpd 1.4.71, and then reviewing the lighttpd.conf.

Personally, I would estimate that **nobody** *who maintains their system* is running a publicly exposed web server and is running the ancient lighttpd 1.4.25 from fink.  I do not believe any immense amount of effort should be spent attempting a "smooth" transition from lighttpd 1.4.25 to lighttpd 1.4.71.  Any of the esoteric or removed modules are unlikely to be used by any more than a vanishingly small number of private configurations, possibly zero.

TheSin-:
these are all assumptions and broken upgrade paths that no packaging system should ever uses.

- Like debian we want to provide the most options we can.
- we should never assume and break a package, there is no way for us to get to users to uninstall before updating nor should they when we have many ways of dealing with it.
- asking a user to build and install just so we can do it again to add function is not really a good UX

It doesn't take that long to package it and add the C/Rs or Metas to force a new plugin to happen, this should be done prior to update.

Of course that's my 2 cents and just cause no one is uses it doesn't mean it's not a dep of something else, I'm pretty sure a few of my packages have it as a depend which means it is installed and will cause dead locks.

gstrauss:
> these are all assumptions and broken upgrade paths that no packaging system should ever uses.
> 
>     * Like debian we want to provide the most options we can.
> 
>     * we should never assume and break a package, there is no way for us to get to users to uninstall before updating nor should they when we have many ways of dealing with it.
> 
>     * asking a user to build and install just so we can do it again to add function is not really a good UX
> 
> 
> It doesn't take that long to package it and add the C/Rs or Metas to force a new plugin to happen, this should be done prior to update.
> 
> Of course that's my 2 cents and just cause no one is uses it doesn't mean it's not a dep of something else, I'm pretty sure a few of my packages have it as a depend which means it is installed and will cause dead locks.

@TheSin- I do not disagree with your statements.  However, I also do not think my security concerns have been addressed.

fink is providing a software package for the lighttpd _**daemon often run as root**_ (important) that is over 13 years old (lighttpd 1.4.25) and has had a number of CVEs and security fixes in the past 13 years.

The maintainer for the fink lighttpd package has posted above stating that he is no longer maintaining the package or even using fink.

From a security standpoint, this is an extreme failure in processes and procedures.  Perhaps fink should consider sending an email once a year to all listed maintainers, and orphaning packages when the listed maintainer is no longer maintaining the package.  Related, when a package is many versions out-of-date from the latest stable upstream release of the package, this might also be flagged by a script.  (The latest lighttpd release is listed in a machine-parseable file https://download.lighttpd.net/lighttpd/releases-1.4.x/latest.txt)

Given the SECURITY issues which you have been made aware of, if only just now, what are the concrete steps that fink maintainers plan to take to correct this?  What is the security response procedure of fink maintainers?  What packages in fink depend on lighttpd?

I help maintain the Debian packaging info at https://salsa.debian.org/debian/lighttpd/.  I am not familiar with the fink "SplitOff"  syntax so while it might be easy for you to update the fink lighttpd package, it is more of a burden for me to attempt to do so.

I am a lighttpd developer and have put together this PR to help assist fink maintainers.  From a security perspective, I hope that fink maintainers will act with urgency on this, or that fink maintainers should remove the grossly out-of-date lighttpd 1.4.25 package in fink.  I urge anyone using lighttpd on Mac to use the up-to-date and maintained lighttpd packages from homebrew or macports.

I am willing to help update fink, but I am not offering to take on maintainership in fink, and I am not willing to spend many, many hours on fink absent commitment from fink maintainers.  (This obviously does not include the currently listed fink lighttpd maintainer who has posted above that he is not maintaining the fink lighttpd package and is no longer using fink at all.)

While my words may be taken as harsh, that is not the intent.  I am an open source _volunteer_ maintainer, too.  I develop and maintain lighttpd.  I am trying to communicate clearly some obvious shortcomings in fink processes and procedures, and to communicate my offer to help with the lighttpd package, but not to take over the lighttpd package in fink.  Again, I would not recommend that anyone use the fink lighttpd package of lighttpd 1.4.25 in its current state.  Updated lighttpd 1.4.71 packages from homebrew and macports are available, and instructions to build lighttpd from source code on Mac are available at https://redmine.lighttpd.net/projects/lighttpd/wiki/InstallFromSource  (I have personally built and run lighttpd 1.4.71 tests under Mac OS 11.7.)

---

I made some changes requested by @nieder in this PR, and tried to respond to the rest of his feedback.  Is someone from the fink maintainers available to make other changes to SplitOff and test the package?  You should be able to modify this PR, and if not, you are welcome to make pull requests to https://github.com/gstrauss/fink-distributions/tree/lighttpd-1.4.71-1.  Since I do not have a test environment for fink, that may be more efficient than asking me to make spelling changes that I am not able to test as efficiently as you are.  Thank you.

gstrauss:
I made some guesses at syntax and pushed an update.  UNTESTED.

lighttpd mod_deflate is a clean replacement for mod_compress and supports the compress.* directives, although those are deprecated.  This is a clean replacement, so mod_deflate replaces mod_compress and that can be done cleanly with packaging directives to uninstall mod_compress and to install mod_deflate.

For the modules now built-in to lighttpd, it is not a problem if the .so (or .dylib) still exist on disk -- lighttpd does not load them and will otherwise ignore them.  However, removing the *fink* packages will also result in removal of other files in those packages, such as config files.  Also possible to break: since the fink package of lighttpd uses this to set `server.modules = ("mod_foo")`, removing those packages might break existing configurations, again, due to the way that *fink* lighttpd package is scripted.  I am not aware of similar problems with lighttpd packaging in other distros.  Another point: this *is not* how Debian packages lighttpd.

https://redmine.lighttpd.net/projects/lighttpd/wiki/Release-1_4_68 documents lighttpd mod_magnet replacements for mod_evasive, mod_secdownload, mod_uploadprogress, mod_usertrack.  Those lighttpd release notes also document updates to TLS defaults and removal of some ssl.* config directives.

https://redmine.lighttpd.net/projects/lighttpd/wiki/Release-1_4_46
In 2017, _over 5 years ago_, lighttpd 1.4.46 introduced mod_openssl in lighttpd.
If fink lighttpd package chooses to package mod_openssl in a SplitOff, then that package will need to be installed along with lighttpd, or else lighttpd.conf trying to load mod_openssl will fail to start.  How should fink package lighttpd 1.4.71 for transition from (ancient) fink lighttpd package 1.4.25?

gstrauss:
@TheSin- wrote:
> Of course that's my 2 cents and just cause no one is uses it doesn't mean it's not a dep of something else, I'm pretty sure a few of my packages have it as a depend which means it is installed and will cause dead locks.

A `grep` through HEAD of https://github.com/fink/fink-distributions/ suggests that there is a single direct, optional dependency on lighttpd from `javascript-common`: `10.9-libcxx/stable/main/finkinfo/web/javascript-common.info` contains:
```
Suggests: <<
        apache2 (>= 2.4.6) | lighttpd | httpd
<<
```

lighttpd, apache2, and nginx packages are marked with `Provides: httpd`

Among the packages that depend on `httpd`, only `php5` is listed under the `10.9` and `10.7` trees, and `daapd` is listed in some earlier trees.

Are there others package in the fink distribution which depend on lighttpd?

---

Highlighting severely out-of-date versions of web software in fink-distrbutions:

fink-distributions `10.9-libcxx/stable/main/finkinfo/web/lighttpd.info` has lighttpd version 1.4.25, released in Nov 2009, over 13 1/2 years ago!

fink-distributions `10.9-libcxx/stable/main/finkinfo/web/apache2.info` has Apache version 2.4.23, released in Jun 2023, almost 7 years ago!

fink-distributions `10.9-libcxx/stable/main/finkinfo/web/nginx.info` has nginx version 1.14.2, released in Dec 2018, over 4 1/2 years ago!

All of these are extraordinarily out-of-date and not recommended on today's production systems, and so are similarly not very appropriate for a test environment on a Mac.

---

By comparison, macports has:

lighttpd 1.4.71 (released May 2023; current latest release)
https://github.com/macports/macports-ports/blob/master/www/lighttpd/Portfile

Apache 2.4.57 (released Apr 2023; current latest release)
https://github.com/macports/macports-ports/blob/master/www/apache2/Portfile

nginx 1.24.0 (released Apr 2023; current latest release is nginx 1.25.0, released less than two weeks ago)
https://github.com/macports/macports-ports/blob/master/www/nginx/Portfile

TheSin-:
putting in lots of work to figure this out that could be directed fixing  the package.   But if you want to make the argument about the security of the package and the risk it poses.  And you make the point no one is using it anyhow...  If no one is using it then the security isn't an issue and we can take the time...  That being said it isn't much time at all.

I'm not arguing that old dated packages are bad, security wise and should be addressed.  I strongly agree with that, but getting a package setup so it doesn't create an upgrade lock is very important.

At the very least there are 2 points.

- it should not have less options/functionality then the current version.
- it should NOT create an upgrade error at all that required the user to try to figure out.  Upgrade path should never need user interaction, especially if it's a fatal one that requires the user to do something that they don't know about and need to try to figure out.

If a module was renamed, just keep the same splitoff with no files in it (Add one DocFiles for package validation), in the desc add "Transitional", add fink-obsolete and a depends on the new modules that is providing the same functionality.  When I'm done my current stuff I can have a look at this and fix it up, but I can not say when that will be as node is a mess.

gstrauss:
> putting in lots of work to figure this out that could be directed fixing the package

I already did that.  I attempted to make changes based on earlier feedback, updated the PR, and marked it ready for review.

> But if you want to make the argument about the security of the package and the risk it poses. And you make the point no one is using it anyhow... If no one is using it then the security isn't an issue and we can take the time...

As a security professional, I can tell you that part of a security response is impact assessment -- attempting to quickly survey who/what is affected and how, and then prioritizing the response as best as possible.

> but getting a package setup so it doesn't create an upgrade lock is very important.

I do not believe that I have argued against that.  It sounds like a good idea.

Maybe you misunderstood my security perspective.  By example: if you are running a bank IT department and there is a security hole being actively exploited, among the range of options is a simple one: shut down the systems and/or disconnect from the internet.  If there is no time for better actions, it *might* be better to take that action than to take no action at all.

Let's get past that.  I have updated the PR as best I could and am awaiting further feedback.

> At the very least there are 2 points.
> * it should not have less options/functionality then the current version.
> * it should NOT create an upgrade error at all that required the user to try to figure out. Upgrade path should never need user interaction, especially if it's a fatal one that requires the user to do something that they don't know about and need to try to figure out.

I agree that base functionality should remain.  However, for very featureful and *modular* servers such as lighttpd, there may be less-used features that might change over, y'know, 13+ years.  If nothing is ever allowed to change, then someone might think there is no need to update software.  (If someone thinks that their systems are secure if they never update them, then all I can say is that there are mountains of evidence that suggests otherwise.  ... To be clear, you *are not* saying that.)

We should attempt to minimize breakage.  lighttpd deprecated some of these for YEARS already.

Back to impact assessment: if impact assessment suggests that there are few people that would be affected by a breaking change, then some level of user action for transition may be more acceptable than if impact assessment suggests that many people will be affected, in which case more effort should be spent on a seamless transition.

Not all features are of equal importance or of equal impact.

To your other point, I agree that effort should be made to ensure a smooth upgrade and to minimize the chance of users encountering errors.

> If a module was renamed, just keep the same splitoff with no files in it (Add one DocFiles for package validation), in the desc add "Transitional", add fink-obsolete and a depends on the new modules that is providing the same functionality. 

The updates I earlier made to this PR adds `Conflicts` to packages that have been removed, and `fink-obsolete` in other places.  I am not sure where you would like to see "Transitional" in desc instead of where I already added "OBSOLETE:" in desc.

> When I'm done my current stuff I can have a look at this and fix it up, but I can not say when that will be as node is a mess.

I look forward to your feedback.

nieder:
Ugh. For unknown reasons, this ticket stopped sending me notifications a long time ago.
I'm now slowly going through the various mod splitoffs to help them transition cleanly, esp those that have been obsoleted should help point towards their replacement.

@gstrauss 
Q1: mod_deflate replaces mod_compress. What happens if new-lighttpd is installed, but mod_compress is around and not yet replaced by mod_deflate. Not that we want that, but I'm trying to parse possible combinations of what users systems might end up like.
Q2: Is there supposed to be a `maxminddb.conf` ? `mod_maxminddb.so` is being built, but there's no conf for it. Many of the other .conf files are generated via the patch file. Should `maxminddb.conf` be created the same way? What should its contents be?
Something like this?
```
global {
  server.modules += ("mod_maxminddb")

  maxminddb.activate = "enable"
  # DB downloaded from https://dev.maxmind.com/geoip/geoip2/geolite2/#Downloads
  maxminddb.db = "@PREFIX@/share/GeoIP2/database.mmdb"
}

gstrauss:
> Q1: mod_deflate replaces mod_compress. What happens if new-lighttpd is installed, but mod_compress is around and not yet replaced by mod_deflate. Not that we want that, but I'm trying to parse possible combinations of what users systems might end up like.

Modern lighttpd transparently uses mod_deflate for mod_compress, rewriting "mod_compress" to "mod_deflate" in the config, and `compress.*` to the equivalent `deflate.*`  Therefore, I do not understand what you mean by "not yet replaced by mod_deflate".  If you're running lighttpd 1.4.56 or later, "mod_compress" is replaced by "mod_deflate.  If mod_deflate is not installed, that is not a fatal error in lighttpd, and the result is that responses are not compressed; proper (uncompressed) responses are returned instead.

> Q2: Is there supposed to be a maxminddb.conf ? mod_maxminddb.so is being built, but there's no conf for it. Many of the other .conf files are generated via the patch file. Should maxminddb.conf be created the same way? What should its contents be?

Each distro might handle things differently, so please follow fink idioms.

Your suggestion looks fine to me, but you'll need to add `maxminddb.env = (...)` with the desired info for lighttpd mod_maxminddb to do anything.  Please see example at https://wiki.lighttpd.net/mod_maxminddb

gstrauss:
@nieder do you have further questions?  (tagging you in case you did not get notified of my prior post)

nieder:
> Modern lighttpd transparently uses mod_deflate for mod_compress, rewriting "mod_compress" to "mod_deflate" in the config, and compress.* to the equivalent deflate.* Therefore, I do not understand what you mean by "not yet replaced by mod_deflate". If you're running lighttpd 1.4.56 or later, "mod_compress" is replaced by "mod_deflate. If mod_deflate is not installed, that is not a fatal error in lighttpd, and the result is that responses are not compressed; proper (uncompressed) responses are returned instead.

The fink engine can have an older splitoff installed with a newer base package given the right conditions and dependency settings (eg foo-tools-v1.0 with library libfoo-v1.1 assuming binary compatibility). Similarly, I was thinking of the case where the package mod_compress-1.4.25 could still be installed after installing lighttpd-1.4.71. That could technically be allowed by the fink engine, though of course it's not what one wants. However, after posting, I noticed that mod_compress depends on the exact version of lighttpd, and so the above scenario would not be possible, anyway.

Therefore, I need to add Conflicts and Replaces fields to lighttpd such that the obsolete splitoff modules are removed when the base package is installed. Otherwise, the old modules would cause a deadlock because they need the old lighttpd package.

nieder:
I think the current version takes care of the various items discussed. The hardest part was avoiding deadlocks from the obsoleted mods and their replacements when upgrading.

gstrauss:
> However, after posting, I noticed that mod_compress depends on the exact version of lighttpd, and so the above scenario would not be possible, anyway.

Correct.  lighttpd takes precautions and refuses to run with modules compiled against a different version of lighttpd.  Also, as I mentioned above, mod_compress is special-cased and is silently transformed to use mod_deflate, and in the case of mod_deflate not being installed, that is special-cased to not be a fatal error.

> Therefore, I need to add Conflicts and Replaces fields to lighttpd such that the obsolete splitoff modules are removed when the base package is installed. Otherwise, the old modules would cause a deadlock because they need the old lighttpd package.

@nieder Thank you for making the effort to iron this all out.

nieder:
Think this is done. If lighttpd has internal code to block modules from other versions, that clarifies what needs to be done with versioning in the Depends: of the SplitOffs.

dmacks:
Thanks glenn!

PR #1033, opened 2023-06-01
Build gcc with current Homebrew patches and remove lib*san on Ventura

dhomeier:
Building the AddressSanitizer and ThreadSanitizer tools libasan and libubsan is broken on macOS 13.x and has thus been disabled in the latest gcc branches:
https://github.com/iains/gcc-darwin-arm64/commit/e722a1f4
Those files are therefore missing from the shlibs SplitOff, adjusted in a separate .info for `Distribution: 13.0` (for lack of an easier way to handle this for now). Also pulling in the current patches from the gcc arm64 team copied from Homebrew.

Tested on 13.4 arm64 and x86_64/Rosetta, and on 10.14.6 x86_64; should especially be tested on 12.x arm64; if this is an acceptable solution, the gcc updates for 12.3.0 and 13.1.0 can be handled in the same way.


nieder:
Thanks for working this.
Built here on 10.14.6. I don't have any arm64 to test it on, unfortunately.

%r needs to be bumped because there are changes in the build (like all the PIE things) that affect the end result.
Does anything external actually use libasan and libubsan ?

dhomeier:
> %r needs to be bumped because there are changes in the build (like all the PIE things) that affect the end result.

Of course, and 11.4.0-1 is already in...

> Does anything external actually use libasan and libubsan ?

Not aware of any packages, and admittedly I don't know exactly what they are good for. Just thinking some users might have a use for them, so as long as they build we may as well distribute them.
Of course it would simplify packaging a bit to strip them on all platforms or disable the build right away (could modify the patch accordingly if there is no config switch otherwise).

dhomeier:
Rebased this, leaving the libsanitizer build enabled for arm64 – I have tested some of the example functions from 
https://github.com/google/sanitizers/wiki/AddressSanitizer and https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html
on 12.7/arm64 and they seem to be generally functional.

cooljeanius:
I tested this PR on Big Sur with Xcode 13, but got these results:
```
$ fink -m install gcc11
Scanning package description files..........
Information about 15255 packages read in 5 seconds.
Running in Maintainer Mode
Validating package file /opt/sw/fink/dists/stable/main/finkinfo/languages/gcc11.info...
Warning: Field "infodocs" is deprecated. (gcc11.info)
Package looks good!
The package 'gcc11' will be built and installed.
Reading build dependency for gcc11-11.4.0-2...
Can't resolve dependency "xcode (>= 14.2.0)" for package "gcc11-11.4.0-2" (no matching packages/versions found)
Exiting with failure.
```

dhomeier:
> Can't resolve dependency "xcode (>= 14.2.0)" for package "gcc11-11.4.0-2" (no matching packages/versions found)

Right; forgot that Xcode 14 again will not be available at older macOS versions – so that would need another separate build for 12.0; or since the requirement is just for the libsanitizer build on arm64, restrict it to that.
There is no working arm64 distribution for Big Sur or earlier, or is there?

cooljeanius:
> > Can't resolve dependency "xcode (>= 14.2.0)" for package "gcc11-11.4.0-2" (no matching packages/versions found)
> 
> Right; forgot that Xcode 14 again will not be available at older macOS versions – so that would need another separate build for 12.0; or since the requirement is just for the libsanitizer build on arm64, restrict it to that.

OK thanks, I can confirm that [6c81990](https://github.com/fink/fink-distributions/pull/1033/commits/6c81990b2e081619a86c687441c592c4b43988ac) fixed this.

> There is no working arm64 distribution for Big Sur or earlier, or is there?

My machine is x86_64, so I wouldn't know...

cooljeanius (approved):
I managed to build this PR successfully on x86_64 Big Sur with Xcode 13; it's probably good to go.

sth0:
This PR allows build of gcc 11 on MacOS 13.6 (Ventura) X86_64 with Xcode 15.0 and associated CLT.

cooljeanius:
@nieder can you approve?

sth0:
Any progress?  TBH this is blocking the build of nmap and numpy which blocks R-base.  macOS 15 / Intel.

Issue #896, opened 2022-03-05
openmpi has binaries and dev files in same splitoff, which causes upgrade mess


nieder:
Our old openmpi package has %p/bin and %/include in the parent package (which then had to be set to `BDO: false`).
I have latest openmpi packaged, but have split files out to match Debian (-bin and dev files in separate packages).
The dev packages C/R each other as expected, and I added `Replaces: openmpi` to the new -bin (and vice-versa) because of the file collision i %/bin.

However, this leaves the opportunity to have old %p/lib/openmpi/libmpi.dylib and new %p/bin/mpicc installed at the same time and this breaks some things (at least libmpeep24). If the entire bin/dev stack is from the same openmpi version, then build succeed.

What's the best way to do this? I'm not too happy with the current layout having all non-library files in the same same `BDO: false` splitoff, but it would work if we keep that arrangement for new openmpi. But distributing the files more accurately doesn't work without fixing Dependees.

nieder:
Packages that (B)Dep on (openmpi)/opempi-shlibs:
- [ ] arpack-ng
- [ ] boost1.55
- [ ] cran-rmpi-r
- [ ] fftw
- [ ] gromacs-mpi
- [ ] libmeep17
- [ ] libmeep24
- [ ] mpi-oct

If the problem is having old-dev with new-bin installed, I guess we can force new-bin to not just Replaces but also Conflicts on old-dev.

PR #1034, opened 2023-06-18
Openmpi3

nieder:
Update for openmpi to v4.1.4, which is a new libN=3
Update deps as necessary.
The current base package `openmpi` includes both end user binaries and dev-files, which requires setting `BuildDependsOnly: false`. See #896.


nieder:
Not sure if my fix for the `BDO: false` issue is the better way to do it since it might technically allow old openmpi dev library present with new mpicc executable. Maybe instead of using an unversioned openmpi-bin, keep it as openmpi3-bin that will C/R other packages past and future that contain the executables so that packages that need both can be explicit at providing the correct matching pair?

PR #1040, opened 2023-07-08
packaging-py: v23.1

nieder:
@dhomeier update for packaging-py, which is the latest upstream.
This only supports py37+ now, so I pushed off the earlier version to a new .info file.
The new release no longer uses setuptools but requires the boostrap-modules-py package.

Tested on 10.14.6 with py37-py310. I need this update for hatchling-py, which is a new backend for packages that use the pep517 system, that I need for an apipkg-py update that I need for another pkg that I need for...

Also tested hatchling-py and a couple other pymods that use packaging-py and they all still build/pass.


nieder:
@dhomeier ping. Will push soon if no issues since it's not a large update/change.

nieder:
pushing

PR #987, opened 2023-02-02
Update config.guess for libijs1 build on arm64

dhomeier:
Fixes #964; needs yet to be tested with dpkg-1.10

PR #1041, opened 2023-07-16
Ghostscript 10.01.2

PovlAbrahamsen:
Updated Ghostscript plus associated packages:

- libijs1 requires updated config.guess and config.sub to build on Apple silicon. These are copied from automake1.15, which has been added as a build dependency. An alternative approach would be to use the updated source code from the Ghostscript tarball, similar to how lcms2mt is built.
- lcms2mt is updated to the latest upstream code. Even though the version number hasn't been updated upstream, the changes shouldn't affect compatibility. mupdf still builds and appears to work with this version
- Ghostscript is updated to the latest upstream version, 10.01.2. Libtiff has been bumped to libtiff6; otherwise everything, including the patchfile, is the same.

All have been tested on 13.4.1 on Apple silicon, and build with "-m".

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/lcms2mt2.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/lcms2mt2.info
@@ -2,10 +2,10 @@ Info2: <<
 Package: lcms2mt2
 # when updating version, check that mupdf still builds
 Version: 2.12
-Revision: 1
-Type: gsversion (9.54.0)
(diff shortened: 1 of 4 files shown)


PovlAbrahamsen:
Sorry, I hadn't spotted #987, which offers a far neater solution for libijs1! PR has been updated to use this fix.

dmacks:
libijs1 now includes the config.guess fix from #987. We'll wind up rebasing/squashing/cherry-picking or similar when this here PR gets merged.

nieder:
@dmacks : can you resolve the conflicts in the web interface? Seems to only be about `DescPackaging:` changes.

PovlAbrahamsen:
Superseded by newer commits.

PR #1042, opened 2023-07-16
fftw3 - add support for Apple silicon and fix preprocessor

PovlAbrahamsen:
First, a bug fix: 
- rather than invoking Apple's preprocessor directly, preprocess via GCC (by calling "gcc-fsf-11 -E"). If we don't do this, we end up with a bunch of errors when checking header usability.
- Also, header paths are CPPFLAGS, not CFLAGS.

And then we add support for Apple silicon:
- an upstream patch for using the Arm `CNTVCT_EL0` counter on Apple silicon (FFTW/fftw3#267)
- remove the Intel-specific command-line arguments when compiling on Apple silicon
- specify support for the `CNTVCT_EL0` counter. Strictly speaking, this probably isn't necessary, as the first patch should enable this support automatically on MacOS. Alternatively, we could specify this argument and omit the patchfile. Probably comes under the heading of "belt and braces"...

This all requires a working installation of GCC 11. Since I'm now using MacOS 13.4.1 on Apple silicon, I'm using #1033. It compiles and passes all tests with "-m".

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/fftw3.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/fftw3.info
@@ -1,13 +1,15 @@
 Info2: <<
 Package: fftw3
 Version: 3.3.10
-Revision: 2
+Revision: 3
 Maintainer: None <fink-devel@lists.sourceforge.net>
 
 Type: gcc (11)
 
 Source:  https://www.fftw.org/fftw-%v.tar.gz
 Source-Checksum: SHA256(56c932549852cddcfafdab3820b0200c7742675be92179e59e6215b340e26467)
+PatchFile: %n.patch
+PatchFile-MD5: aa44173ecfe7a85cbc78c76ddcff971c
 BuildDepends:  gcc%type_raw[gcc]-compiler
 Depends: %N-shlibs (= %v-%r)
 BuildDependsOnly: True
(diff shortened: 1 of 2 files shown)


nieder:
Seems like a lot of good optimizations are now found (10.14.6):
```
checking whether C compiler accepts -mtune=native... yes
checking whether C compiler accepts -malign-double... yes
checking whether C compiler accepts -fstrict-aliasing... yes
checking whether C compiler accepts -fno-schedule-insns... yes
checking whether C compiler accepts -O3 -fomit-frame-pointer -mtune=native -malign-double -fstrict-aliasing -fno-schedule-insns... yes
```
I can't test arm64.

The old `CPP=which cpp` check was introduced in 51f99bb2f8d30f6c1ecb09deecd220453248b46b. Any idea as to why the generic system cpp might have been chosen? This seems better, but just curious.

Also, do you know why the code is rebuilt in InstallScript before running `make check`?
Why is `--enable-float` added here and not originally?

For the InstallScript `./configure ...`, the two commands in the %m conditional are identical except for the addition of `--enable-sse` in !arm64. That's already included in %c and conditionalized on %m. Can we get rid of the InstallScript conditional and have a single command?

PovlAbrahamsen:
The first run of configure & make & install builds the double-precision libraries. The second runs, in InstallScript, build and install the single-precision libraries. It’s a strange solution, but it works! And two passes are necessary to install both sets of libraries.

The old commit that you refer to specifies the new header paths for 10.9+. But I have no idea why generic CPP was specified; perhaps there was a problem with early builds of gcc? Either way, on 13.4 using generic CPP here produces a bunch of errors in configure which result in options (and possibly optimisations) not being chosen correctly - even though it does build.

nieder:
Grrr. Never got the email that you had responded.
Thanks for the clarifications.

The conditional check in InstallScript to build the single-precision libraries can be removed and just keep the same command as in CompileScript with `--enable-float` appended. `--enable-sse` is already conditionalized in ConfigureParams, so there's no need to and an extra if..then..else.

nieder (approved):
LGTM. The only issue left is the duplicated command in InstallScript, which can be changed after merging.

nieder:
Good catch on the extra `2`.  Leaving as is.

PR #1043, opened 2023-07-20
transition openmotif4 to two_level libxt

nieder:
Based on @PovlAbrahamsen work on #1003 , this moves all packages using openmotif4-2.3.4 and libxt-flat to openmotif4-2.3.8 and two_level libxt.

Openmotif4 couldn't just be updated in place because some packages built against flat libxt/openmotif would fail if then running against the two_level libraries.

So I made a new package `openmotif4-2level` for v2.3.8 linking to libxt, that puts the libraries into a hidden directory so they can coexist with the older `openmotif4` v2.3.4 linking to libxt-flat. The dev and bin packages go straight into %p and C/R with the older openmotif4.

This PR is all packages and 1 library using libxt-flat, now transitioned to the new openmotif4-2level and libxt.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/ddd.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/ddd.info
@@ -1,10 +1,20 @@
 Package: ddd
 Version: 3.3.12
-Revision: 10
+Revision: 12
 GCC: 4.0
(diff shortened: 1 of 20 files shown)


nieder:
No complaints in a month, so merging. Can always revert packages back if something broke.

PR #1044, opened 2023-07-23
gmt6: fix for Apple Silicon

PovlAbrahamsen:
A quick patch to allow GMT6 to build on Apple Silicon - from GenericMappingTools/gmt#4584. Depends on #1042 to have fftw3 working on Apple Silicon.


nieder (approved):
LGTM.
From upstream and should does not have any effect on existing x86_64

PR #1046, opened 2023-08-08
Update pasdoc-gui.info

kamischi:
Change source. Older server may go out of service soon.


nieder (approved):
But not sure why these were all separate pull requests from individual branches.

PR #1051, opened 2023-08-08
Update pascal-p5.info

kamischi:
Change source. Old server may go out of service soon.


nieder (approved):

nieder:
Did not confirm that all necessary files were actually uploaded and checksums were still correct.

PR #1054, opened 2023-08-29
lrzsz - Update - new version compliant with modern compiler.

sth0:
(Apologies if this is not the correct procedure to create a PR)  Still getting used to this.

MacOS 11+ had a bunch of compiler errors and rejected syntax from this very old code (1997?).  Patched and built with fink -m and passed on two systems
MacOS 13.5.1 Apple Silicon
MacOS 10.14.X Intel

Still a bunch of warnings during compile but have not figured out the issue with the MACRO structure to fix them.  Still this is useful utility implementing X,Y,Z MODEM communication which is necessary for some older equipment.


nieder (commented):
Built fine and got rid of a lot of warning for me on 10.14.6 (which is one of your test systems).

If you got anything from elsewhere, can you reference them in DescPackaging to simplify tracking in the future if more updates are necessary?
Optional: would it be too much work to split up the patchfile into a few patch files segregated by fix type (implicit declarations, arm64 fixes, other types of code changes, etc)?

sth0:
It would be a lot of work to track back the patches to different reasons.  I don't think there was much that was arm64 related.
I did update the config.guess file from autoconf but I don't think it made much difference.  That is probably a big chunk of the patch file.  I tried running autoconf on the tarball but it was a complete disaster.

The rest of the patches  were converting
foo(a, b)
int a, b;
{}

to 

foo( int a, int b)
{}

I think that is the last release was December 1998 so the code is well over the drinking age. <https://ohse.de/uwe/software/lrzsz.html>
I see that the author ported it to GitHub about 3 years ago but has not built a release version or tarball.
I did not use any other sources for the update.

nieder (approved):

PR #1057, opened 2023-09-20
lcms2mt: update to latest upstream

PovlAbrahamsen:
An update to the latest upstream code, in sync with the recent update to Ghostscript.

Maintainer is @dmacks


dmacks:
Kinda confusing, still being "2.12" as upstream's actual version. Will look at it.

PovlAbrahamsen:
It looks like some of the upstream code is from an alpha version of 2.13 - so there are some references to this version in some files - but configure.ac still sets the library version to 2.12.

dmacks:
Thanks for figuring out the versioning! I did a slightly different .info/.patch for it.

PR #1070, opened 2023-10-01
Remove read_only_stubs flag

sth0:
Remove read_only_stubs flag
This flag causes problems in Xcode 15 which does not seem to pass it correctly to ld.

gripped read_only_stubs from packages.  Two other packages have it commented out.


nieder:
Thought I had pinged @kamischi since lz4 is his.
Otherwise fine by me.

kamischi:
Go ahead. I am fine with the fix.

nieder (approved):

PR #1083, opened 2023-10-13
netcdf updates

PovlAbrahamsen:
New upstream versions of NetCDF:

- NetCDF-C version 4.9.2 bumps library version to 19. Existing packages already have conflicts set up for netcdf-c19.
- NetCDF-Fortran8 version 4.6.1: use gcc 12, library compatibility version increased.
- NetCDF-CXX4: use new C library; having a file called "VERSION" conflicts with system headers, so we have to make sure that configure doesn't create this file (which doesn't appear to be used anyway)...
- NCO: new upstream version 5.1.8; use new NetCDF library
- ncview: use new NetCDF library

These have been built on MacOS 13.6 on x86_64 with "-m", and appear to work correctly.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/nco.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/nco.info
@@ -1,6 +1,6 @@
 Info2:<<
 Package: nco
-Version: 4.9.5
+Version: 5.1.8
 Revision: 1
 Description: The NetCDF Operators
 License: GPL3
@@ -9,7 +9,7 @@ Maintainer: None <fink-devel@lists.sourceforge.net>
 # Prerequisites
 Depends: <<
 	libgsl25-shlibs,
-	netcdf-c18-shlibs,
+	netcdf-c19-shlibs,
 	udunits2,
 	udunits2-shlibs
 <<
@@ -18,7 +18,7 @@ BuildDepends: <<
 	fink (>= 0.28),
 	fink-package-precedence,
 	libgsl25-dev,
-	netcdf-c18,
+	netcdf-c19,
 	texinfo (>> 4.13),
 	udunits2-dev (>= 2.1.22-1)
 <<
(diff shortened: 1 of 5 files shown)


nieder:
On 10.14.6:
```
libtool: compile:  gcc -DHAVE_CONFIG_H -I. -I.. -I../include -I../include -I../oc2 -I../libnczarr -D_LARGEFILE_SOURCE -I../libsrc4 -I../libncpoco -I/sw/opt/hdf5.v1.12/include -I/sw/include -O2 -fno-strict-aliasing -MT libnczarr_la-zmap_zip.lo -MD -MP -MF .deps/libnczarr_la-zmap_zip.Tpo -c zmap_zip.c  -fno-common -DPIC -o .libs/libnczarr_la-zmap_zip.o
In file included from zmap_zip.c:15:
/sw/include/zip.h:53:10: fatal error: 'zipconf.h' file not found
#include <zipconf.h>
         ^~~~~~~~~~~
1 error generated.
```

The file exists, so for some reason, it's not picking up the correct -I flags to find it.
```
$ dpkg -S zipconf.h
libzip2: /sw/lib/libzip/include/zipconf.h
```

PovlAbrahamsen:
Interesting; I’m pretty sure I built against the system libzip rather than Fink’s. Might need a buildconflict - or a dependence on Fink’s version. I’m afraid I’m away from my computer at the moment, so won’t have a chance to do anything further until next Wednesday at the earliest.

nieder:
```
$ pkg-config --cflags libzip
-I/sw/include -I/sw/lib/libzip/include
```
So ./configure is clearly not using pkg-config to find the flags. I don't know why libzip2 tries to hide one of it's headers in a separate subdirectory. But by installing the newer libzip5 instead, the netcdf-c19 build finished successfully and passed tests with these dependent packages:
```
Found use of headers from 8 fink packages:
	hdf5.200.v1.12
	libblosc1-dev
	libcurl4
	libiconv-dev
	libxml2
	libzip5
	libzstd1-dev
	szip
```

PovlAbrahamsen:
I've now updated the dependencies; hopefully they are correct!

PR #1109, opened 2023-11-21
Libharu new libversion

dmacks:
Update to latest upstream, migrate all reverse-depends  to the latest libversion, and convert older now-unused older libversions to -shlibs-only stubs.


nieder (approved):
Untested, but LGTM.
Just add notes to libharu2.2.1 and libharu2.3.0 that they're shlibs only stubs now.

nieder (approved):

PR #1110, opened 2023-11-22
Libcurl upgrades

dmacks:
Continuating work from "[Fink-users] Update to cURL?"

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/net/libcurl4.info
+++ b/10.9-libcxx/stable/main/finkinfo/net/libcurl4.info
@@ -1,6 +1,6 @@
 Package: libcurl4
 Version: 8.4.0
-Revision: 2
+Revision: 3
 Description: Lib. for transferring files with URL syntax
 DescDetail: <<
 	curl is a command line tool for transferring files with URL syntax,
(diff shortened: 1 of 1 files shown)


sth0:
Haven't managed to send a pull request, but here is the edit I added and it seems to compile and test.

 
 	#Set up to use system's ldap
+	#I think the xcrun will work on systems back to kernel
+	# version 15 (MacOSX 10.11 El Capitan) - sth
 	InclPREFIX="/usr/include"
 	if [ "$(uname -r | cut -d. -f1)" -ge 18 ]; then
-		InclPREFIX="/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include"
+		InclPREFIX="`xcrun --sdk macosx --show-sdk-path`/usr/include"
 	fi

sth0:
Compiles and passes test MacOS 13.6.3-X86 / Xcode 15.2 / CLT 15.1

sth0:
I this ready to merge?  Are there issues?

dmacks:
Upstream has newer versions now. But the main focus here is reducing the spaghetti by using fink's dap. So may as well get this thing merged...

PR #1080, opened 2023-10-11
gcc 12 to 14 updates with libsanitizer and Distribution=13.0, 14.0, 15.0 variants

dhomeier:
Counterpart of #1033 for gcc versions 12.3 and 13.2, with patches from the gcc arm64 team copied from Homebrew.
Adding the same experimental build of AddressSanitizer and ThreadSanitizer tools as for gcc11 – however libasan in particular is untested with gcc13.

Tested on 13.4 and 12.7 arm64 and x86_64/Rosetta, and on 10.14.6 x86_64; still needs testing on latest 13.x arm64 with Xcode 15.

gcc12
PASS 10.14.6/x86_64 @dhomeier @nieder
PASS 13.7/x86_64 @nieder 
FAIL/PASS 15.4/x86_64 with `with-build-time-tools` [comment](https://github.com/fink/fink-distributions/pull/1080#issuecomment-2771937824) @PovlAbrahamsen
PASS arm64 @dhomeier 

gcc13
PASS 10.14.6/x86_64 @dhomeier @nieder
PASS arm64 @dhomeir

gcc14
FAIL 10.13 [comment](https://github.com/fink/fink-distributions/pull/1080#issuecomment-3149892645) @dmacks 
PASS 10.14.6/x86_64 @dhomeier @nieder
FAIL/PASS 15.4/x86_64 with `with-build-time-tools` [comment](https://github.com/fink/fink-distributions/pull/1080#issuecomment-2771937824) @PovlAbrahamsen 
PASS arm64 @dhomeir

PR #1113, opened 2024-01-04
R-base 4.1: Patch ATSFont calls to CGFont calls, update gcc11 to gcc12

sth0:
R-base4.1 does not compile on MacOS 13.  This OS release removed the deprecated ATSFont calls for the newer CGFont calls.  This replaces the calls with the corresponding new API.  Also updated the link to gcc12-shlibs from gcc11-shlibs.


nieder:
R packages that use gccXX hardcode the version that R-base was built against. Which is why you see lines like this:

https://github.com/fink/fink-distributions/blob/60fcd1badf9c742f90fe5064bfb2961865653818/10.9-libcxx/stable/main/finkinfo/libs/rmods/cran-tidyr-r.info#L57

If we update r-base41 to gcc12, then we have update all relevant cran-XXX-r41 packages to use gcc12 (apart from gccXX availability issues across macOS versions).

sth0:
Will regress to gcc11-shlibs

sth0:
I have been using the PR #1080 <https://github.com/fink/fink-distributions/pull/1080> which seems to allow building of gcc11 / 12 / 13 on all the systems I have tried.

sth0:
pushed commits to regress to gcc11 compiler and bumped revision number.

dmacks:
This patch looks like upstream's https://bugs.r-project.org/show_bug.cgi?id=18426 so we should cite it as such in the .info (might be cleaner to have that attachment literally as a PatchFile2 instead of merged into the main PatchFile).

sth0:
I had trouble applying it as a separate patch and thus combined it.  I can try to separate them out as separate patches if the consensus is that is "cleaner".

dmacks:
Doesn't really matter whether it is a seperate .patch or not, whatever is easier. But if that's the origin of the patch, definitely need to document that fact in the .info.

sth0:
Will update the implementation description.

sth0:
Added comment to BuildDesc tag and removed references to PCRE1 which is not used for R version 4.1

PR #1121, opened 2024-01-30
Pytest 7

nieder:
Some pymods are starting to need pytest >=7. Importantly, the update here activates it's own tests w/out needing to download tons of other pymods through tox/pip.
Tested py38/p310 on 10.14.5


dmacks:
The new-version pytest-pyXX is for 3.7-3.10 but it has a dependency on exceptiongroup-pyXX and some others that only exists for 3.8-3.10. So pytest-py37 has unsatisfiable deps. The exceptiongroup-pyXX is especially problematic, as exceptiongroup-py.info has a note " Do not provide above 3.10 or below 3.8".

nieder:
Should be fixed by f62be80c0021e3737f877ab3c62af348b6389a3c. Don't know why I had the minimum version restricted before.

PR #1123, opened 2024-02-06
snowballstemmer-py: v2.2.0

nieder:
new version required for newer sphinx-py
tested py27/35/38/310 on 10.14.6


nieder:
Merging after a week. Simple version bump and sphinx is the only package that uses this.

PR #1124, opened 2024-02-10
Bleach py update

nieder:
Updates bleach-py to the latest upstreams. The package itself has been loosely deprecated by upstream, but we might as well have the latest versions. There's a new similar binding called `nh3`, but it requires `ammonia`, which requires rust.


nieder:
No response in 3 months. Will merge this week.

PR #1128, opened 2024-02-22
lighttpd: update to lighttpd 1.4.74

gstrauss:
lighttpd: update to lighttpd 1.4.74

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
@@ -1,5 +1,5 @@
 Package: lighttpd
-Version: 1.4.73
+Version: 1.4.74
 Revision: 1
 DocFiles: AUTHORS COPYING NEWS README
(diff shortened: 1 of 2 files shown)


gstrauss:
(might need minor patch to find ../compat/sys/queue.h)

dmacks:
With the patch, builds fine on my 10.13 (once I updated the Patch-MD5 for the new .patch). I see it's using lua51, which is pretty old. The configure script seems to accept up to lua54, and it builds equivalently for me if I update the deps for that.

gstrauss:
Thanks.  I updated the MD5 for lighttpd.patch and pushed the change to this PR.

FYI: lighttpd mod_magnet supports lua 5.1 and later, including 5.2, 5.3, and 5.4.  Also supported is lua-jit, which is based on lua 5.1.

fink maintainers may safely choose to modify lighttpd.info to use lua54.

gstrauss:
Replaced by #1130 for lighttpd 1.4.75

PR #1136, opened 2024-06-13
Rustc update to 1.78.0

nieder:
bump to 1.78.0
Package includes cargo, which used to be a separate package, but cargo 0.79.0 is included as part of 1.78.0 dist
I was able to build a couple sample .rs files with this, and cargo was able to download some crates and use them to compile another thing.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/languages/rustc.info
+++ b/10.9-libcxx/stable/main/finkinfo/languages/rustc.info
@@ -1,6 +1,7 @@
 Package: rustc
-Version: 1.30.0
+Version: 1.78.0
 Revision: 1
+Architecture: x86_64
 Description: Safe Fast System Programming Language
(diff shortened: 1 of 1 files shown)


nieder:
So the dylibs created by rustc include a hash in the filename that's unique to the compiler settings and host. Therefore, doesn't seem to generate a reproducible build. Ugh.

nieder:
I take that back. Tried building on 10.14.6 and 12.7, and in both cases the hash on the dylibs was the same. So maybe it's just dependent on build settings.

dmacks:
It would be great to get rust on apple-silicon. There are no "arm64-apple-darwin" bootstrap files on their server (based on %m), but there are "aarch64-apple-darwin" files. I guess we need separate .info or a Type: token for the arch string it ants  conditional on %m rather than using %m directly.

PR #1139, opened 2024-06-30
Update astroid info for python versions 3.9 and 3.10

sth0:
Simple addition of 3.9 and 3.10 to python versions for astroid.  I have a package that I haven't pushed an update for years but have a bunch of these updates in my local tree.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/astroid-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/astroid-py.info
@@ -1,7 +1,7 @@
 Info3: <<
 
 Package: astroid-py%type_pkg[python]
-Type: python (3.4 3.5 3.6 3.7 3.8)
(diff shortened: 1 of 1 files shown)


nieder:
Needs #1140

nieder:
Also need to update lazy-object-proxy-py to have py310 variant
```
WARNING: While resolving dependency "lazy-object-proxy-py310" for package "astroid-py310-2.2.5-3", package "lazy-object-proxy-py310" was not found.
Can't resolve dependency "lazy-object-proxy-py310" for package "astroid-py310-2.2.5-3" (no matching packages/versions found)
Exiting with failure.
```

The lazy-object-proxy-py package might benefit from the newer releases than what we have (although 1.10.0 needs newer setuptools than what we can currently have).

sth0:
see PR #1144, for update to lazy-object-proxy.
(I have a lot of these simple updates!)

nieder (approved):
lgtm

PR #1140, opened 2024-06-30
Update python versions for logilab-common to include 3.9 and 3.10

sth0:
Simple change again just adding python 3.9 and 3.10 to package.


nieder (approved):
LGTM for adding variants.
@schwehr if he wants to consider updating to latest upstream

PR #1141, opened 2024-06-30
Update pylint for python 3.9 and 3.10

sth0:
Added python 3.9 and 3.10 to the info file as options


nieder:
Needs #1139, which needs #1140

nieder (approved):
lgtm.

I'm going to upgrade pylint next to latest upstream 3.2.5 which is for py38-py310

PR #1142, opened 2024-06-30
Update pycrypto  info for python 3.9 and 3.10

sth0:
Update info file to include python 3.9 and 3.10


nieder:
Fixed in eeca8b9e8e093bea754abcf5508c7bc54d9a7434
Also had to fix test behavior since this fails w/ py38+. Is one of your other submissions needing pycrypto? It's very old, unmaintained, and insecure.

Currently, these Fink packages use it:
- [ ] keyring-py
- [ ] libcloud-py
- [ ] ortfileconverter-py
- [ ] python-potr-py
- [ ] tlslite-ng-py
- [ ] python-tlslite-py
- [ ] bzr-py
- [ ] duplicity-py
- [ ] google-api-python-client-py

sth0:
Yes, I am desperately trying to get duplicity-py updated from version 1.X to 3.0 and python 2.7 to python 3.10.  Possibly the pycrypto dependency is gone.  I will check with the duplicity maintainers.

PR #1143, opened 2024-06-30
First release of s3transfer python module package version 0.10.2

sth0:
New package for the s3transfer python module for automated transfers to S3 repositories.  Builds on 14.5 and 13.6


nieder (changes requested):
Use `https://files.pythonhosted.org/packages/source/s/s3transfer/s3transfer-%v.tar.gz` as the source URL since that avoids having to change when the hash changes for every single release.

sth0:
Updated per suggestion

nieder (approved):
Builds fine for me, but Homepage field is 404. Will fix after merging

PR #1144, opened 2024-07-03
Update lazy-object-proxy for python 3.9 and 3.10

sth0:
Simple update to lazy-object-proxy to allow astroid-py310 to build.  Hopefully this is the last dependancy for pylint update.
Have added this PR as a depends for pull request #1139
Did not update version of lazy-object-proxy since was pointed out that needs updated setuptools which will take extensive testing.


nieder (approved):
LGTM.

regarding updates, according to https://github.com/ionelmc/python-lazy-object-proxy/blob/v1.9.0/pyproject.toml we have the necessary dependencies to update to 1.9.0 for py37-py310

PR #1146, opened 2024-07-03
Updated Monotonic to include Python 3.9 and 3.10

sth0:
Added python 3.9 and 3.10 to the available python versions for monotonic


nieder (changes requested):
Homepage field points to wrong package.
Any chance you can update to v1.6 (from 2021 vs 1.5 from 2018)?

sth0:
Will adjust package.  Not sure if the update is possible without breaking a lot of things, there was an issue with some utilities.  Didn't check the homepage, since I didn't modify it, will correct.

sth0:
Updated per suggestions

nieder (approved):
LGTM.
FYI, upstream has archived this project, so 1.6 looks to be the final release.

PR #1147, opened 2024-07-03
requests-oauthlib added support for python 3.9 and 3.10

sth0:
Just added support for python 3.9 and 3.10, working through the dependencies slowly


nieder (approved):
OK, after locally updating (and pushing) requests-mock-py

PR #1148, opened 2024-07-04
fasteners-py updated for python 3.9 and 3.10

sth0:
Again just added python 3.9 and 3.10 and bumped build version.  Cannot update to latest since uses different build system and there is no setup.py so I am lost on how to build it.  Depends on PR #1146

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/fasteners-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/fasteners-py.info
@@ -1,7 +1,7 @@
 Info2: <<
 Package: fasteners-py%type_pkg[python]
 Version: 0.15
-Revision: 6
+Revision: 7
 Distribution: <<
 	(%type_pkg[python] = 34 ) 10.9,
 	(%type_pkg[python] = 34 ) 10.10,
@@ -29,7 +29,7 @@ Distribution: <<
 	(%type_pkg[python] = 36 ) 10.15
 <<
 
-Type: python (2.7 3.4 3.5 3.6 3.7 3.8)
+Type: python (2.7 3.4 3.5 3.6 3.7 3.8 3.9 3.10)
 
 Description: Python package providing useful locks
 License: BSD


nieder:
Releases through 0.17 still use setup.py: https://github.com/harlowja/fasteners/tree/0.17

For projects using pyproject.toml (PEP517 build system), look at alabaster-py.info as an example. You'll need to make sure bootstrap-modules-py%type_pkg[python] is a build dependency, and the pyproject.toml file in the source will indicate what build backend it uses (setuptools, flit, hatchling, etc). Setup.cfg or requirements.txt or similar often indicate what other (build)dependencies are needed.

The CompileScript and InstallScript in alabaster-py.info are the bare minimum needed for building the wheel (like a pre-compiled archive of the build), and then installing the wheel using pyproject.toml.

sth0:
Following your suggestion, I have a package for fasteners that builds version 0.19 for python 3.9 and 3.10.  The older versions of python are left at version 1.5.  I used the hashed version of the source, since that matches the SHA256 hash given by Pipy.  The recommended non-hashed path for source gives a different SHA256 hash which worries me.  What is different in the two tarballs?

nieder:
The non-hashed URL is `https://files.pythonhosted.org/packages/source/f/fasteners/fasteners-%v.tar.gz` . Note the extra directory between `/f/` and tarball name. The `e3b0c442...` sha256 is that for an empty file. When using the full fixed path, I get `b4f37c3ac52d8a445af3a66bce57b33b5e90b97c696b7b984f530cf8f0ded09c`, which matches what you got from the direct download and the PyPI checksum

sth0:
I stated at that URL too long...  Fixed and pushed an update.

sth0:
Pushed with updated versions and distributions so python 3.6 and below use the old 0.15 and 3.7 and above use the 0.19 versions.

nieder (approved):
Works except for 2 failing tests using eventlets (which we don't have). Since eventlets upstream discourages people from using it any more, I'll disable those two tests after merging.

PR #935, opened 2022-12-26
Update and fix docutils-py

dhomeier:
Fixes #931 using the latest available upstream versions for Python up to 3.10 and < 3.7, respectively.
I have not had a recent system with the older Python versions available to test the 1.18.1 package on.
The point of failure seems to have been adding `docutils/utils` to the local `PYTHONPATH`. This overshadowed system modules like `math`, and I don't see it required – the local utils versions are already imported with the their relative path, e.g. as `from docutils.utils import math`.

PR #1150, opened 2024-07-05
diskcache-py initial package

sth0:
diskcache-py is needed to test new version of fasteners.  This version will install.
However this will NOT run the tests, since it needs django to be installed, which needs docutils to be installed.  docutils is horribly out of date and  will not build with python 3.10 (or maybe much earlier), the current installed version needs eventlet-py which is very deprecated and will not build for python 3.  I would rather not try to update an infinite number of packages.  And leave docutils-py to a later date and someone more competent.  With diskcache installed, fasteners will pass all tests.  ( see PR #1148 )

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/diskcache-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/diskcache-py.info
@@ -1,0 +1,44 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info4: <<
+Package: diskcache-py%type_pkg[python]
+Version: 5.6.3
+Revision: 1
+Type: python (3.5 3.6 3.7 3.8 3.9 3.10)
+Description: Python package for disk & file cache library
+License: BSD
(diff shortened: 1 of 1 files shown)


nieder:
Docutils update is at #935

nieder:
Same issue with source URL as in #1148: needs the extra directory in path to get the tarball.
If the tests are currently known to be failing (not due to the package itself), just comment out the entire InfoTest field. This lets people who normally build with `-m` succeed, while keeping the field as a reminder.
Rest of package looks OK.

sth0:
> Docutils update is at #935

Reading through that thread, it is clear why I didn't get involved in docutils, it seemed a bit more complex than I wanted to dive into.  In addition that thread stalled out about a year ago and is not ready for prime time.

nieder:
boostrap-modules-py is only for py36 and up. This has py35, which will therefore fail.

sth0:
Removed python 3.5 build.  Should now be just 3.6 to 3.10

nieder (approved):

PR #1152, opened 2024-07-06
boto3-py new package to replace/update boto-py

sth0:
Older boto-py has been split into two utilities.  boto3-py depends on botocore-py ( PR #1151 ).  The boto3-py codes has no testing phase.


nieder (changes requested):
No py311 just yet.

nieder:
The older boto-py had a bunch of utilities in %p/bin. These are now gone from botocore and boto3. Is there another package that has them or are they deprecated. Just curiosity, doesn't affect getting this in.

sth0:
Removed ref to python 3.11 per suggestion.  I do not see any command line boto utilities in this package.  The old boto should "install along side" according to the documentation but that API is deprecated.  I can see where the boto utilities are useful but I don't see a way to install them.

nieder (approved):

PR #1153, opened 2024-07-06
New package jmespath-py to parse JSON coded text.

sth0:
The jmespath-py package is a required package for boto3-py package (PR #1152)


nieder (approved):
LGTM.
All tests also pass with `%p/bin/python%type_raw[python] setup.py test || exit 2` in TestScript.

PR #1155, opened 2024-07-13
Mock update to version 5.1.0

sth0:
Update mock to version 5.1.0, rename old version for python < 3.4 with mock current version 3.0.5
mock now supports python 3.10
passes fink -m tests on MacOS 14.5-Sonoma / Xcode 15.4 (command line tools 15.3)
Have emailed maintainer (Charles Lepple) who gave permission to update as I see fit.  Did original update to 3.0.5 to support python 3.10 in 2019.  Updated to 5.1.0 2024.


nieder (changes requested):
py36 is dist restricted to 10.9 -> 10.15, but that got removed in the updated .info. Besides that pkgs look like they built and work OK.

sth0:
removed python 3.11 reference and moved python 3.6 to mock version 3.0.5 build.  This should setup all versions correctly.

nieder (approved):

PR #1157, opened 2024-08-23
duplicity - new upstream update to 3.0.0

sth0:
Update of duplicity encrypted backup software from version 0.7.18.1 to version 3.0.0.  Passes all tests.  Builds with python 3.10 only.  Working on updating newest release 3.0.2 but that fails several tests and is recommended to update python 3.10 and pytest to version 8.3.2 which will delay getting this update out and working.

The documentation is on the web and so not built by the package.  This would also require several new and updated modules and is not necessary.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/duplicity.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/duplicity.info
@@ -1,68 +1,77 @@
 Package: duplicity
-Version: 0.7.18.1
-Revision: 3
+Version: 3.0.0
+Revision: 1
 Description: Encrypted backup using rsync algorithm
 License: GPL
 Homepage: http://duplicity.nongnu.org/
-Maintainer:  Scott Hannahs <shannahs@users.sourceforge.net>
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
 
 # Dependencies.
 BuildDepends: <<
 	fink (>= 0.24.12-1),
-	librsync (>= 0.9.7-1002)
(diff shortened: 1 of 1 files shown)


nieder:
* What's the purpose for needing gcc-12 vs system-clang?
* With the use of `PatchFile` removed, you can also remove the physical duplicity.patch file.
* Is tox-py310 actually needed or left over? There's currently no tox-py310 in fink-dists. Looking at the code, I only see using tox to then run pytest, but here we run pytest directly (which is preferred anyway). Is there another use I'm missing here?
* I get this error with using `-build-base=%i` during compile script:
```
/sw/bin/python3.10 setup.py build  --librsync-dir=/sw --build-base=/sw/build.build/root-duplicity-3.0.0-1/sw
running build
running build_py
creating /sw/build.build/root-duplicity-3.0.0-1
error: could not create '/sw/build.build/root-duplicity-3.0.0-1': Permission denied
```

Getting rid of tox-py310 and the `--build-base` flag, all tests pass, so packaging just needs to be cleaned up.

sth0:
- no preference to system-clang vs gcc-12, just left over from previous version
- My oversight, didn't update the distribution for git
- There was a tox.ini file so I assumed that tox was needed.  I have a patch for tox-py310, I might as well submit it in separate PR.  Not missing anything, just my misread of the source.
- Had build base=%i since the deb analysis was claiming that files were installed outside of the root-xxx directory.  Will check again, might have been fixed in upstream since I started this update.

Update pushed.

PR #1158, opened 2024-08-23
update libiconv and gettext to latest upstream for building on macOS15

nieder:
Updates libiconv and libgettext/gettext-tools to their latest upstreams. Includes some fixes that are needed for soon to be released macOS15.

Taken from fink/fink#267, which is for bootstrapping, and this is for the installed tree.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/base/dpkg.info
+++ b/10.9-libcxx/stable/main/finkinfo/base/dpkg.info
@@ -1,11 +1,11 @@
 Package: dpkg
 Version: 1.10.21
-Revision: 1251
+Revision: 1252
 GCC: 4.0
 BuildDepends: fink (>= 0.30.0)
 Depends: <<
 	gzip,
-	libgettext8-shlibs (>= 0.20.2-1),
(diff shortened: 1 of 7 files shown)


nieder:
Will merge Thursday if no comments

dmacks:
Seems fine by eye, except gettext-tools:CompileScript: a stream of pushd without popd seems odd.

nieder:
Fixed the pushd stream. Also worked around some known harmless test fails.

nieder:
@dmacks ping. Apparently my merging is blocked w/out a review

dmacks:
on it...

dmacks:
Builds fine on 10.13. Would be nice to freshen up the .patch so they don't warn about offset and fuzz.

nieder:
ugh. forgot that dpkg needs to keep up with gettext and updating it currently fails finding `../../gettext-runtime/intl/libintl.la`

nieder:
patches refreshed, and dpkg updated since it needs to use the refreshed patches

dmacks (approved):

PR #1163, opened 2024-09-22
duplicity - Upstream update to V3.0.2

sth0:
Upstream bug fix update, also changed pytest to reliably and repeatedly pass all tests.


nieder (approved):
LGTM. All tests passed on 10.14

PR #1165, opened 2024-10-31
Update to compface face to patch implicit definition as int

sth0:
Newer compiler choked on implicit definition.  Now is explicitly "int".  Builds under OS 15.1 and installs.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/graphics/compface.info
+++ b/10.9-libcxx/stable/main/finkinfo/graphics/compface.info
@@ -3,16 +3,16 @@ Version: 1.5.2
 Revision: 3
 Source: http://ftp.xemacs.org/pub/xemacs/aux/%n-%v.tar.gz
 Source-Checksum: SHA256(a6998245f530217b800f33e01656be8d1f0445632295afa100e5c1611e4f6825)
-Source2: mirror:debian:pool/main/libc/libcompface/libcompface_1.5.2-5.diff.gz
+Source2: mirror:debian:pool/main/libc/libcompface/libcompface_%v-5.diff.gz
 Source2-Checksum: SHA256(0587f531d09aa229618e4f648ca085a816a8d35cb4d35e216446c7462ffef733)
 PatchFile: %n.patch
-PatchFile-Checksum: SHA256(32305e958a4ef1303369852eec1a5cab3eb7ed07e795dd9741ec4358abef933a)
+PatchFile-Checksum: SHA256(2c32e97dc4b8064083a35f3fd09391030310660b457ae850a4f41e1cf8b4ecd4)
 PatchFile2: %n-implicit-declarations.patch
-PatchFile2-MD5: 6ece31839e166a829152bbb1644e6a69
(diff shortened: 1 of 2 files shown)


TheSin-:
in this one you change the version from 1.4 to 1.5 won't that change the contents of the deb?  isn't that a var that is injected into the code?

sth0:
Actually I couldn't find where that variable was actually used, but thought it should match the Major/Minor version of the code.  Not sure why it was set.  It doesn't affect the naming of the dylibs.  Will it affect the deb?  Don't know.

TheSin-:
might affect headers or other files, I don't know the build at all so I can't say but maybe @nieder will know more?

nieder:
Near as I can tell VERSION in Makefile is only used to set the current_version and compat_version of the dylib as part of our own patch for linking. This patch to file.c is not changing anything about the API/ABI of the library, since it's just making explicit was previously implicit (or failing in newer Xcode) and therefore no changes to the deb.
IFF indeed something was changing in the library ABI and we needed to bump current_version, then we would also need to increase our package `Revision`.

sth0:
So we are good?  I will leave the Revision number unchanged.

nieder:
No, we don't want to increase current_version or compat_version in the dylib for the implicit-int fix.

sth0:
Ok, but this is version 1.5.2 and so shouldn't that compat_version be 1.5 ?  I guess I am confused between the source version and the library version.  I was thinking that having the Version at 1.4 for a 1.5 distribution of the code was a problem in the previous release of the package.

nieder:
compatibility_version or current_version for a library do not need to follow the software version. This is an example where the two don't track. But it is possible to do the same exact software changes, yet have current_version track the software version.

1. software version=1, compat_version of libFOO.1.dylib=1, current_version of dylib=1
2. add new features to library which don't break previous functions: software version=1.1, compat_v=1, current_v=1.1
3. major addition of new features, but old functions of library are still valid: software version=2, compat_v=1, current_v=2
4. super massive change that breaks compatibility with previous releases (eg functions removed). This requires a new library instalL_name from libFOO.1.dylib to libFOO.2.dylib. software version=3, compat_v=1, current_v=1. compat_v and current_v _can_ (but don't have to) be brought back down to 1 because libFOO.2.dylib is an entirely new library as far as the linker is concerned.

TL;DR : compat_version and/or current_version don't need to track the software version.

Looking at the history of compface, it looks like our original patch for 10.2 added compat_v matching software version because upstream didn't know about OS X, so 1.4 was as good a number as any. However, when switching to version=1.5, compat_v didn't get bumped because the library was still compatible. current_version probably should have been bumped, however, assuming new functions were added between 1.4 and 1.5.

sth0:
Thanks for the info, I had an idea that there was a compatibility version but not sure where that was defined/stored.  I think the "current version" is just a weird concept and not sure of its use.  The software version breaking compatibility when there is a major change and changing the library numbering make sense.

So here is the latest push on the PR with the VERSION reset back to 1.4 and the dot-info file cleaned up a bit.  i.e. use of SHA256 checksums and %v substitutions.

nieder (approved):

PR #1166, opened 2024-10-31
Update locale-gettext for MacOS 15 / perl 5.34.1

sth0:
Added perl version 5.34.1 as possible version for system perl in macOS 15.X

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/locale-gettext-pm.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/locale-gettext-pm.info
@@ -16,7 +16,7 @@ Replaces: %N-doc
 Source: mirror:cpan:authors/id/P/PV/PVANDRY/gettext-%v.tar.gz
 Source-Checksum: SHA256(909d47954697e7c04218f972915b787bd1244d75e3bd01620bc167d5bbc49c15)
 SourceRename: Locale-gettext-%v.tar.gz
-Type: perl (5.16.2 5.18.2 5.18.4 5.28.2 5.30.2 5.30.3)
+Type: perl (5.16.2 5.18.2 5.18.4 5.28.2 5.30.2 5.30.3 5.34.1)
 Distribution: <<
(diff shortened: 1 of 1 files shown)


nieder:
Ugh. i guess we need a standalone perl5341 package for non sequoia (16.x).

sth0:
Or I can distribution restrict it to 15.X ?  Double ugh?

nieder:
For future perlmod adds, you can keep them in a single PR, especially if they're all heading towards providing Deps for a final package. Otherwise we'll end up with 500+ pull requests.

dhomeier:
@nieder, I have ~272 5.34.1 perlmods ready to be pushed; should they still go into individual commits, or can that also be done en bloc (could probably really be one semi-automatically)?

TheSin-:
I'd say it depends on the block, all the ones where you just add a variants and that's all can be one.  Anything that is more involved and might need testing should be separate IMHO.  You don't want to block a bunch cause of one complex change.

TheSin-:
but you have some some don't sit on them, I'm going down this road right now and I'm just wasting my time if you have it all done already.

TheSin-:
@dhomeier you don't have bash for 15.x do you?  Save me that mess?

sth0:
I bundled a bunch of perlmods in the past and was asked to separate it out.  I can go either way, but I will bundle them in the future.

TheSin-:
like I said in my comment tit depends on the bundle.  If it's just added a variant then bundle them, if it's a big change/patching etc etc then separate them.

dhomeier:
> but you have some some don't sit on them, I'm going down this road right now and I'm just wasting my time if you have it all done already.

Will do. Trying to test build them at least on 15.1 here; compiling 500-600 packages is gonna take some time...

TheSin-:
yeah I did 100 today but all just updated I didn't do before I updated ;)  Going to wait on perl though cause I don't want to duplicate efforts.

dhomeier:
Now awaiting review in #1181

PR #1167, opened 2024-10-31
Updated help2man for perl 5.34.1 and macOS 15.0

sth0:
Simple update of file, builds and installs on macOS 15.1, shouldn't change anything for other OS and perl versions.


TheSin-:
not sure that needed a rev up honestly.  Since it won't affect old installs

sth0:
I agree, I updated it in my local tree during testing and I wanted to make sure it was using the local version and not the distribution version.  If it makes life easier, I  will remove that edit.

In a general question, when should that version number be bumped?  Only if it requires a rebuild for existing installations of he same version?

TheSin-:
if the contents of the deb file changes.  Don't want to make everyone rebuild it if nothing has changed.

TheSin- (approved):

PR #1168, opened 2024-10-31
Update to glib2-shlibs to compile under CLT 16.1 / macOS 15.X

sth0:
Added cast to strcmp function, line 2361 of gdesktopappinfo.c needed (int (*)(const void *, const void *)) before strcmp for consistency.

I applied existing patch file to source, edited gdesktopappinfo.c and created new patch file, but there seem to be more differences between the patch files than the one edit.
(getting closer to building ghostscript)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/gnome/glib2-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/gnome/glib2-shlibs.info
@@ -25,7 +25,7 @@ Replaces: glib2 (<< 2.12.0-1)
 Source: mirror:gnome:sources/glib/2.22/glib-%v.tar.bz2
 Source-Checksum: SHA256(0d1f6bf8aaeab772f2370229eefda45bef434e3f29a7d1d8e5dfafeaa1d8ad14)
 PatchFile: %n.patch
-PatchFile-MD5: 52b15fd8c1d1b57d7404370adbdcb8fd
+PatchFile-Checksum: SHA256(4d8136cd557884684ca81e6164f29966a8b2be79281373c37acec95d441582bf)
 PatchScript: <<
 	sed 's,@PREFIX@,%p,g' < %{PatchFile} | patch -p1
 
--- a/10.9-libcxx/stable/main/finkinfo/gnome/glib2-shlibs.patch
+++ b/10.9-libcxx/stable/main/finkinfo/gnome/glib2-shlibs.patch
@@ -1,6 +1,6 @@
-diff -Nurd -x'*~' glib-2.22.4.orig/configure glib-2.22.4/configure
---- glib-2.22.4.orig/configure	2010-01-06 19:48:41.000000000 -0500
-+++ glib-2.22.4/configure	2023-02-24 00:46:35.000000000 -0500
+diff -Nurd -x*~ glib-2.22.4-orig/configure glib-2.22.4/configure
+--- glib-2.22.4-orig/configure	2010-01-06 19:48:41
++++ glib-2.22.4/configure	2024-10-31 17:17:47
 @@ -4801,7 +4801,7 @@
  #define HAVE_CARBON 1
  _ACEOF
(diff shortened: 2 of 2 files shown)


nieder:
I looked at what's going on with the patch, and the generated directories after patching are identical except for the fixing of strcmp. Only thing I can see is that the patch comes out different when diff is `/usr/bin/diff` vs `/sw/bin/diff` and the funkyness of diffing two patch files (esp seen in glib/gunicollate.c) is due where system-diff vs fink-diff decide to put unchanged lines between changed stretches.

Here's the current fink patch done with system-diff. The blank like after `gsize i;` got deleted as part of that chunk.
```diff
@@ -210,28 +210,31 @@
 {
   gchar *result;
   gsize result_len;
-  gsize i;
-
-  /* Pretty smart algorithm here: ignore first eight bytes of the
-   * collation key. It doesn't produce results equivalent to
-   * UCCompareCollationKeys's, but the difference seems to be only
-   * that UCCompareCollationKeys in some cases produces 0 where our
-   * comparison gets -1 or 1. */
-
-  if (key_len * sizeof (UCCollationValue) <= 8)
-    return g_strdup ("");
+  long *lkey = (long *) key;
 
-  result_len = 0;
-  for (i = 8; i < key_len * sizeof (UCCollationValue); i++)
-    /* there may be nul bytes, encode byteval+1 */
-    result_len += utf8_encode (NULL, *((guchar*)key + i) + 1);
+  /* UCCollationValue format:
+   *
+   * UCCollateOptions (32/64 bits)
+   * SizeInBytes      (32/64 bits)
+   * Value            (8 bits arrey)
+   *
```
And here's your patch that seems to have been done with fink-diff, where the blank line after `gsize i;` was kept, but another blank line was deleted instead.
```patch
@@ -210,28 +210,31 @@
 {
   gchar *result;
   gsize result_len;
-  gsize i;
+  long *lkey = (long *) key;
 
-  /* Pretty smart algorithm here: ignore first eight bytes of the
-   * collation key. It doesn't produce results equivalent to
-   * UCCompareCollationKeys's, but the difference seems to be only
-   * that UCCompareCollationKeys in some cases produces 0 where our
-   * comparison gets -1 or 1. */
-
-  if (key_len * sizeof (UCCollationValue) <= 8)
-    return g_strdup ("");
-
-  result_len = 0;
-  for (i = 8; i < key_len * sizeof (UCCollationValue); i++)
-    /* there may be nul bytes, encode byteval+1 */
-    result_len += utf8_encode (NULL, *((guchar*)key + i) + 1);
+  /* UCCollationValue format:
+   *
+   * UCCollateOptions (32/64 bits)
+   * SizeInBytes      (32/64 bits)
+   * Value            (8 bits arrey)
+   *
```
The final files are identical, but the patch is different, and diffing two patches makes it seem that the change is huge.

nieder:
For the record, on 10.14, this is the originally a warning:
```
gdesktopappinfo.c:2361:75: warning: incompatible pointer types passing 'int (const char *, const char *)' to parameter of type 'GCompareFunc'
      (aka 'int (*)(const void *, const void *)') [-Wincompatible-pointer-types]
      if (!g_list_find_custom (desktop_file_ids, new_desktop_file_ids[i], strcmp))
                                                                          ^~~~~~
../glib/glist.h:90:25: note: passing argument to parameter 'func' here
                                         GCompareFunc      func);
                                                           ^
```

sth0:
I should have put in the warning/error that led to the patch.  It appeared to me that the diffs were just syntactically different but not semantically, but I wanted another set of eyes to confirm that.  At one point I started an update to glib but that got confusing and I never got it working well.  So patching the compiler warning/error seemed a good way to move forward.

nieder (approved):
LGTM. Gets rid of the warning on 10.14 and 13.6 so will assume it gets rid of the error on 15.X

nieder:
BTW, @dmacks can give you the rundown as to why we're with this old glib2.

dmacks:
> BTW, @dmacks can give you the rundown as to why we're with this old glib2.

The next few releases had self-test failures. Later-later, the gnome project switched to meson and we don't have that. A few attempts were made, but we never got a package that was fully portable to all platforms without circular dependencies and that passed self-tests everywhere. The main detail was python, deciding whether to use a single version (make this a runtime program with a private lib stash) and if so whether to use system-python to avoid a heavy dep-tree but that means different packages for different OSX, or lang-varianted (and also a single front-end via update-alternatives or something).

A stripped-down python (like in my dists fork and maybe a PR) would allow a light-weight single-version-for-all.

But I don't know (because I haven't looked in years) whether any packages would want to interact with meson at a python level (as if a public lib rather than just a runtime tool), in which case we would need lang-varianted.

PR #1170, opened 2024-11-01
blt package update: Merged separate architecture, patched configure file

sth0:
Old style configure file with no explicit type for main.  Updated for macOS 15.X / CLT 16.X  Merged two explicit architecture files into a single info file since there were no differences.  Very old source but there seems to be no replacement for many uses of it.

Sent email to maintainer, but mail bounced as invalid address.


nieder (approved):

nieder:
For the record, the original split seems to have been between i386 and x86_64 that needed different patches.

PR #1181, opened 2024-11-08
perlmods variants for 5.34.1 and 15.0

dhomeier:
Adding versions for system perl on 15.x
First commit is a bulk addition to perlmods that already have a 5.30.3 variant requiring no further changes (except a few filename fixes required for case-sensitive FS), second the same for crypto modules.
The third batch is adding new `Distribution: 15.0` versions for 5.34.1. I am not sure about the `Revision` policy for those packages (often remaining unchanged between 12.0 and 13.0, but incremented for 14.0), but since they are switching to a new systemperl, I have generally incremented it by another 100.

Built, validated and tested on 15.1 with Xcode 16.1 (arm64); some test failures for these modules:

>
- [ ] dynaloader-functions-pm-0.003-801
- [ ] devel-callchecker-pm5341-0.008-1
- [x] io-all-pm-0.51-1
- [ ] perl-critic-dynamic-pm5341-0.05-1
- [ ] b-hooks-endofscope-pm5341-0.24-1
- [ ] class-accessor-installer-pm5341-1.100.880-3
- [ ] class-accessor-complex-pm5341-1.100.880-2
- [ ] class-accessor-constructor-pm5341-1.111.590-2
- [ ] module-build-xsutil-pm-0.19-1
- [ ] devel-nytprof-pm5341-6.11-1
- [ ] namespace-clean-pm5341-0.27-1
- [ ] template-pm5341-2.27-1
- [ ] moose-pm5341-2.2207-1
- [ ] function-parameters-pm5341-2.001003-1
- [ ] mouse-pm5341-2.5.10-2
- [ ] cgi-compress-gzip-pm5341-1.03-1
- [x] kwalify-pm-1.22-1
- [x] module-info-pm-0.35-1 (supposedly fixed in 0.39, but that needs b-utils-pm5XXX, so that causes varianting problems)
- [x] sort-packed-pm5341-0.08-1
- [x] email-valid-pm5341-1.203-1 (1.204 commited, which has fix. cf Perl-Email-Project/Email-Valid#53)

Some 10 modules failed compilation with Xcode 16 and are not yet updated, plus their dependent packages; to be posted to a separate issue.

PR #1183, opened 2024-11-09
Support arm processor in libtidyp-shlibs

dhomeier:
Only a small patch to config.guess required, should even continue to support powerpc ;-)
the rest of the build proceeds without further changes.
I was able to build html-tidy-pm in #1181 with this (15.1/Xcode 16.1 arm64).


nieder:
On at least some x86_64, `uname -p` equals `i386`.
This is the case statement being fixed:
```patch
 	case `uname -p` in
 	    *86) UNAME_PROCESSOR=i686 ;;
-	    powerpc) UNAME_PROCESSOR=powerpc ;;
+	    *) UNAME_PROCESSOR=`uname -p` ;;
 	esac
```
Does the case exit immediately after a match, or does it keep traveling down the choices? If it keeps going down, then x86_64 will first be set to `UNAME_PROCESSOR=i686` as expected, but then get switched to `UNAME_PROCESSOR=i386`.

dhomeier:
It returns on the first match, the `*86` in that case. I've tested that `i386` builds as `i686-apple-darwin24.1.0`.

PR #1184, opened 2024-11-11
gnutls30-3.7 patch for updated compiler

sth0:
Function with undeclared type specified defaulted to int.  Now a specified declaration, int declaration in certs.c added,
Looked at newer upstream version (3.8.4) but not sure about Shlibs versioning.


nieder (approved):

nieder:
According to https://www.gnutls.org/download.html , gnutls v3.8.x is from their "next" stable release branch. THe download site shows 3.8.8 as the latest from that branch (not the 3.8.4 on their homepage).

Have you tried to build v3.8.4? If so, what are the full filenames of the various dylibs created?

PR #1186, opened 2024-11-12
ncftp fix for macOS 15 / Xcode 16.1

sth0:
patched configure to explicitly define main() to return int, patched in lots of places.    Builds with -m option and package and deb file look good.  Should not change deb package on older systems.


dmacks (approved):
lgtm

Issue #1092, opened 2023-10-22
libbind6 is old


dmacks:
Our package is 5.999, upstream is v9.  Our libbind6 has conflicts/replaces suggesting we used to have a bind9 package, but I can' find it. It seems the version that comes with OS X (even back to 10.13 if not before) is newer than ours, so it's a problem (#1076) that our old can mask it.

nieder:
The only packages that use libbind6 are etherape (@microtherion ) and htmldoc (@dmacks ).  Both could use an update.
Bind9 is here: https://github.com/fink/fink-distributions/blob/master/10.4/stable/main/finkinfo/net/bind9.info

If system-bind is newer than this and we don't want to bother updating bind, then just axe it, BC the two packages against it, and let system-bind take over.

dmacks:
htmldoc uses system bind via a -D flag, but has a doc note including the string "libbind6":)

cooljeanius:
> htmldoc uses system bind via a -D flag, but has a doc note including the string "libbind6":)

...so it sounds like we can ignore it, then?

cooljeanius:
Note that this was one of the things contributing to #1075, btw

PR #1190, opened 2024-12-01
etherape: v0.917

nieder:
update to last GTK2. Remove useage of fink-bind library since system-bind is newer and better.

See #1092


dmacks:
10.13's resolv.h is 19991006; fink libbind6-dev's is 20080808. Seems like a detail for improvement if ours should be upgraded.

No problem using OS X if it's good enough and .info can be Very Sure to avoid seeing fink's. Historically, BConflicts has been fragile. Can the autoconf detection be short-circuited to avoid detecting what we don't want even if it does exist? Fink's headers are in a subdir (I see you stopped passing it manually) and apple's got the functions in libSystem (force-fail libbind detection).

Could you clarify "We already force use of single threaded resolve"? Seems like a detail for improvement if the whole bind stack could be made multithreaded.

nieder:
This is the configure check output for threading:
```
checking for gethostbyaddr_r... no
checking for gethostbyaddr_r in -lbind... no
checking for gethostbyaddr_r in -lresolv... no
checking for gethostbyaddr_r in -lnsl... no
configure: gethostbyaddr_r not found, trying with gethostbyaddr
checking for gethostbyaddr... yes
configure: WARNING: Found only gethostbyaddr, while gethostbyaddr_r does not exists, resolving is restricted to single thread.
```
The ones ending in `_r` are available in  /sw/lib/libbind.4.dylib from the libbind6 package. They are not available on system-bind (at least with 10.14.6 (and I think I checked 12).

src/names/thread_resolve.c is what uses those functions, with this code:
```c
#ifdef FORCE_SINGLE_THREAD
         /* if forced single thread, uses gethostbyaddr */
         result=0;
         resultptr = gethostbyaddr (&curitem->ip.addr8, 
                          address_len(curitem->ip.type), curitem->ip.type);
#else
         /* full multithreading, use thread safe gethostbyaddr_r */
         result = gethostbyaddr_r (&curitem->ip.addr8, 
                          address_len(curitem->ip.type), curitem->ip.type, 
                          &resultbuf, extrabuf, sizeof(extrabuf), 
                          &resultptr, &errnovar);
         if (result != 0 && errnovar == ERANGE)
            g_my_critical("Insufficient memory allocated to gethostbyaddr_r\n");
#endif
```

In the configure output above, you can see where it says resolving is restricted to a single thread by not finding gethostbyaddr_r (and defines FORCE_SINGLE_THREAD). With libbind6 installed, it does find that symbol and keeps FORCE_SINGLE_THREAD unset. However, our existing patch file forces `#define FORCE_SINGLE_THREAD` w/out any conditionals, so we always travel into the non-multithreaded branch in the code regardless if gethostbyaddr_r is available or not.

https://github.com/fink/fink-distributions/blob/d27710a8b96b5d116db4a0ecae055f1dc1447f17/10.9-libcxx/stable/main/finkinfo/net/etherape.patch#L30-L34

We have libbind-6.0b1 (Jan 2009). The last release was 6.0final from March 2009: https://downloads.isc.org/isc/libbind/

cooljeanius (approved):

dmacks:
Our libbind does have the reentrant calls, but etherape would need to pass ```-pthread``` to enable that API section (not unreasonable). However, etherape as of 0.9.12 (current in fink) is calling the _r function in a way that does not match our libbind6's API of it. Newer (.18) etherape doesn't use the _r function even in reentrant mode, so once we have the newer gtk and can use the newer etherape, this set of work-arounds can be scrapped. Please add a link to this item in the Desc so future maintainers can understand it.

nieder:
Any issues left?

dmacks (approved):
lgtm

Issue #1193, opened 2024-12-12
launch does not build on Xcode 16


sth0:
I am not sure how it builds on previous versions either.  I think it is the same failure on arm64 and x86_64

```
/opt/sw/src/fink.build/launch-1.2.5-1/launch-1.2.5/launch.xcodeproj: warning: The macOS deployment target 'MACOSX_DEPLOYMENT_TARGET' is set to 10.9, but the range of supported deployment target versions is 10.13 to 15.1.99. (in target 'launch' from project 'launch')
warning: ONLY_ACTIVE_ARCH=YES requested with multiple ARCHS and no active architecture could be computed; building for all applicable architectures (in target 'launch' from project 'launch')
warning: Traditional headermap style is no longer supported; please migrate to using separate headermaps and set 'ALWAYS_SEARCH_USER_PATHS' to NO. (in target 'launch' from project 'launch')
** BUILD FAILED **


The following build commands failed:
	Ld /opt/sw/src/fink.build/launch-1.2.5-1/launch-1.2.5/launch.build/launch.build/Objects-normal/x86_64/Binary/launch normal x86_64 (in target 'launch' from project 'launch')
(1 failure)
```
It seems that Xcode 16 cannot build pre macOS 11 code.  Should there be a Distribution version for 10.15 and previous and one for macOS 11+ that has a different deployment target?  I think I can patch the Xcode project for modern systems.

nieder:
upstream commit for dealing with Xcode12.4+, which doesn't help with the MACOSX_DEPLOYMENT_TARGET issue, but gets some possible fixes for other Xcode problems that may arise.
https://github.com/nriley/launch/commit/0248d0583bf2dcccf9177cc86151cd90d7215d02

Why set MACOSX_DEPLOYMENT_TARGET to 11.0 vs 10.13 like the error suggests?

If MACOSX_DEPLOYMENT_TARGET needs to be edited to 11.0, I think it would be easiest to make new .info called `launch-1.2.5.info`, with `Distribution: 10.9, 10.10, 10.11, 10.12, 10.13, 10.14, 10.14.5, 10.15` with no changes from current .info, and update current .info to have `Distribution: 11.0, 11.3, 12.0, 13.0, 14.0, 15.0` and an updated MACOSX_DEPLOYMENT_TARGET in `launch.xcodeproj/project.pbxproj`

sth0:
[...]

PR #1197, opened 2024-12-19
Launch 1.2.5 update

sth0:
Launch will not build on newer Xcode that does not support minimum systems back to 10.9.  Updated 
Divided into 2 packages one for compilers with minimum system 10.9 and one for minimum systems 10.13

See Issue #1193

Tested on macOS 10.14.5, 15.2, 14.X


nieder:
DescPort says "broken into <=10.15 and >= 11", but the split in `Distribution:` is between 10.12 and 10.13

sth0:
Yes, have an update and improve distribution listing.  Just need to get to pushing it.

sth0:
Pushed minor changes as recommended.

nieder (approved):
LGTM

PR #1199, opened 2024-12-23
ncftp upstream update to 3.2.7

sth0:
Needed CPPFLAGS=-DHAVE_STDLIB_H to configure
also added explicit int declaration to main in configure.
Builds on macOS 15.1 / Xcode 16.2
Passes validation and builds in maintainer mode


dmacks (approved):
LGTM

PR #1208, opened 2025-01-03
libavcodec60 patch to compile on Sequoia

sth0:
This cleans up a definition to match the call.  Silences a LOT of deprecated function warnings.


dmacks:
Good find! This should also go into older versions. Testing now...

dmacks:
Just to clarify, ```-Wno-deprecated-declarations``` is a cosmetic fix and helps debug other things, not actually part of the functional solution to the FTBFS?

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
```diff
@@ -110,9 +111,12 @@ Source-Checksum: SHA256(62fe9fd098cb537d4b61731b11049e1ac179f38c870a41b8e9d556af
 ###
 PatchFile: %n_texinfo72.patch
 PatchFile-MD5: 460868fa2b167e7c57f605d1018c4374
+PatchFile2: %n.patch
```
dhomeier, line 114: would be less confusing to have that defined as `PatchFile` and the extra texinfo one as `PatchFile2`, but that's purely cosmetic.

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
```diff
@@ -110,9 +111,12 @@ Source-Checksum: SHA256(62fe9fd098cb537d4b61731b11049e1ac179f38c870a41b8e9d556af
...
+PatchFile2-Checksum: SHA256(9abfedbb69b60f2d1c182a31e517d49c8d58906993488617865d995ec9394fb7)
 PatchScript: <<
 	#!/bin/sh -ev
 	%{default_script}
+	patch -p1 < %{PatchFile2}
```
dhomeier, line 119: It appears this is already done by `%{default_script}`; with this I am getting a
> Reversed (or previously applied) patch detected!

failure.

dhomeier (changes requested):
Without the extraneous `patch` built and tested successfully on 15.2/Xcode 16.2 arm64.

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
dhomeier, line 119:
```suggestion
```
It appears this is already done by `%{default_script}`; with this I am getting a second application of `PatchFile2` and
> Reversed (or previously applied) patch detected!

failure.

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
nieder: Not needed. Keep the patches in the order they are, please. if anything, rename PatchFile2 (%n.patch) to %n-incompatible-type.patch

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
nieder: Correct. `%{default_script}` applies all listed PatchFileN from  the .info, so you don't need to manually list each one as well. If `%{default_script}` is omitted, then each PatchFile must be applied manually.

dmacks:
The same incompatible-pointer-types issue is on libavcodec58-4.0, libavcodec59-5.0, and libavcodec60-6.0 and the same patch fixes it on all three. I did it as a perl one-liner in PatchScript, but I generally prefer patchfiles, and the fact that this affects so many versions (and even in their newer versions than we have)...I agree it's clearer to have it in its own clearly-named patchfile2.

10.9-libcxx/stable/main/finkinfo/graphics/libavcodec60-6.0-shlibs.info
sth0: Ah, I didn't realize that from the default script.  Will change.  I knew that patch was not in the first patch file.

sth0:
"-Wno-deprecated-declarations is a cosmetic fix and helps debug other things"

Well actually I put that in there to help debug since there were MANY deprecations and I couldn't find the actual errors easily.  When the deprecated declarations become missing then it will be an error to debug.  But the output was huge with those declarations.  In that I found the error now I can put that extra output back in the code.  Hopefully these are fixed in version 7?

sth0:
In looking at applying the patch to other versions of libavcodec. I see that libavcodec56-2.4-shlibs and libavcodec59-5.0-shlibs are distribution restricted so I can't test on my usual system.  These are restricted to < 10.15.  However, the intermediate version libavcodec58-4.0-shlibs is not distribution restricted.  Do you want that patch applied to versions 56 and 59 or just 58 and 60?

dmacks:
libavcodec56-2.4-shlibs does not have the relevant code-block. It does have two other incompatible-pointer-types diagnostics. But given it's a dist-restricted ancient unused version and this coding situation is only a warning rather than an error on those older systems, I'm not inclined to look at fixing them.

sth0:
Thanks, I had the patch set up for 59 and 56 as well.  But wasn't sure of the policy for submitting them.  There seemed to be two different opinions on naming the patches so I held off until there was an agreement.  If there had been clear direction I would have reordered the two.
-Scott

PR #1210, opened 2025-01-11
lighttpd: update to lighttpd 1.4.77

gstrauss:
lighttpd: update to lighttpd 1.4.77

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
@@ -55,7 +55,7 @@ Source: https://download.lighttpd.net/lighttpd/releases-1.4.x/lighttpd-%v.tar.xz
 Source-Checksum: SHA256(acafabdbfa2267d8b6452d03d85fdd2a66525f3f05a36a79b6645c017f1562ce)
 
 PatchFile: %n.patch
-PatchFile-MD5: 4a26526ea8a9fa2ce46ab3648ff29165
+PatchFile-MD5: 361dc4d4e302f9ee7c7eb5610c3aca9c
 PatchScript: <<
   cat %{PatchFile} | sed 's|@PREFIX@|%p|g' | patch -p1
   glibtoolize --copy --force
@@ -70,7 +70,7 @@ ConfigureParams: <<
 	--enable-dependency-tracking \
 	--with-pcre2 \
 	--with-webdav-props \
-	--without-webdav-locks \
+	--with-webdav-locks \
 	--with-lua \
 	--with-maxminddb
 <<
--- a/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
+++ b/10.9-libcxx/stable/main/finkinfo/web/lighttpd.info
@@ -1,5 +1,5 @@
 Package: lighttpd
-Version: 1.4.76
+Version: 1.4.77
 Revision: 1
 DocFiles: AUTHORS COPYING NEWS README
 Description: Lightning fast web server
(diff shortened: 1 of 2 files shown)


dmacks (dismissed):
LGTM. I see a few support features are not enabled (brotli, xxhash, some others) that are lightweight dependencies. Worth trying to enable them?

gstrauss:
xxhash is provided as part of lighttpd (and I generally keep it up to date), though there is an option to build lighttpd against a system-provided xxhash shared library.  lighttpd 1.4.77 ships with xxHash v0.8.3.  fink-distribution currently has xxHash v0.8.0, though that won't make much difference for lighttpd use off `XXH32()`.

If fink is used on low-resource systems, it might be better to leave brotli out.  However, on more modern hardware, then yes, lighttpd mod_deflate support for brotli is recommended.

lighttpd can be built with many features enabled or disabled, so I look to the fink maintainers for guidance and assistance configuring the fink-specific syntax in the `*.info` files.

BTW, lighttpd mod_webdav implements UUIDv4, so does not rely on libuuid anymore, and web/lighttpd.info can use `--with-webdav-locks` instead of `--without-webdav-locks`, though it does not really matter since mod_webdav locks are enabled if `--with-webdav-props` is enabled and the dependencies (libxml2 and sqlite3) are present.

dmacks:
Definitely no objection to keeping small support libraries onboard. The reason I thought about those two during ./configure because they are self-contained (would not have a large dependency tree brought along. I pushed fink's xxhash to 1.4.77, thanks for the heads-up:) For brotli, it uses cmake, so it's a lightweight library itself but maybe not a simple build...meh.

Feel free to make any packaging changes you think are best.

gstrauss:
> Feel free to make any packaging changes you think are best.

I added a commit to simplify the lighttpd.conf.
Let's see what your peers think.
Feel free to edit the PR and remote that additional commit.

dmacks:
Simpler==good! The .patch MD5 needs to be updated in the .info.

I don't use lua in fink, but looking at the ./configure output, there's an upstream bug:
```
configure: Couldn\'t find lua54
```
Shell vs autotools quoting.

gstrauss:
> Simpler==good! The .patch MD5 needs to be updated in the .info.

lighttpd.info MD5 updated for lighttpd.patch

> `configure: Couldn\'t find lua54`

lighttpd `configure` tests for multiple versions of lua and multiple spellings.
Some "error" trace from configure is expected.  This part of `configure` hasn't changed in recent versions of lighttpd, so this presumably worked with lighttpd-1.4.76 in fink.  Would you share the lines in the `./configure` output above and below that line.  lighttpd `./configure --with-lua` will error out and will not build lighttpd if a version of lua to use was not found, and `--with-lua` was specified.

lighttpd `./configure` contains this:
`for luaname in lua54 lua5.4 lua-5.4 lua53 lua5.3 lua-5.3 lua52 lua5.2 lua-5.2 lua51 lua5.1 lua-5.1 lua; do` ...
so lighttpd `./configure` will check all those paths before finally trying `lua` and presumably finding `lua.pc`.

...I'll replace `Couldn\'t` with `Could not` in lighttpd `./configure` for a future release.

gstrauss:
rebased PR onto tip of fink master branch

gstrauss:
ping.  This was filed over two months ago.  Updated to lighttpd 1.4.78.  Then rebased PR onto tip of fink master branch.

@dmacks 
> Would you share the lines in the `./configure` output above and below that line. lighttpd `./configure --with-lua` will error out and will not build lighttpd if a version of lua to use was not found, and `--with-lua` was specified?

I forgot to change the spelling of "Couldn't" in `configure.ac` for lighttpd 1.4.78, but the commit to do so is on lighttpd git master.

nieder:
@dmacks ping

dmacks (approved):
lgtm, sorry this idled for so long!

PR #1214, opened 2025-01-29
llvm-15: new pkg at v15.0.7

nieder:
Uses internal bootstrap process to build clang with system-compiler, then uses that to rebuild clang and everything else. Follows Debian's SplitOff.
To add maybe: flang, OpenMP


nieder:
I guess a question would be if we want a monolithic package with just shlibs, tools, and dev splitoffs, or the current multi-splitoff that mirrors what Debian does.

BTW, I have llvm-18 built, but only on 13.7+ (might work on 12). It fails to build on 10.14 because the system-libc++ is too old.

PR #1221, opened 2025-02-20
fasteners 0.19 needs setuptools-tng

sth0:
The build process requires setuptools-tng.  Added to bdeps.


nieder:
Since it's just a simple BDep addition, I've committed the BDep change directly. Thanks for finding that.

PR #1223, opened 2025-02-21
Update tox-py to 4.24.1

sth0:
New build system for tox-py requires python version 3.8 or later.
Keep old tox-py 3.14.0 for python version <= 3.7, added newer tox-py 4.24.1 for python version >=3.8
Patch file only for version 3.14


sth0:
Starting down a rabbit hole here, seems needs pyproject-api-py  seem PR#1238

PR #1224, opened 2025-02-21
libtool2: new version

dmacks:


dmacks:
Comparable build transcript and identical test results as previous (2.4.7) on my 10.13, but given how fundamental this package is, I'd like some others to test-build it on "very recent" platforms. This also removes the help2man dependency. that's a perl-versioned system-perl mess, so getting that mess out of this low-level non-perl tree should help reduce perl chaos when new platforms are released.

wrengr:
I'm very new to this, but I cloned a copy of this PR and tried running the `./inject.pl` script but it gave me the `ERROR: Package is incomplete.` error. What else should I be doing?

nieder:
Usually the easiest way to test a PR is manually download the changed files and move them to `/opt/sw/fink/dists/local/main/finkinfo/` from your working Fink tree (can create subdirectories in there). Then if the `Trees:` field in `/opt/sw/etc/fink.conf` has `local/main` in the list of locations (it should by default), fink will find the files when trying to build. Once you're done with testing, you can remove any files in local/main as needed.

Alternatively, you can add the remote branch (https://github.com/dmacks/fink-distributions:libtool2-254 in this case) via git to your existing Fink tree, checkout that branch, and then try building. I'm not as comfortable constantly checking out remote git branches and cleaning up afterwards, but the instructions given above say these are the commands:
```
git checkout -b dmacks-libtool2-254 master
git pull https://github.com/dmacks/fink-distributions.git libtool2-254
```
YMMV on those commands so be careful.

nieder:
Built and passed all tests for me on 13.7.1

dmacks:
There's a packaging note:
```
# TODO: investigate these two test failures:
#  146: C++ exception handling (exceptions.at:398)
#  172: Run tests with low max_cmd_len  (cmdline_wrap.at:44)
# (it's been happening since at least 2.4.6 if not earlier)
```
I don't understand the test-harness runner enough to know how to mark them XFAIL, but that's why the TestScript doesn't abort.

nieder:
I saw the note. Was just correcting my oversight that I said all tests pass when it should have been all tests known to pass passed.

dmacks:
Ah ok, we'll mark this "better and not worse."

PR #1225, opened 2025-02-21
pre-commit-py 4.1.0 new package

sth0:
pre-commit-py 4.1.0 new package only for python 3.9+ needed for black which is also only 3.9+

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pre-commit-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/pre-commit-py.info
@@ -1,0 +1,62 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info2: <<
+Package: pre-commit-py%type_pkg[python]
+Version: 4.1.0
+Revision: 1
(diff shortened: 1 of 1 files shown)


nieder (changes requested):
Test failure:
```
sudo -u fink-bld [ENV] sh -c /tmp/fink.mT1hA
PYTHONPATH=. /sw/bin/python3.10 test_mccabe.py || exit 1
/sw/bin/python3.10: can't open file '/sw/build.build/pre-commit-py310-4.1.0-1/pre_commit-4.1.0/test_mccabe.py': [Errno 2] No such file or directory
### execution of PYTHONPATH=. failed, exit code 1
```
`test_mccabe.py` doesn't exist in the tarball.

2) setup.cfg says it requires `nodeenv`, but it's not present in Depends or BuildDepends.
3) Can you alphabetize the BuildDepends field, and make sure none of those should actually be Depends?

sth0:
I think this can be merged now?

sth0:
Updated to version 4.6.0 while waiting.  with python 3.10 passes fink maintainer mode.

PR #1228, opened 2025-04-01
nodeenv-py version 1.9.1 new package to satisfy pre-requisite

sth0:
nodeenv is a pre-req for pre-commit.  This version has the tests commented out since I *think* they need two more new packages and not sure how many more after that.  Builds and installs with the -mv options in fink.  See PR #1225
Uses modern bootstrap module so limited to python >= 3.8

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/nodeenv-py.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/pythonmods/nodeenv-py.info
@@ -1,0 +1,55 @@
+# -*- coding: ascii; tab-width: 4 -*-
+Info4: <<
+Package: nodeenv-py%type_pkg[python]
+Version: 1.9.1
+Revision: 1
+Type: python (3.8 3.9 3.10)
+Description: Tool to create isolated node.js environments
+License: BSD
+Homepage: https://pypi.org/project/nodeenv
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
+
+Source: https://files.pythonhosted.org/packages/source/n/nodeenv/nodeenv-%v.tar.gz
+Source-Checksum: SHA256(6ec12890a2dab7946721edbfbcd91f3319c6ccc9aec47be7c7e6b7011ee6645f)
+
+Depends: <<
+	python%type_pkg[python]
+<<
+BuildDepends: <<
+	bootstrap-modules-py%type_pkg[python],
+	setuptools-tng-py%type_pkg[python]
+<<
+
+CompileScript: <<
+	PYTHONPATH=%p/share/bootstrap-modules-python%type_pkg[python] %p/bin/python%type_raw[python] -m build --wheel --no-isolation --skip-dependency-check
+<<
+
+#InfoTest: <<
+#	TestDepends: <<
+#		pytest-py%type_pkg[python] (>= 4.3.0)
+#	<<
+#	TestScript: <<
+#		PYTHONPATH=%b/src %p/bin/python%type_raw[python] -m pytest -k -vv || exit 2
+#	<<
+#<<
+
+InstallScript: <<
+    #!/bin/sh -ev
+	PYTHONPATH=%p/share/bootstrap-modules-python%type_pkg[python] %p/bin/python%type_raw[python] -m installer --destdir %d dist/*.whl
+    for i in nodeenv ; do
+        mv %i/bin/$i %i/bin/$i-py%type_pkg[python]
+    done
+<<
+DocFiles: AUTHORS CHANGES LICENSE README README.rst README.ru.rst
+
+DescDetail: <<
+nodeenv (node.js virtual environment) is a tool to create isolated node.js
+environments.  It creates an environment that has its own installation
+directories, that doesn't share libraries with other node.js virtual
+environments.
+Also the new environment can be integrated with the environment which
+was built by virtualenv (python).
+If you use nodeenv feel free to add your project on wiki: Who-Uses-Nodeenv.
+<<
+#Info4
+<<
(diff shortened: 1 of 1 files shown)


nieder (changes requested):
It builds. I enabled tests to see what happens, and two tests fail, looking for node and/or nodejs, which is not surprising since this is for creating node environments:
```
E           subprocess.CalledProcessError: Command '('coverage', 'run', '-p', '-m', 'nodeenv', '-n', 'system', '/private/tmp/pytest-of-fink-bld/pytest-0/test_smoke_n_system_special_ch0/nenv (production env)')' returned non-zero exit status 1.

/sw/lib/python3.8/subprocess.py:364: CalledProcessError
-------------------------------------------------------------------------------- Captured stderr call --------------------------------------------------------------------------------
  File "/sw/build.build/nodeenv-py38-1.9.1-1/nodeenv-1.9.1/nodeenv.py", line 1548, in <module>
    main()
  File "/sw/build.build/nodeenv-py38-1.9.1-1/nodeenv-1.9.1/nodeenv.py", line 1130, in main
    create_environment(env_dir, args)
  File "/sw/build.build/nodeenv-py38-1.9.1-1/nodeenv-1.9.1/nodeenv.py", line 1014, in create_environment
    install_activate(env_dir, args)
  File "/sw/build.build/nodeenv-py38-1.9.1-1/nodeenv-1.9.1/nodeenv.py", line 946, in install_activate
    assert shim_node, "Did not find nodejs or node system executable"
AssertionError: Did not find nodejs or node system executable

...
dyld: cannot load 'node' (load command 0x80000034 is unknown)
sh: line 1: 84409 Abort trap: 6           node --version
```

So node should be at least included as a TestDepends once you activate tests, and should also be a RuntimeDepends or Depends.

`coverage-py` also would need to be added to TestDepends once those are enabled.

sth0:
I left the testing turned off since the source for nodejs was not available.  It searched several of the mirrors and the source but no luck.  If the distfiles.finkmirrors gets fixed then I will revisit it.  This just seemed like an endless chain of depends that was dragging me further from getting lftp fixed and used as part of duplicity which is my original goal.
"When you are up to your knees in alligators, it is hard to remember that your original intent was to drain the swamp"

sth0:
Also, if I do find the source code for nodejs, it doesn't build with an error.  The nodejs is version 12 and current version is 24 and upstream is not interested in fixing it for a valid reason.
[https://github.com/nodejs/node/issues/54187](https://github.com/nodejs/node/issues/54187)

dmacks:
Upstream is 1.10.0 now. Changing this PR's .info to that (and updating SHA256), the package builds cleanly for all three variants. It would be useful to document why the testing is turned off (but the explanation is reasonable, and I support getting this thing pushed to enable work on other things as long as this here is sufficient for that work.

nieder:
@sth0 ping.
some pytest plugins are starting to need pre-commit, so it would be nice to get this in to get pre-commit in to get those.

sth0:
I have added pre-commit PR#1225 so then this can be merged?  I am now having a dependency conflict.  For some reason building nodeenv-py310 seems to want to install
```
The following 7 additional packages will be installed:
 gyp-py39 libnode72-shlibs nodejs openssl110-dev python39 python39-shlibs
 six-py39
```
Which are python 3.9 packages!! ?  It also installed openssl110-dev which removes openssl300-dev which is also needed so the package doesn't build.  I think nodejs requires python39 to build and openssl110-dev.
```
 nodejs
The following 6 additional packages will be installed:
 gyp-py39 libnode72-shlibs openssl110-dev python39 python39-shlibs six-py39
```

sth0:
I think that installing nodejs it removes openssl300 first, installs openssl110 and then tries to build python39 which fails since it needs openssl300.

And even more confusing (at least to me) is the libnode72-shlibs replaces nodejs and for some reason it is trying to install both?

sth0:
Anyway, nodeenv is updated to version 1.10.0 here and completes all tests with python 3.10

sth0:
Get these merged in soon?  My local tree is growing!  :-)

dmacks:
nodeenv-pyXX -> nodejs and nodejs -> *-py39. That latter seems to be for internal use, so it would be great to update nodejs to py310. However, @TheSin- our ancient nodejs doesn't build for me on arm/tahoe regardless, starting with:
```
In file included from ../deps/v8/src/api/api-arguments.cc:5:
In file included from ../deps/v8/src/api/api-arguments.h:8:
In file included from ../deps/v8/src/api/api.h:9:
In file included from ../deps/v8/src/execution/isolate.h:24:
In file included from ../deps/v8/src/execution/isolate-data.h:10:
In file included from ../deps/v8/src/codegen/external-reference-table.h:10:
In file included from ../deps/v8/src/builtins/accessors.h:10:
In file included from ../deps/v8/src/objects/property-details.h:12:
../deps/v8/src/utils/utils.h:333:22: error: constexpr variable 'kMax' must be initialized by a constant expression
  333 |   static constexpr T kMax = static_cast<T>(kNumValues - 1);
      |                      ^      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../deps/v8/src/objects/code.h:441:36: note: in instantiation of template class 'v8::internal::BitField<v8::internal::Code::Kind, 1, 5>' requested here
  441 |   static_assert(NUMBER_OF_KINDS <= KindField::kMax, "Code::KindField size");
      |                                    ^
../deps/v8/src/utils/utils.h:333:29: note: integer value 31 is outside the valid range of values [0, 15] for the enumeration type 'Kind'
  333 |   static constexpr T kMax = static_cast<T>(kNumValues - 1);
      |                             ^
```
and is also passing ```-mmacosx-version-min=10.9``` (leading to a mile of warnings) even though I'm on 26.1.

PR #1232, opened 2025-04-15
Cmake system libs

nieder:
Move cmake to use system-libs where appropriate instead of it's own internal ones.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/cmake.info
+++ b/10.9-libcxx/stable/main/finkinfo/devel/cmake.info
@@ -10,14 +10,16 @@ BuildDepends: <<
 	fink-package-precedence,
 	libcurl4,
 	liblzma5,
-	libncurses5
+	libncurses5,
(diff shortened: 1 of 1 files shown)


nieder:
no comments? will commit this weekend

dmacks:
Love the concept. Not sure if it's better to enable all system-libs and then disable selectively vs only selectively enabling. We do have libuv.

nieder:
I'm not sure there's a way to whitelist only a subset of system libs vs blacklisting like now. I also only picked common ones to avoid pulling in a bunch of dependencies that few people would have via other packages.

dmacks:
> I'm not sure there's a way to whitelist only a subset of system libs vs blacklisting like now. I also only picked common ones to avoid pulling in a bunch of dependencies that few people would have via other packages.

Makes sense. I didn't know if it was "use if available" vs "use if available or else build our own copy if not".

dmacks (approved):
lgtm

PR #1222, opened 2025-02-20
librsync update to version 2.3.4

sth0:
Update very very old librsync v0.9.7 to new version 2.3.4, new source, use cmake, etc.
Contacting maintainer Evi Vanoost <vanooste@rcbi.rochester.edu>  but due to age not a lot of expectations.

PR #1233, opened 2025-04-15
Creation of librsync2 as update to librsync.

sth0:
Continuation from PR #1222 (librsync) now with clean PR and hopefully fully conformant.
This is version 2.3.4 and installs versioned and unversioned libraries as needed by duplicity and rdiff-backup the only users of librsync.  The original librsync could technically be EOL or discarded.

Not sure what I fixed.  I had this setup already done but did not pass tests.  Now passes maintainer mode tests.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/librsync2.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/librsync2.info
@@ -1,0 +1,82 @@
+Package: librsync2
+Version: 2.3.4
+Revision: 2
+Description: Algorithm for remote file synchronization
+License: GPL
+Homepage: https://github.com/librsync/librsync
+Maintainer: Scott Hannahs <shannahs@users.sourceforge.net>
+
+# Dependencies.
+Depends: %n-shlibs (= %v-%r)
+BuildDepends: <<
+	cmake,
+	bzip2-dev (>= 1.0.2-12),
+	popt (>= 1.7-1),
+	fink-buildenv-modules,
+	fink-package-precedence
+<<
+BuildDependsOnly: true
+Conflicts: librsync
+Replaces: librsync
+# Unpack Phase.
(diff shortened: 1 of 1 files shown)


nieder:
If the executable rdiff of this new version is a drop in replacement for the older version, then we should rename the -bin package as `librsync-bin`, not `%N-bin` (which resolves to `librsync2-bin`). This way, users that have librsync-bin installed will get the new version w/out having to uninstall first.

_If_ the two -bin packages are not compatible, then the -bin packages will have to Conflicts/Replaces each other, since they share common files.

nieder:
Will also need pkgconfig in BuildDepends since the build uses that to find popt. Rest looks OK so far.

sth0:
Ok, I have those changes, but now when trying to build the other package rdiff-backup which is now python versioned to -py39 and -py310 I get an error that the installed files are no -py versioned.  The original package was versioned only for py-27 and so did not have the -pyXX variants.  Should I just hard version the package to py-310 and move on?

This is also a question for the duplicity package that relies on librsync2-py  I have set it to py-310 and left it there.  The utility will run under either version.

nieder:
For rdiff-backup, I think it's best to keep it unvarianted, just as a single package that happens to use py310. I don't see a reason why someone would need to choose between running it as different python versions. End user programs that just happen to be python based are often like this. Pick a single python to build/run, and it doesn't matter for the end user.

You have the same thing with duplicity: it's an end user program that uses python and you chose python310 as the backend provider.

If you have variants, then you would have to variant all the files outside the pythonX.XX/site-packages directory so that they can coexist, and then add update-alternatives, etc etc, so a big mess if it's not needed.

nieder (approved):
LGTM. thanks

sth0:
>For rdiff-backup, I think it's best to keep it unvarianted
Works for me, I will revert my edits and test rdiff-backup.

PR #1234, opened 2025-04-18
gnulib v20250303

nieder:
This is a WIP.
Use new gnulib-tool python process that's significantly faster than shell based process (35 vs 98 minutes)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/devel/gnulib-python3.patch
+++ b/10.9-libcxx/stable/main/finkinfo/devel/gnulib-python3.patch
@@ -1,0 +1,60 @@
+diff -ruN gnulib-ac9dd00-orig/build-aux/vcs-to-changelog.py gnulib-ac9dd00/build-aux/vcs-to-changelog.py
+--- gnulib-ac9dd00-orig/build-aux/vcs-to-changelog.py	2025-03-03 15:02:04.000000000 -0600
++++ gnulib-ac9dd00/build-aux/vcs-to-changelog.py	2025-04-17 06:29:03.000000000 -0500
+@@ -1,4 +1,4 @@
+-#!/usr/bin/python3
++#!@FINK_PREFIX@/bin/python3.10
+ # Main VCSToChangeLog script.
+ # Copyright (C) 2019-2025 Free Software Foundation, Inc.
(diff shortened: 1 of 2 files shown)


nieder:
libemf is the only package currently using gnulib-2023 (borrows byteswap.h). Switching it to gnulib-2025 causes this build error in libemf:
```
In file included from libemf.cpp:75:
./byteswap.h:23:3: error: "Please include config.h first."
 #error "Please include config.h first."
  ^
./byteswap.h:28:1: error: unknown type name '_GL_INLINE_HEADER_BEGIN'
_GL_INLINE_HEADER_BEGIN
^
./byteswap.h:34:1: error: expected unqualified-id
extern "C" {
^
./byteswap.h:113:1: error: use of undeclared identifier '_GL_INLINE_HEADER_END'
_GL_INLINE_HEADER_END
^
```

This is the libemf bug report: https://sourceforge.net/p/libemf/bugs/6/

sth0:
The latest stable release is 202501 not 20250303 are we going with a snapshot?

nieder:
As far as I know there are no stable releases, so we've just grabbed a specific commit as our 'release' in the past. I chose 20250303 because that's the same date Debian is using.

sth0:
Fair enough, I was using <https://git.savannah.gnu.org/gitweb/?p=gnulib.git;a=summary> which has a list of tagged stable releases at the bottom of the page.  But your version builds just fine.

dmacks:
I worked on the gnulib package a bunch not that long ago. "Gnulib does not make releases. It is intended to be used at the source level."(https://www.gnu.org/software/gnulib/)

nieder:
libemf was the only active user of gnulib, and I've just patched it to not need gnulib in 1f06cfd188cd4fd7699de7c68e3a5e711dc371a5

sth0:
Works for me, we can drop gnulib from the distribution?

nieder:
Things in the future may still need it and it doesn't hurt to keep carrying it and it won't get used unless called explicity.

dmacks (approved):
lgtm (10.13). Not sure this will fix vs break-even-more lftp, but best to start there with a recent gnulib

Issue #1235, opened 2025-04-18
Coreutils does not build on OS 15.4.1/Xcode 16.3


sth0:
obstack.c does not compile with this error.  Does it just need to change print_and_report to a pointer?  (e.g. &print_and_report)  AFAICT, it is trying to make the obstack_alloc_failed_handler point to the function report_and_abort.  But I may have confused the number of indirections here.

Is this the extra error checking in clang 16.X?

Coreutils is needed for gnulib see PR#1234
```
  CC       lib/pthread_mutex_timedlock.o
lib/obstack.c:351:31: error: incompatible function pointer types initializing 'void (*)(void) __attribute__((noreturn))' with an expression of type 'void (void)' [-Wincompatible-function-pointer-types]
  351 | __attribute_noreturn__ void (*obstack_alloc_failed_handler) (void)
      |                               ^
  352 |   = print_and_abort;
      |     ~~~~~~~~~~~~~~~
  CC       lib/putenv.o
...
Package manager version: 0.45.99.git
Distribution version: selfupdate-git Tue Apr 15 10:09:34 2025, 15.0, x86_64
Trees: local/main stable/main
Xcode.app: 16.3
Xcode command-line tools: 16.3.0.0.1.1742442376
Max. Fink build jobs:  10
```
This is mentioned as a bug report on a self-update execution but the package was not referenced.
https://sourceforge.net/p/fink/mailman/message/58819239/

sth0:
Fixed with PR #1236

PR #1236, opened 2025-04-18
Coreutils patch for newer Clang

sth0:
Seems Clang doesn't treat the attribute as equivalent to the keyword.  Simple one line patch.  This is an upstream patch reference in the info file.  https://savannah.gnu.org/bugs/?63349

Passes maintainer mode referencing that InfoDocs: section is deprecated did not update anything else in the package. Revision bumped to force recompile.  Should be compatible with earlier versions of Clang that do not flag this inconsistency.
(answered own question in Issue #1235)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/coreutils.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/coreutils.info
@@ -1,7 +1,7 @@
 Package: coreutils
 Version: 8.32
 Epoch: 1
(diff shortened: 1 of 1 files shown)


nieder:
Have you looked at updating to latest coreutils v9.7?

sth0:
Haven't looked at updating.  Every time I do, I end up having to fix a half dozen more packages....  :-(
I will check it out.

PR #1241, opened 2025-05-04
perlmodule updates for macOS 15 / Perl 5.34.1

sth0:
Simple updates so that dependencies are satisfied, one new file for distribution 15.0

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/perlmods/package-generator-pm-15.0.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/perlmods/package-generator-pm-15.0.info
@@ -1,0 +1,21 @@
+Info2: <<
+Package: package-generator-pm
+Version: 1.106
+Revision: 1102
+Distribution: 15.0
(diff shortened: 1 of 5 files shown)


nieder (approved):
straightforward enough.
For text-charwidth-pm, you didn't need to bump %r because you were adding a new variant and the deb for existing variants would not have changed.

sth0:
Thanks, I have a bunch of more perlmods that I ended up creating.  I had to bump the OS versioned perlmods but not the perl versioned perlmods.

sth0:
Had to remove some of the version checks on some perlmods.  If there is a virtual version that is installed, then that does not have a numerical version attached to it and the version requirement can not be satisfied.

dmacks:
Versioned deps are used when a module actually needs a certain minimum version of the dependency that some perl-core do not have. Better to add what's missing

sth0:
@dmacks, I figured it out.  I needed to bump the revision number to get it to recognize the newer fink version vs the virtual package.  It kept insisting on the blank version number of the virtual package that is part of perl 5.34.1

Issue #1243, opened 2025-05-06
libnewt0-shlibs will not build on recent macOS


sth0:
It seems that libnewt0-shlibs requires BOTH Python2.7 and Python3.5.  Neither of these is available on macOS 15 / Xcode 16 but not sure when they disappeared.   Trying to build ca-certificates and not sure why this windowing toolbox is needed but here we are.  I believe it is a dependency for term-readline-gnu-pm5341.

Is there a generic "system-python" that could be used?

In looking at the page for lib newt, it seems that Python is optional and doesn't need a specific version.
https://archlinux.org/packages/extra/x86_64/libnewt/

Source:
https://pagure.io/newt
https://pagure.io/newt/releases

nieder:
libnewt declares both python27 and python35 because it needs to build the bindings when building the library and can't do it using a previously installed library.

That being said, our libnewt is 0.52.19 from 2016. Latest release is [0.52.25 from March 2025](https://pagure.io/newt/releases). The [upstream bug report](https://pagure.io/newt/issue/3) from @dmacks about the python problem was supposedly fixed in 2018.
@TheSin-

sth0:
Thanks, I will try updating to upstream and use the --with-python flag in configure.  But how do I pick which versions that fink should use?  Do I build all available versions of python and then use them?  So I would version the Distribution for the dependencies on different available python versions?  So macOS 15 would build against python 3.7, 3.8, 3.9 and 3.10?  Seems excessive.  And then need a different list of python versions for macOS 10.9 -> 14.0?  That would be a total of something like 15 different distribution versioned files.

PR #1249, opened 2025-05-25
libnewt0-shlibs update to version 0.52.25

sth0:
removed python libs since did not want explicit python version dependency for various OS versions
       Now builds on macOS 15
set new source download
alphabetized depends
removed dependence on debian utilities and docs
There are a lot of commented lines that may be obsolete on this update.  Should remove those lines before merging
See [Issue #1243](https://github.com/fink/fink-distributions/issues/1243)

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
+++ b/10.9-libcxx/stable/main/finkinfo/libs/libnewt0-shlibs.info
@@ -1,63 +1,60 @@
 Package: libnewt0-shlibs
-Version: 0.52.19
-Revision: 1.3
+Version: 0.52.25
+Revision: 1
 ###
 BuildDepends: <<
-	fink (>= 0.32),
-	fink-package-precedence,
-	slang2,
-	popt,
-	gettext-tools,
-	fribidi-dev,
-	libgettext8-dev,
-	tcltk-dev,
-	python27,
-	python35,
 	autoconf2.6,
 	automake1.15-core,
-	docbook-utils
+	docbook-utils,
+	gettext-tools,
+	fink (>= 0.32),
+	fink-package-precedence,
+	fribidi-dev,
+	libgettext8-dev,
+	popt,
+	slang2 (>= 2.0.4-2),
+	tcltk-dev
 <<
 #	sgmltools-lite replaced with docbook-utils
 Depends: <<
-	slang2-shlibs,
-	libgettext8-shlibs
+	libgettext8-shlibs,
+	slang2-shlibs
 <<
 Recommends: <<
 	fribidi-shlibs
 <<
 ###
-Source: mirror:debian:pool/main/n/newt/newt_%v.orig.tar.xz
-Source-Checksum: SHA256(ca927dcb3b5ebe8ab291a1f5988b442267b2b0d0466b8c9fac01b46fe521eab0)
-SourceDirectory: newt-%v
-Source2: mirror:debian:pool/main/n/newt/newt_%v-1.debian.tar.xz
-Source2-Checksum: SHA256(74961cb745650e9d2099bf05ce0e78903dd649a6e8ae836de79edfcff2da3b91)
-Source2ExtractDir: newt-%v
+source: https://releases.pagure.org/newt/newt-%v.tar.gz
+Source-Checksum: SHA256(ef0ca9ee27850d1a5c863bb7ff9aa08096c9ed312ece9087b30f3a426828de82)
+#Source2: mirror:debian:pool/main/n/newt/newt_%v-1.debian.tar.xz
+#Source2-Checksum: SHA256(74961cb745650e9d2099bf05ce0e78903dd649a6e8ae836de79edfcff2da3b91)
+#Source2ExtractDir: newt-%v
 ###
 PatchScript: <<
 # Patch the Debian Patches before applying
-perl -pi -e 's,libfribidi\.so\.0,libfribidi\.0\.dylib,g' debian/patches/bidi.patch
-perl -pi -e 's,/usr,%p,g' debian/patches/bidi.patch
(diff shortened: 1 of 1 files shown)


TheSin-:
All my packed I keep in line with Debian as I use fink for testing Debian things.  So I’ll have to make verianta much to my dismay

nieder:
Debian has a single python variant (py313). We don't need to have umpteenth variants. I would say just py310 and be done with it. Add Conflicts/Replaces on the py310 variant with the older py27 and py35 to upgrade anyone updating the package from the old %v.

Here's the list of patches in current debian: https://salsa.debian.org/mckinstry/newt/-/tree/debian/master/debian/patches?ref_type=heads
And this is the debian tarball for Source2: http://deb.debian.org/debian/pool/main/n/newt/newt_0.52.25-1.debian.tar.xz

Does the package really need automake-1.16 ? I don't see changes to Makefile.am

TheSin-:
I'd say 3.11 at least since that's buster default, going further back seems odd.

dt-justin:
oh you are right I have 3.10, no wonder stuff is wonky for me with python dev :D. Guess 3.10 it is 🤮

TheSin-:
oh you are right I have 3.10, no wonder stuff is wonky for me with python dev :D. Guess 3.10 it is 🤮

Man I keep using my wrong account :\

nieder:
> The usual way is to have one .info for the library itself (-shlibs and -dev, for example) and another .info for the python bindings (-py27, -py310, etc.).

This assumes that the bindings can be built independently of building the library. I can't remember what package, but there was some library that had to all be built at the same time (couldn't use compiled/installed library when making the bindings). I think the solution was to 1) have library in own .info, 2) in a separate .info (for all variants), build the library with bindings turned on for pythonX, build the binding, install everything, then delete all but the binding when making the .deb.

sth0:
@dmacks , that seems like a possibility but the original was not constructed that way for some reason.  It is not clear to me that the bindings for python 3.5 and python 2.7 were included in the build as SplitOffs because they had to be built at the same time?  It builds libnewt0-shlibs and then also newt-py35 and newt-py27 debs.  This may be the issue @nieder suggested. I am not sure of the "fink way" of dealing with this.

I fundamentally don't understand what those split offs are doing.  For example SplitOff3 lists as one of the "Files:" lib/python2.7.  Does that mean that SplitOff is install a whole modified library for that version of python.  Here is the Splitoff for python 2.7.  It installs some examples and "peanuts.py" and "popcorn.py" which don't seem to be versioned.
```
SplitOff3: <<
       Package: newt-py27
       Depends: <<
               %N (= %v-%r),
               slang2-shlibs,
               python27
       <<
       InstallScript: << 
# examples
install -d -m0755 %i/share/doc/%n/examples
install -m0755 peanuts.py %i/share/doc/%n/examples/
install -m0755 popcorn.py %i/share/doc/%n/examples/
       <<
       Files: <<
               lib/python2.7
       <<
       DocFiles: CHANGES debian/copyright
       Description: NEWT module for Python
       DescDetail: <<
This module allows you to built a text UI for your Python scripts
using newt.
       <<
<<
```

nieder:
The examples in that SplitOff3 clip are in fact versioned because they're being installed into %n (lowercase), which is the name of the current SplitOff `newt-py27`, which is versioned. %N (uppercase) is the name of the parent package.

The packaging rules are that first `InstallScript` for the parent package is run, then `InstallScript` for the SplitOff, then the rules from `SplitOff::Files` are copied over from the parent buildroot into the SplitOff buildroot directory, and finally `SplitOff::DocFiles` are copied.
This is what I found in my archive as the contents of an old deb (v0.52.19)
```
$ dpkg-deb -c /sw/fink/debs/newt-py27_0.52.19-1.3_darwin-x86_64.deb 
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/lib/
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/lib/python2.7/
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/lib/python2.7/site-packages/
-rwxr-xr-x root/admin    45804 2019-12-09 04:19 ./sw/lib/python2.7/site-packages/_snack.so
-rw-r--r-- root/admin    31105 2019-12-09 04:19 ./sw/lib/python2.7/site-packages/snack.py
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/share/
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/share/doc/
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/share/doc/newt-py27/
-rw-r--r-- root/admin    12161 2019-12-09 04:19 ./sw/share/doc/newt-py27/CHANGES
-rw-r--r-- root/admin      803 2019-12-09 04:19 ./sw/share/doc/newt-py27/copyright
drwxr-xr-x root/admin        0 2019-12-09 04:19 ./sw/share/doc/newt-py27/examples/
-rwxr-xr-x root/admin     1218 2019-12-09 04:19 ./sw/share/doc/newt-py27/examples/peanuts.py
-rwxr-xr-x root/admin     3054 2019-12-09 04:19 ./sw/share/doc/newt-py27/examples/popcorn.py
```

sth0:
Ok, so I made a bunch of split offs for the various python versions.  HOWEVER, it seems that the depends for all split offs is followed and so all versions of Python are built when building the base library.  So I guess my only way forward here is to split the files into a libnewt0.shlibs/libnewt0-dev/whiptail/libnewt-pic/libnewt-tcl version and then create a newt-py version.

But the newt-py versions are not part of the main build?  So do I just reload the source, patch it and then save the python libraries and documentation to the correct spot and there is no build phase?  @dmacks did a lot of work on this and it seems a very complex package.  I am not sure I am up to modifying it correctly.

PR #1250, opened 2025-05-25
docbook-utils update source file source

sth0:
Update link to source file since old link is non-existent and finkmaster non-functional


nieder (changes requested):
```
$ fink validate ~/Downloads/docbook-utils.info 
Validating package file /Users/nieder/Downloads/docbook-utils.info...
Warning: Field "http" is unknown. (docbook-utils.info)
```

sth0:
This push has the corrections and just tested it and builds and installs in maintainer mode.

nieder (approved):

PR #1254, opened 2025-06-02
Pkgconf to replace pkg-config

dmacks:
Following Debian's lead, we install this new rewrite at its own name ('pkgconf', bin/pkgconf) and have a 'pkgconfig' package that supplies a bin/pkg-config symlink to it for compatibility and maintaining functionality of dependencies. It has a new library suite as well, so we have -shlibs/-dev packages here.


dmacks:
This picks up the latest upstream version and full package-suite, building on #923

sth0:
The pkgconf 2.X allows xft2-dev to build and install.  So that is one data point.  According to comments in dot-info file, ppkg-conf was selected to allow building on Apple Silicon. Not sure what that issue was.  Will try to set that up to test on an AS system.

dmacks:
@sth0, what OS X and CPU do you have? I'm on an ancient 10.13 (intel/64-bit) so that is "fairly old, but not our oldest". Will need build-testing on "new" and Apple-silicon, and ideally back to 10.9. We aren't able to run its own self-test for now (due to circular dependencies in the test world itself).

sth0:
This is on my home Intel MacOS 15.5 / Xcode 16.4
I have access to a work laptop (M1) and a desktop Studio M1 Max.  all running 15.5
And an Intel iMac running 10.14 since I have a 32 bit app that I haven't managed to update.
I will try to test the laptop.

dmacks:
Just a note, I plan to squash this set of commits and update the desc notes before merging...lots of transient things here of no value.

PR #1256, opened 2025-08-03
libxaw3dxft: v1.6.4

PovlAbrahamsen:
This is a new version of libxaw3dxft, now moved from Sourceforge to https://github.com/DaveFlater/libXaw3dXft. The major change here is that the library has been renamed from libXaw3d.8.dylib to libXaw3dxft.8.dylib. I have created a copy with the old name (and then update the install name). Hopefully this should work for future updates, while maintaining backward compatibility. Tested on MacOS 15.6 Intel and Apple Silicon. Maintainer is @akhansen


dmacks:
This is a clever approach to maintaining compatibility when our original package and its Shlibs filename were out-of-sync and the new upstream version gets them into sync. Minor nit is that you'd also want to symlink the new .pc to be available at the old one's name.

However, the new version actually appears to be not-backwards-compatable. I see multiple public symbols in the old .dylib and .h that are not present in the new one. Given that, and our existing package's naming oddity, it is probably simpler to make a clean break, with a brand new package-name. The old one doesn't include the .dylib ABI-version in its name, so "libxaw3dxft8"  (or "libxaw3dxft.8") and associated -shlibs are available. That requires no renaming, linking, or other work on your part to maintain compat.

PovlAbrahamsen:
That makes sense. I'll try to write an updated, simplified package this evening, and will update this PR when ready.

PR #1257, opened 2025-08-03
xtide: v2.15.6

PovlAbrahamsen:
The latest version of xtide. Requires #1256. Maintainer is @akhansen


PovlAbrahamsen:
I've marked this as a draft, as I'm encountering some crashes with both the old and new version. I'm still trying to figure out where the bug is!

PR #1258, opened 2025-08-10
gv: patch to use latest libxaw3dxft8

PovlAbrahamsen:
A revision to gv, to compile against the latest version of libxaw3dxft. I've also cleaned up a few other bits in the info-file, removing a few unnecessary bits (e.g. the calls to pkg-config, which are now called by configure), and adding a dependency to the fontconfig2 and xft2 shlibs (which are linked in the executable).


dmacks (approved):
lgtm (deps could be upgaded even further, but switching lib is the major improvement we need)

PR #1259, opened 2025-08-10
ncview v 2.1.11

PovlAbrahamsen:
New upstream version of ncview. Switched from Fink's libxt to the built-in version. Briefly tested with libXaw3dxft, but haven't enabled it here (the extra lines are commented out in the patchscript).

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/sci/ncview.info
+++ b/10.9-libcxx/stable/main/finkinfo/sci/ncview.info
@@ -44,7 +44,7 @@ PatchScript: <<
 # The next lines can be used to switch from "2D" Athena widgets to "3D" (with
 # appropriate dependencies). But really, this looks better in 2D.
 #	perl -pi -e 's|-lXaw|-lXaw3dxft|' configure
-#	perl -pi -e 's|-lXaw3d3d|-lXaw3dxft|' configure
+#	perl -pi -e 's|-lXaw3dxft3d|-lXaw3dxft|' configure
 #	perl -pi -e 's|X11/Xaw/|X11/Xaw3dxft/|' configure src/ncview.includes.h
 <<
 ConfigureParams: <<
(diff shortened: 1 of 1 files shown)


PovlAbrahamsen:
I have tried linking to Fink's libxt - and get the following error:

> "Error: attempt to add non-widget child "ccontourform" to parent "Downloads/slp.mon.mean.nc" which supports only widgets

When I link against the system libxt, it just works. This is on Apple Silicon, MacOS 15.6, Xcode 16.4. I have tried rebuilding Fink's libxt, and linking against libxt-flat instead, but get the same result.

I have just changed computer at work, and have bootstrapped Fink, highlighting a few bugs in other packages when building on Apple Silicon with the latest Xcode. I'll submit these separately in a moment!

dmacks:
IIRC, that runtime error is a result of having a dependency mix'n'match libxt providers (ABI incompat of shared structs, or static vs shared, and maybe even breaks during X11 updates...something like that). The goal of having fink's own libxt was to enable packages to be fully self-consistent.

nieder:
So something somewhere along the linking way from ncview is probably using system/X11-libxt?

nieder:
I just freshly built and installed ncview-2.1.9-2 (on 13.7) and it's also failing with `Error: attempt to add non-widget child "ccontourform" to parent "/Users/nieder/Downloads/part.nc" which supports only widgets`. It links to libxt-shlibs. If I force Xquartz-libxt, then it runs.

These are the the non-Xt libraries ncview links against:
```
Fink libraries:
expat1-shlibs
libpng16-shlibs
netcdf-c19-shlibs
udunits2-shlibs

X11 libraries linked:
/opt/X11/lib/libICE.6.dylib
/opt/X11/lib/libSM.6.dylib
/opt/X11/lib/libX11.6.dylib
/opt/X11/lib/libXaw.7.dylib
```
Of those Fink ones, none link to any X11 libraries. And only netcdf has a largish dep tree. And that's also X11 free.
/opt/X11/lib/libXaw.7.dylib above does link to /opt/X11/lib/libXt.6.dylib.

nieder:
Curious: if you install the working 2.1.9, what's the output of `otool -L /opt/sw/bin/ncview` ? And does that differ for the failing ncview-2.1.11 built against fink-libxt?

PR #1260, opened 2025-08-12
doxygen: patch for latest clang compiler

PovlAbrahamsen:
Building doxygen failed with MacOS 15.6 and Xcode 16.4, because of missing template argument lists after the template keyword. I have modified the patch to update deps/TinyDeflate/gunzip.hh to the latest version from https://github.com/bisqwit/TinyDeflate - including the fix from #bisqwit/TinyDeflate#9


dmacks (approved):
lgtm

PR #1261, opened 2025-08-12
libblosc1: v1.21.6

PovlAbrahamsen:
I tried building libblosc1 version 1.14.4 on MacOS 15.6 and Xtools 16.4 - and somehow `internal-complibs/zlib-1.2.8/zutil.h` tried to redefine fdopen, resulting in various errors - in addition to lots of warnings about missing function prototypes. The latest upstream version, version 1.21.6, compiles with much fewer warnings, and completes all tests with "-m", with no patches required.

Maintainer is @kamischi


nieder:
built for me and passed tests on 13.7.1

PR #1264, opened 2025-08-16
emacs: updated emacsen-common, new emacs28 and emacs30

PovlAbrahamsen:
This pull request contains several changes:

- disabled emacs23 through emacs26 on Arm. They do not currently pass the configure stage.
- bumped revisions and added conflicts/replaces with new versions up to emacs30 to all previous infofiles
- updated emacsen-common to the latest version from Debian. This addresses several warnings in emacs30, and has various other changes/improvements. emacs28 still works with the previous version.
- added emacs28-nox and emacs30-nox
  - we won't be able to build versions of emacs28 and above for X11 until we have GTK+2 >= 2.24 in Fink. The X11 options have been left in the infofile for future use, but have not been tested.
  - JIT compilation using libgccjit is not enabled in emacs30

I have tested on MacOS 15.6, Apple Silicon, XCode 16.4. I've tried various combinations of upgrading and downgrading emacsen-common and emacs packages (though only emacs28-nox and emacs30-nox build on Arm) - and it seems robust! I have not tested on Intel, or with older versions of emacs.

There are various warnings about the scripts calling install-info - but the infofiles seem to work correctly.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/editors/emacsen-common.info
+++ b/10.9-libcxx/stable/main/finkinfo/editors/emacsen-common.info
@@ -1,42 +1,84 @@
 Package: emacsen-common
-Version: 1.4.15
-Revision: 6
+Version: 3.0.8
+Revision: 1
+BuildDepends: fink (>= 0.32)
 Suggests: emacsen
-Source: mirror:sourceforge:fink/%n_%v.tar.gz
-Source-Checksum: SHA256(659577da86a0c636831152a5597253fdc0a6315215beeb664d0687962e93195f)
+Source: mirror:debian:pool/main/e/%n/%n_%v.tar.xz
+Source-Checksum: SHA256(6145aba63f38a611ce99ee2c46c4213d3f76aa4760df1544dd29217fa95c6feb)
 SourceDirectory: %n-%v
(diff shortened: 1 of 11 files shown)


10.9-libcxx/stable/main/finkinfo/editors/emacsen-common.info
```diff
@@ -1,42 +1,84 @@
...
+#
+# Normally, it would be unsafe to depend on emacsen-common files like
+# this in the preinst, but our assumption for this mitigation is that
+# since emacs flavors depend on emacsen-common and since they should
+# have a installed/FLAVOR file if and only if they're "ready to go"
+# (i.e. via their postinst configure), then the emacs-remove script
+# will be available inside the loop.
+
+mkdir -p %p/var/lib/emacsen-common/state/flavor/pending
+for flavor in %p/var/lib/emacsen-common/state/flavor/installed/*[!~]; do
```
nieder, line 63: For me on 13.7/x86_64 with only emac26 installed, the installed versions are in the physical file `%p/var/lib/emacsen-common/installed-flavors`. The above `.../state/flavor/installed/...` path doesn't exist. Is this path something that started with emacs28 and/or emacsen-common-3 ?

PR #1266, opened 2025-08-19
libcfitsio10-shlibs: v4.6.2

PovlAbrahamsen:
Here is an updated version of libcfitsio10. The patchfile is no longer required; we can disable static libraries via configure. Bumped GCC to version 14. Tested on MacOS 15.6, Apple Silicon, Xcode 16.4.


PovlAbrahamsen:
> Is gcc14 actually in fink yet?

Oops - no, it isn't! I am using #1080 (with the additional fix from my comment).

PR #1267, opened 2025-08-19
postgresql12: fix for recent libxml2, link to Fink's python

PovlAbrahamsen:
There are two fixes here:

- A change in the `libxml2` headers cause an error with recent versions of clang (in `src/backend/utils/adt/xml.c`).
- And in recent versions of MacOS, we don't have the required Python headers unless the full Xcode app is installed. So one fix would be to add a build dependency on `xcode.app`. But I've opted to link to Fink's `python310`/`python310-shlibs` instead.

Tested on MacOS 15.6, Apple Silicon, Xcode 16.4. Maintainer is @wodev

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/database/postgresql12.info
+++ b/10.9-libcxx/stable/main/finkinfo/database/postgresql12.info
@@ -1,7 +1,7 @@
 Info4: <<
 Package: postgresql%type_pkg[postgresql]
-Version: 12.10
-Revision: 2
+Version: 12.22
+Revision: 1
 Type: postgresql 12
 Description: PostgreSQL open-source database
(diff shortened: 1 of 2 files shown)


nieder:
I agree unfortunately on having to use Fink python over system-python. Between the python2/python3 split and the header uncertainty, it's easier to manage with a defined release.

Any chance of updating to last postgresql12 (12.22)?

PovlAbrahamsen:
Info-file has been updated to use version 12.22. The updated patch is no longer needed. I have built with "-m", and it passes all tests on MacOS 15.6.1, Apple Silicon, Xcode 16.4.

wodev (approved):

PR #1268, opened 2025-08-19
expect: fixes for C99 and Arm

PovlAbrahamsen:
Two fixes here:

- Added `UpdateConfigGuessInDirs: tclconfig` to allow building on Arm
- `configure` is full of code that doesn't declare a type for `main`; this isn't allowed in C99, resulting in tests failing with recent versions of Clang, preventing detection of libraries that are present.

Tested on MacOS 15.6, Apple Silicon, Xcode 16.4.

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/utils/expect.info
+++ b/10.9-libcxx/stable/main/finkinfo/utils/expect.info
@@ -1,58 +1,89 @@
 Package: expect
 # dylib install_name is tied to package micro version number
-Version: 5.45
-Revision: 206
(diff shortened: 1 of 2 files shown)


nieder:
@dmacks 
Worth bumping to 5.45.4 from 2018 and not our 2010?

dmacks:
Pleae give the just-pushed 5.45.4-1 ([da7a9eb](https://github.com/fink/fink-distributions/commit/da7a9eb466b0dfca4c20ac625219b730fab25706)) a try.

PR #1273, opened 2025-09-14
gpgme11: new version

dmacks:
Unifies the .dylib, of which only the main one was really used we think


dmacks:
There are some odd-looking autoconf things...could someone please test-build this on a "fairly new" OSX? Also would like a second-opinion on the unification of the three .dylib, technically breaking Shlibs Policy but probably not in a way that affects existing packages.

dmacks:
Well this is silly. @nieder already solved it with a new %n suite.

PR #1275, opened 2025-10-03
New versions of xiph packages

dmacks:


nieder:
libgraphiz on here by accident?

nieder:
Pushed your patches (minus graphviz chunks) with the addition of `--disable-static` to libtheora1 to avoid a collision with the older libtheora0 (1.1.1-6):

```
Unpacking libtheora1-shlibs (1.2.0-1) ...
dpkg: error processing archive /opt/sw/fink/dists/stable/main/binary-darwin-x86_64/graphics/libtheora1-shlibs_1.2.0-1_darwin-x86_64.deb (--install):
 trying to overwrite '/opt/sw/lib/libtheoraenc.a', which is also in package libtheora0 1.1.1-6
dpkg-deb: error: subprocess paste was killed by signal (Broken pipe: 13)
```

PR #1283, opened 2025-11-02
gnupg2 and gpg-errors-shlibs minor version update

sth0:
Since I was into gnupg I updated to the latest version 2.5.12 -> 2.5.13 (minor version update with some security bug fixes).
Required update of gpg-errors-shlibs by a version as well.  This updated the dylib version from 40 to 41.  Again was a minor version with just updating the hash and the version numbers.

Both packages build under Xcode 26.0.1 / macOS 15.7.1 Intel in maintainer mode.

Probably did not need all the definitions for paths in the gnupg2.info file, but I kept getting errors that it could not find the libgpg-errors library.  Turned out it wanted a newer version and that the path was fine, but they don't hurt.

I realize that crypto packages should be well tested, but don't have access to an array of systems anymore.  Should all existing packages that depend on libgpg-error be tested for linking to version 41 vs. the current version 40?

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/crypto/gnupg2.info
+++ b/10.9-libcxx/stable/main/finkinfo/crypto/gnupg2.info
@@ -12,7 +12,7 @@ BuildDepends: <<
 	libcurl4,
 	libgcrypt20 (>= 1.11.0-1),
 	libgettext8-dev,
-	libgpg-error (>= 1.51-2),
+	libgpg-error (>=  1.56-1),
 	libiconv-dev,
 	libksba8 (>= 1.6.3-1),
 	libssh2.1,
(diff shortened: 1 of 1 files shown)


nieder:
The changes too libgpg-error.info are good.

Since the gnupg2 build does this: `checking for GPG Error - version >= 1.56... no`, then BuildDepends and Depends field need to be updated from current 1.51-2 to 1.56-1. Fink doesn't track the compatibility version of a dylib when it comes to dependencies, so we use the actual package version.

We don't want to drag around the `--with-FOO-prefix` options since they're all pointing to %p/lib anyway and it's extra work to carry forward and then have to remember why it was (not really) needed in the first place.

Are the configure params that add `2` to all the paths still needed? I thought adding `--program-suffix=2` was sufficient? Or are there tools where `--program-suffix` is not enough?

sth0:
I think I can eliminate all of those.  I moved them aside and then in inspecting the deb contents it seems to have all the same file paths.  I will update the pull request when I am able.

dmacks:
lgtm. GnuPG is now up to 2.5.16 (they seem to be doing a teeny-release a month, and skiped .14) and gnupg-error is 1.58 (another rapid release project), but may as well lock this ticket in and then move forward.

PR #1226, opened 2025-02-21
Black v 25.1.0 new package 

sth0:
uncompromising Python code formatter for python 3.9+

PR #1293, opened 2026-01-05
duplicity upstream update to V3.0.7

sth0:
Updated to V3.0.7 in conjunction with gnupg2 update so passes all tests but pylint.  pylint is horribly out of date.

Tested on (macOS 15.7.3 / Xcode 26.2 / Intel)


nieder:
Needs #1226

PR #1299, opened 2026-05-17
Maven update version to 3.9.16, reordered PatchFile lines

sth0:
Tried to build, but source was missing, updated to current version (2.9.16) and hash.  Passed building with -m option on MacOS 26.4.1/Intel system with no other changes.  Don't see any issues with other system versions as long as Java version is installed?  Added @RangerRick as maintainer to review PR.

Moved the patch file definition and hash to before the patch operation.


nieder (approved):
LGTM

PR #1307, opened 2026-07-17
pbr-py version update

sth0:
updating from version 5.1.0 to 7.0.3, simple version update and checksum.  Passes fink -m validation.


nieder (approved):
lgtm

PR #1309, opened 2026-07-26
curl4 upstream update.  

sth0:
No changes to build so should work on OS versions.   Builds with -m options and passes tests.  Should satisfy user request for Medium severity security patches.


dmacks (approved):
lgtm

Issue #1115, opened 2024-01-06
xcodebuild is broken with fink-bld home directory on macOS 14 / Xcode 15


dhomeier:
On Sonoma 14.1.2 with Xcode 15.1 builds calling `xcodebuild` fail because it is trying to create stuff in the `fink-bld` user's Directory Services `NFSHomeDirectory`, which is `/var/empty`, rather than the `$HOME` set up by Fink.
There are various defaults in `showBuildSettings` using this, all of which afaics can be pointed to Fink's `$HOME` by setting 3 variables, but `xcodebuild` then still tries to create cache and log directories, the location of which I have not found any settings to control (for building `launch` below):
```
/usr/bin/xcodebuild -UsePerConfigurationBuildLocations=NO SYMROOT=. OBJROOT=. HOME=$HOME MODULE_CACHE_DIR=${HOME}/Library/Developer/Xcode/DerivedData/ModuleCache.noindex SHARED_PRECOMPS_DIR=${HOME}/Library/Developer/Xcode/DerivedData/PrecompiledHeaders
2024-01-06 01:06:24.395 xcodebuild[96395:26968448] [MT] DVTDownloadable: Error creating downloadable cache directory file:///var/empty/Library/Caches/com.apple.dt.Xcode/Downloads/
Command line invocation:
    /Applications/Xcode.app/Contents/Developer/usr/bin/xcodebuild -UsePerConfigurationBuildLocations=NO SYMROOT=. OBJROOT=. HOME=/tmp/fink-build-HOME.o1Oo_jWsBi MODULE_CACHE_DIR=/tmp/fink-build-HOME.o1Oo_jWsBi/Library/Developer/Xcode/DerivedData/ModuleCache.noindex SHARED_PRECOMPS_DIR=/tmp/fink-build-HOME.o1Oo_jWsBi/Library/Developer/Xcode/DerivedData/PrecompiledHeaders

User defaults from command line:
    IDEPackageSupportUseBuiltinSCM = YES
    UsePerConfigurationBuildLocations = NO

Build settings from command line:
    HOME = /tmp/fink-build-HOME.o1Oo_jWsBi
    MODULE_CACHE_DIR = /tmp/fink-build-HOME.o1Oo_jWsBi/Library/Developer/Xcode/DerivedData/ModuleCache.noindex
    OBJROOT = .
[...]

PR #1313, opened 2026-08-18
telnet: allow building on newer Xcodes (>15?) 

### LINKED ISSUE #1296: No telnet beyond 11.0 distro
Our telnet.info is Apple's https://github.com/apple-opensource/remote_cmds v60, which doesn't build on my 26.0 (looks like issue #1115), which is a problem because recent OS X stopped supplying telnet. There is a newer release v63 on that apple repo, which is still almost 6 years old, and that I have not yet tried. Anyone know why we are using Apple's rather than GNU's? I tried inetutils-2.7 (what Debian is currently using)...its build-system is a bit of a mess, but it gave me a functional telnet executable.

nieder:
by using -derivedDataPath flag for xcodebuild.

This keeps the build process from trying to write to HOME=/var/empty See #1115
Closes #1296

### DIFF
--- a/10.9-libcxx/stable/main/finkinfo/net/telnet.info
+++ b/10.9-libcxx/stable/main/finkinfo/net/telnet.info
@@ -1,7 +1,7 @@
 Package: telnet
 Version: 60
-Revision: 1
-Distribution: 10.13, 10.14, 10.14.5, 10.15, 11.0
+Revision: 2
+Distribution: 10.13, 10.14, 10.14.5, 10.15, 11.0, 11.3, 12.0, 13.0, 14.0, 14.4, 15.0, 26.0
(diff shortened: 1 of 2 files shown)


nieder:
Tested to build on 26.6/arm64 and 10.14.6.

Can we get rid of the `Distribution:` field? It was originally set by @akhansen because Sierra included telnet, so it was added to High Sierra and later. But there's no real reason to keep it out of 10.9-10.12. Those people don't want it, they can just not install it and it makes .info maintenance easier.

dhomeier:
Tested to work on 27.0 beta 6 as well!

nieder:
Tested to work on 10.14 and 27, which is very broad coverage.

